PR #55, opened 2021-02-25
self encryption auth for public keys fix

Sameeraja-Bharathula:


at_client_mobile/lib/src/at_client_service.dart
```diff
@@ -403,6 +403,16 @@ class AtClientService {
       ..key = AT_ENCRYPTION_SHARED_KEY;
     var oldSelfKeyValue = await atClient.get(atKey);
     if (oldSelfKeyValue == null || oldSelfKeyValue.value == null) {
+      //check and store selfEncryption key
+      var newSelfEncryptionKey =
+          await atClient.getLocalSecondary().getEncryptionSelfKey();
+      if (newSelfEncryptionKey == null || newSelfEncryptionKey == '') {
+        newSelfEncryptionKey =
+            await _keyChainManager.getSelfEncryptionAESKey(currentAtSign);
```
murali-shris, line 411: ios will maintain the data in keychain even on uninstall..in android what happens to this call to keyChainManager.getSelfEncryptionAESKey(..) ?

Sameeraja-Bharathula: This will be used when a user updates the app without uninstalling and authenticated previously with cram secret. As auth cram is not storing the data to local Secondary.

murali-shris (approved):

PR #82, opened 2021-03-22
get encryption shared key from local secondary

naresh0689:


at_client/lib/src/service/encryption_service.dart
```diff
@@ -421,4 +416,22 @@ class EncryptionService {
 
     return sharedWithPublicKey;
   }
+
+  Future<String> _getEncryptedSharedKey(String sharedBy) async {
+    //1.lookup shared key
+    var sharedKeyLookUpBuilder = LookupVerbBuilder()
```
murali-shris, line 422: please check decrypt(..) method..we have used isCached flag while looking in local  secondary..this method and what is inside decrpt should be identical

murali-shris (approved):

kalluriramkumar (approved):

PR #83, opened 2021-03-24
export encryption_util.dart

Arkangel12:
This is used at the `ServerDemoService` it the most basic example and it can be confusing for newcomers not to know from where or why this is sending a pedantic error besides if your project is running `flutter analyze` this will fail, you can omit this rule (not recommended) but if it needs to be used in the method `encryptKeyPairs` from `server_demo_service.dart`  it should be exported. 

This is the broken [pedantic rule](https://dart-lang.github.io/linter/lints/implementation_imports.html) `Don't import implementation files from another package.`


murali-shris:
@Arkangel12 change you have suggested is in the trunk branch. 
https://github.com/atsign-foundation/at_client_sdk/blob/trunk/at_client/lib/at_client.dart

PR #123, opened 2021-04-19
Stream sync response

sitaram-kalluri:


at_client/lib/src/client/remote_secondary.dart
```diff
@@ -71,16 +76,16 @@ class RemoteSecondary implements Secondary {
...
     var regexString = (regex != null && regex != 'null' && regex.isNotEmpty)
         ? ':$regex'
         : ((_preference.syncRegex != null && _preference.syncRegex.isNotEmpty)
             ? ':${_preference.syncRegex}'
             : '');
-
     atCommand += '$regexString\n';
-    var syncResult = await atLookUp.executeCommand(atCommand, auth: true);
+    atLookupSync.syncCallback = syncCallback;
+    var syncResult = await atLookupSync.executeCommand(atCommand, auth: true);
```
VJag: Can this be return await atLookupSync.executeCommand(atCommand, auth: true);?

at_client/lib/src/manager/sync_manager.dart
```diff
@@ -110,8 +113,8 @@ class SyncManager {
       // cloud is ahead if server commit id is > last synced commit id in local
       if (serverCommitId > lastSyncedCommitId) {
         // send sync verb to server and sync the changes to local
-        var syncResponse =
-            await _remoteSecondary.sync(lastSyncedCommitId, regex: regex);
+        var syncResponse = await _remoteSecondary.sync(lastSyncedCommitId,
+            regex: regex, syncCallback: _syncLocal);
         if (syncResponse != null && syncResponse != 'data:null') {
           syncResponse = syncResponse.replaceFirst('data:', '');
           var syncResponseJson = jsonDecode(syncResponse);
```
VJag: Can we prevent creation of "syncResponseJson ". If we use jsonDecode(syncResponse) with For will it be called everytime?

sitaram-kalluri: Since we passing a function callback, control does not come into the IF condition. Hence removed the code and tested it. It is working fine.

cpswan (commented):
Similar to last time... I'm seeing changes to code, but without accompanying tests or where those tests are run.

murali-shris (approved):

PR #172, opened 2021-06-10
Notification changes

naresh0689:
**- What I did**

**- How I did it**

**- How to verify it**


at_client/lib/src/client/at_client_impl.dart
```diff
@@ -196,9 +195,9 @@ class AtClientImpl implements AtClient {
     var builder = DeleteVerbBuilder()
       ..isCached = isCached
       ..isPublic = isPublic
-      ..sharedWith = sharedWith
+      ..sharedWith = sharedWith!
```
murali-shris: delete:phone@alice is a valid command. sharedWith will be null and will fail here

at_client/lib/src/client/at_client_impl.dart
```diff
@@ -434,9 +433,9 @@ class AtClientImpl implements AtClient {
       String? sharedWith,
       bool isDedicated = false}) async {
     var builder = ScanVerbBuilder()
-      ..sharedWith = sharedWith
-      ..sharedBy = sharedBy
-      ..regex = regex
+      ..sharedWith = sharedWith!
+      ..sharedBy = sharedBy!
+      ..regex = regex!
```
murali-shris: regex is optional  for scan.. regex! will throw runtime error if regex is not passed

murali-shris:
please re-visit whereever you have marked variables with !
If the any value is null at runtime, it will fail

murali-shris:
@naresh0689 
Please check analysis issues
https://github.com/atsign-foundation/at_client_sdk/pull/172/checks?check_run_id=2794144470

PR #177, opened 2021-06-21
Stream v2

murali-shris:
**- What I did**

**- How I did it**

**- How to verify it**


at_client/lib/src/client/at_client_impl.dart
```diff
@@ -889,18 +901,21 @@ class AtClientImpl implements AtClient {
       if (streamResult != null && streamResult.startsWith('stream:done')) {
         remoteSecondary.atLookUp.connection!.close();
         streamResponse.status = AtStreamStatus.COMPLETE;
+        onDone(streamResponse);
       }
     } else if (result != null && result.startsWith('error:')) {
       result = result.replaceAll('error:', '');
       streamResponse.errorCode = result.split('-')[0];
```
VJag, line 927: Should the sender keep track of the last successful byte written

VJag: Or should the server expose stream:status:12345

This should return:data: {"status":"Inprogress/Complete/NA/cancelled", "bytesTransferred" : "1234", "transferInitiatedOn": "<mills>"}

at_client/lib/src/manager/stream_manager.dart
```diff
@@ -0,0 +1,49 @@
...
+  StreamManager._internal(this._atSign);
+
+  /// onDone[AtStreamResponse]
+  /// onError[AtStreamResponse]
+  Future<void> send(
+  String sharedWith, String filePath, Function onDone, Function onError) async {
+
+  }
+
+  Future<void> resume(
```
VJag: Stream.sender.send.  Stream.recipient.resume

VJag: In my opinion, we need the following:

Sender side:
-------------
Stream.Sender.send(), status(), cancel()

Recipient side:
---------------
Stream.Recipient.resume(), status(), cancel()

PR #182, opened 2021-06-24
backward compatibility for sync stream

sitaram-kalluri:
**- What I did** Provide backward compatibility for sync stream. 

**- How I did it** Add an optional parameter isStream to sync method. 
If set to true sync stream is initiated. Else regular sync is initiated.

**- How to verify it** 

### DIFF
--- a/at_client/lib/src/manager/sync_manager.dart
+++ b/at_client/lib/src/manager/sync_manager.dart
@@ -37,6 +37,7 @@ class SyncManager {
   late AtClientPreference preference;
 
   /// Calling sync with [regex] will ensure only matching keys are synced.
+  /// Setting [isStream] to true will process sync using stream
   /// sync will be retried on any connection related errors.
   /// [onDone] callback is invoked when sync completes successfully.
   /// e.g onDone callback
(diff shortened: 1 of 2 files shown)


at_client/lib/src/manager/sync_manager.dart
```diff
@@ -37,6 +37,7 @@ class SyncManager {
   late AtClientPreference preference;
 
   /// Calling sync with [regex] will ensure only matching keys are synced.
+  /// Setting [isStream] to true will process sync using stream
```
murali-shris, line 40: we have to elaborate a bit here..

at_client/lib/src/manager/sync_manager.dart
```diff
@@ -115,9 +118,25 @@ class SyncManager {
     }
   }
 
-  Future<void> _pullChanges(SyncObject syncObject, {String? regex}) async {
-    await remoteSecondary.sync(syncObject.lastSyncedCommitId, _syncLocal,
-        regex: regex);
+  Future<void> _pullChanges(SyncObject syncObject,
+      {String? regex, bool isStream = false}) async {
+    // If isStream is false, process regular sync response and return.
+    if (!isStream) {
```
murali-shris: if(isStream) {
// return await remoteSecondary.sync(...);
}
// non stream sync code here.

murali-shris (approved):

PR #214, opened 2021-07-22
Stream resume support

murali-shris:
**- What I did**

Support for stream resume , redesign stream methods and add parsers

**- How I did it**

- Deprecated old stream method
- added new method createStream in spec
- moved stream logic from at_client_impl to StreamSender and StreamReceiver classes

**- How to verify it**
- Run stream sender and stream receiver code in samples
- To test stream resume, stop the receiver after certain bytes have been received. Call stream resume with start byte = received bytes+1


gkc:
@murali-shris Is there code here which will be useful when we revisit stream properly? If so, please mark as draft for now, otherwise please close and delete the branch.

PR #231, opened 2021-08-09
utc time zone and statusCode check added

sachins-geekyants:


murali-shris (approved):

at_client/lib/src/service/file_transfer_service.dart
```diff
@@ -23,6 +23,9 @@ class FileTransferService {
       FileTransferObject fileTransferObject, String downloadPath) async {
     try {
       var response = await http.get(Uri.parse(fileTransferObject.fileUrl));
+      if (response.statusCode != 200) {
+        return '';
```
murali-shris: we are returning fileDownloadpath from this function..if caller gets empty string..he has to do additional check before downloading file..instead a cleaner solution will be return a custom object from this method
e.g.
FileDownloadResponse {
String filePath;
int statusCode;
bool isError;
}

sachins-geekyants: @murali-shris added a model for downloadResponse.

PR #236, opened 2021-08-19
Skip stats notification.

sitaram-kalluri:
**- What I did** Skip stats notification

**- How I did it** Add if condition to skip notification from stats service,

**- How to verify it**

### DIFF
--- a/at_client/lib/src/service/notification_service_impl.dart
+++ b/at_client/lib/src/service/notification_service_impl.dart
@@ -82,14 +82,11 @@ class NotificationServiceImpl implements NotificationService {
       final atNotifications = notificationParser
           .getAtNotifications(notificationParser.parse(notificationJSON));
       atNotifications.forEach((atNotification) async {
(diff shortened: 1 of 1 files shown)


at_client/lib/src/service/notification_service_impl.dart
```diff
@@ -82,6 +82,12 @@ class NotificationServiceImpl implements NotificationService {
       final atNotifications = notificationParser
           .getAtNotifications(notificationParser.parse(notificationJSON));
       atNotifications.forEach((atNotification) async {
+        //StatsNotificationServer sends notification to
+        // keep monitor alive with notificationId is -1.
+        // Do nothing here.
+        if(atNotification.notificationId == '-1'){
+          return;
```
VJag: @kalluriramkumar It's not a return. When -1 loop should "continue". Do nothing is not "return" since it is in a loop.

murali-shris (approved):

PR #242, opened 2021-08-24
Switch atsign

murali-shris:
**- What I did**

**- How I did it**

**- How to verify it**


at_client/lib/src/manager/at_client_manager.dart
```diff
@@ -0,0 +1,33 @@
...
+  AtClientManager._internal();
+
+  factory AtClientManager.getInstance() {
+    return _singleton;
+  }
+
+  void setCurrentAtSign(
+      String atSign, String namespace, AtClientPreference preference) {
+    _atSign = _atSign;
+    currenAtClient = AtClientImpl(_atSign, namespace, preference);
```
VJag: I believe the map inside should ensure that for the same @sign we don't create multiple instances of AtClient

at_client/lib/src/service/notification_service_impl.dart
```diff
@@ -191,6 +192,12 @@ class NotificationServiceImpl implements NotificationService {
     _init();
     return _controller.stream;
   }
+
+  @override
+  void listenToAtSignChange(AtClient atClient) {
+    stop();
```
VJag: I am not sure if this right? lets discuss once.

PR #275, opened 2021-09-13
Validations Added

yahu1031:
## What I did ##

- Added Validations

### DIFF
--- a/at_client/lib/src/client/remote_secondary.dart
+++ b/at_client/lib/src/client/remote_secondary.dart
@@ -76,7 +76,7 @@ class RemoteSecondary implements Secondary {
       ..regex = regex;
 
     var atCommand = syncVerbBuilder.buildCommand();
-    return await atLookUp.executeCommand(atCommand, auth: true);
+    return atLookUp.executeCommand(atCommand, auth: true);
   }
 
   ///Executes monitor verb on remote secondary. Result of the monitor verb is processed using [monitorResponseCallback].
--- a/at_client/lib/src/util/at_client_validation.dart
+++ b/at_client/lib/src/util/at_client_validation.dart
@@ -1,37 +1,50 @@
 import 'package:at_client/at_client.dart';
+import 'package:at_client/src/util/network_util.dart';
 import 'package:at_commons/at_commons.dart';
+import 'package:at_utils/at_utils.dart';
 
 class AtClientValidation {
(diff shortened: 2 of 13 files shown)


at_client/lib/src/util/at_client_validation.dart
```diff
@@ -1,37 +1,50 @@
...
   }
 
   /// Validates the metadata of the key.
   /// Throws [AtKeyException] if metadata has invalid values.
-  static void validateMetadata(Metadata? metadata) {
+  static void _validateMetadata(Metadata? metadata) {
     if (metadata == null) {
       return;
-    }
+    } else
```
sitaram-kalluri: Please remove else, if "IF" condition satisfies, exception is thrown or gets to next if condition.

at_client/lib/src/util/at_client_validation.dart
```diff
@@ -51,4 +66,44 @@ class AtClientValidation {
       throw AtKeyException('$atSign does not exist');
     }
   }
+
+  /// Validates the atKey.
+  /// Throws [UnAuthorizedException] if atKey is a cached one.
+  static Future<void> validateAtKey(AtKey atKey, {AtClientPreference? preferences}) async {
+    _validateKey(atKey.key!);
+    _validateNamespace(atKey.namespace);
+    if ((atKey.metadata != null && atKey.metadata!.isCached) || atKey.key!.startsWith('cached:')) {
```
sitaram-kalluri: Please add a condition to avoid caching of self keys

if sharedBy equals currentAtSign and isCaching is true, throw exception.

at_client/lib/src/client/remote_secondary.dart
```diff
@@ -76,7 +76,7 @@ class RemoteSecondary implements Secondary {
       ..regex = regex;
 
     var atCommand = syncVerbBuilder.buildCommand();
-    return await atLookUp.executeCommand(atCommand, auth: true);
+    return atLookUp.executeCommand(atCommand, auth: true);
```
sitaram-kalluri, line 79: why is await removed?

yahu1031: during returns, we don't need awaits

yahu1031: it will return the value when it was fetched. so we don't need to await there.

PR #284, opened 2021-09-16
set data change in sync result

murali-shris:
**- What I did**

**- How I did it**

**- How to verify it**

### DIFF
--- a/at_client/lib/src/manager/monitor.dart
+++ b/at_client/lib/src/manager/monitor.dart
@@ -135,7 +135,7 @@ class Monitor {
       }, onError: (error) {
         _logger.severe('error in monitor $error');
         _handleError(error);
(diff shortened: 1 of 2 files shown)


at_client/lib/src/manager/monitor.dart
```diff
@@ -135,7 +135,7 @@ class Monitor {
       }, onError: (error) {
         _logger.severe('error in monitor $error');
         _handleError(error);
-      });
+      }, cancelOnError: true);
```
VJag: Should we have this in this PR?

VJag (approved):

at_client/lib/src/manager/monitor.dart
murali-shris: not required..I will revert

PR #319, opened 2021-10-27
Fix minor bugs in encryption_service.dart

### LINKED ISSUE #321: Analyse decryption issues
Analyse and come up with approach for below scenarios

- [ ] Public key changes
- [ ] Encryption key changes

sitaram-kalluri:
**- What I did**

**- How I did it**

**- How to verify it**


at_client/lib/src/service/encryption_service.dart
```diff
@@ -119,6 +126,11 @@ class EncryptionService {
   /// Used for local lookup @bob:phone@alice
   Future<String?> decryptLocal(String? encryptedValue, String? currentAtSign,
       String sharedWithUser) async {
+    if (encryptedValue == null || encryptedValue.isEmpty) {
```
murali-shris, line 129: can we also log in the caller when encryptedValue is null.so we will know for which key the encryptedValue is null

murali-shris (approved):

PR #332, opened 2021-11-18
Desktop compatible

sarika01:
Replacing flutter_keychain package with biometric_storage to support desktop platforms


cpswan (approved):

gkc:
As per discussion at the stand-up; in order to have a graceful upgrade, we need to do something like the following:

On creation of the KeyChainManager singleton instance for an app:
* Check if there is any data in biometric_storage
  - If no data found
    - Copy all existing key-values from flutter_keychain to biometric_storage
    - Verify that all of the key-values have been copied from flutter_keychain to biometric_storage

cpswan (approved):

Issue #347, opened 2021-12-12
Split unit tests from functional and end-to-end tests


gkc:
**Describe the bug**
Unit tests should have no external dependencies. Currently unit tests, functional tests and end-to-end testss are mixed together in at_client/test; instead we should have unit tests only in at_client/test and create new top-level directories for functional tests and end-to-end tests as we have in the at_server repo

**To Reproduce**
run dart test from the command line or IDE. Some of the tests will fail as they need to be able to contact a remote secondary.

**Expected behaviour**
The tests all pass as they do not have an external dependencies.

gkc:
Fixed by #348

gkc:
After a chat with cpswan earlier, I will move those two test files (atclient_notify_test.dart and atclient_put_test.dart) into at_functional_tests rather than at_end2end_tests so we're consistent on definitions of functional vs end2end across our repos

gkc:
Forgot to close this, was done before the holidays

PR #348, opened 2021-12-12
Separate unit tests and end2end tests

gkc:
**- What I did**
Split out some tests from at_client/test to a new at_end2end_test directory, as they were in fact end-to-end tests rather than unit tests; also added code coverage measurement for the unit tests.

**- How I did it**
* created new top-level directories at_functional_test and at_end2end_test for functional tests and end-to-end tests
* moved some tests from at_client/test/functional_tests to at_end2end_test/test as they are in fact end-to-end tests not unit tests
* Changed version for dependency at_persistence_secondary_server from 3.0.8 to 3.0.11 in order to pick up the new isKeyExists method
* Modified at_client_sdk.yaml GitHub workflow to separate out the test runs for unit tests and end2end tests, and run unit tests with code coverage measurement uploading to codecov.io

**- How to verify it**
at_client_sdk.yaml workflow completes successfully when this pull request is created

Split out some tests from at_client/test to a new at_end2end_test directory, as they were in fact end-to-end tests rather than unit tests; also added code coverage measurement for the unit tests.

### DIFF
--- a/.github/workflows/at_client_sdk.yaml
+++ b/.github/workflows/at_client_sdk.yaml
@@ -56,9 +56,24 @@ jobs:
         working-directory: at_client
         run: dart run test/functional_test/check_test_env.dart
 
-      # Run unit and functional tests
-      - name: Run unit and functional tests
+      # Run unit tests
+      - name: Run unit tests, with code coverage measurement
         working-directory: at_client
+        run: dart test --concurrency=1 --coverage="coverage"
+
+      - name: Convert coverage to LCOV format
+        working-directory: at_client
+        run: pub run coverage:format_coverage --lcov --in=coverage --out=coverage.lcov --packages=.packages --report-on=lib
+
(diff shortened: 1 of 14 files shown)


gkc:
Addresses #347

gkc:
This is a restructuring and effectively a no-op as it doesn't add any new tests, just reorganises things a bit, following the same pattern as in the at_server repo where we have unit tests under each of the packages, and then we have separate directories for functional tests and end-to-end tests

@purnimavenkatasubbu @VJag  tagging you on this PR to make sure you have visibility

nickelskevin (approved):

gkc:
@VJag @purnimavenkatasubbu After a chat with @cpswan earlier, I will move those two test files (atclient_notify_test.dart and atclient_put_test.dart) into at_functional_tests rather than at_end2end_tests so we're consistent on definitions of functional vs end2end across our repos

PR #363, opened 2021-12-17
Fixed up functional test directory structure

gkc:
**- What I did**
Fixed up functional test directory structure

**- How I did it**
1) Moved functional tests from at_end2end_test where I incorrectly put them earlier
2) Moved all functional-test-related setup from various places in at_client/test to at_functional_test/test
3) Modified .github/workflows/at_client_sdk.yaml accordingly

**- How to verify it**
GitHub workflow runs and executes functional tests successfully

Fixed up functional test directory structure


cpswan (commented):
Do we need to keep the end2end tests directory, or can that be dropped for now?

gkc:
> Do we need to keep the end2end tests directory, or can that be dropped for now?

Good idea, yeah it's not used so I've removed it

cpswan (approved):

PR #375, opened 2022-01-12
refactoring for AtClientImpl, SyncServiceImpl and NotificationServiceImpl

### LINKED ISSUE #353: replace AtClientManager.getInstance()
eliminate use of AtClientManager.getInstance() and instead have AtClientImpl, NotificationService, SyncService become aware of each other. An interim step would be to add methods like getNotificationServiceForAtSign(atSign), getSyncServiceForAtSign(atSign), getAtClientForAtSign(atSign)

murali-shris:
**- What I did**

**- How I did it**

**- How to verify it**

### DIFF
--- a/at_client/lib/src/client/at_client_impl.dart
+++ b/at_client/lib/src/client/at_client_impl.dart
@@ -46,7 +46,7 @@ class AtClientImpl implements AtClient {
   @override
   EncryptionService? get encryptionService => _encryptionService;
 
-  late AtClientManager _atClientManager;
+  AtClientManager? _atClientManager;
 
   final _logger = AtSignLogger('AtClientImpl');
   static final Map _atClientInstanceMap = <String, AtClient>{};
@@ -64,6 +64,7 @@ class AtClientImpl implements AtClient {
   }
 
   @Deprecated("Use [create]")
+
   /// use [create]
   static Future<void> createClient(String currentAtSign, String? namespace,
       AtClientPreference preferences) async {
@@ -89,10 +90,8 @@ class AtClientImpl implements AtClient {
   }
 
   static Future<AtClient> create(
-      AtClientManager atClientManager,
-      String currentAtSign,
-      String? namespace,
-      AtClientPreference preferences) async {
+      String currentAtSign, String? namespace, AtClientPreference preferences,
+      {AtClientManager? atClientManager}) async {
     currentAtSign = AtUtils.formatAtSign(currentAtSign)!;
     if (_atClientInstanceMap.containsKey(currentAtSign)) {
       return _atClientInstanceMap[currentAtSign];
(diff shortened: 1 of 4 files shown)


at_client/lib/src/manager/at_client_manager.dart
```diff
@@ -30,24 +30,28 @@ class AtClientManager {
 
   AtClientManager._internal();
 
+  @deprecated
   factory AtClientManager.getInstance() {
```
murali-shris: @gkc I am not fully able to remove the singleton related artifacts in order to maintain backward compatibility. Within the at client SDK we will be removing the references to AtClientManager.getInstance(). However in widgets and app code, there will be references to AtClientManager.getInstance() which will return the static object _singleton. I am skeptical whether this may cause any bugs (having constructed instance of AtClientManager and _singleton)

murali-shris: I made a change this morning. Instead of creating new instance of AtClientManager inside setCurrentAtSign, I passed "this" to AtClientImpl.create and SyncServiceImpl.create..So callers using getInstance the old way will have the singleton reference of AtClientManager..callers using the new constructor AtClientManager(_atSign)  will have one manager per atsign. Is this fine?

gkc: Yes I think that's fine, given the objectives at this point. There's nothing fundamentally wrong with how this works from a client usage perspective; at this point this change is aimed at making the AtClient, SyncService and NotificationService implementations more testable

at_client/lib/src/client/at_client_impl.dart
```diff
@@ -79,15 +80,18 @@ class AtClientImpl implements AtClient {
...
     _preference = preference;
     _namespace = namespace;
   }
 
-  static Future<AtClient> create(String currentAtSign, String? namespace,
+  static Future<AtClient> create(
+      AtClientManager atClientManager,
+      String currentAtSign,
+      String? namespace,
       AtClientPreference preferences) async {
```
gkc: For backwards compatibility, you could make atClientManager an optional param and if it's not supplied use the current static mechanism to get the instance

at_client/lib/src/service/sync_service_impl.dart
```diff
@@ -42,21 +42,26 @@ class SyncServiceImpl implements SyncService, AtSignChangeListener {
 
   final _logger = AtSignLogger('SyncService');
 
-  static Future<SyncService> create(AtClient atClient) async {
+  late AtClientManager _atClientManager;
+
+  static Future<SyncService> create(
+      AtClientManager atClientManager, AtClient atClient) async {
```
gkc: See my comment in AtClientImpl.create(...) re backwards compatibility

at_client/lib/src/client/at_client_impl.dart
murali-shris: done

at_client/lib/src/manager/at_client_manager.dart
```diff
@@ -30,24 +30,27 @@ class AtClientManager {
 
   AtClientManager._internal();
 
+  @deprecated
```
gkc: Shouldn't deprecate, as this is still a valid usage by apps. We're just deprecating for internal usage within the library for structural and testability reasons.

gkc (approved):

PR #387, opened 2022-02-01
Refactor get

sitaram-kalluri:
**- What I did** Refactor get method for readability.

**- How I did it**

**- How to verify it**

### DIFF
--- a/at_client/lib/src/client/local_secondary.dart
+++ b/at_client/lib/src/client/local_secondary.dart
@@ -120,6 +120,9 @@ class LocalSecondary implements Secondary {
       if (builder.isCached) {
         llookupKey += 'cached:';
       }
+      if(builder.isPublic){
+        llookupKey += 'public:';
+      }
       if (builder.sharedWith != null) {
         llookupKey += '${AtUtils.formatAtSign(builder.sharedWith!)}:';
       }
--- a/at_client/lib/src/client/verb_builder_manager.dart
+++ b/at_client/lib/src/client/verb_builder_manager.dart
@@ -1,0 +1,80 @@
+import 'package:at_client/at_client.dart';
+import 'package:at_client/src/client/secondary.dart';
+import 'package:at_commons/at_builders.dart';
+import 'package:at_commons/at_commons.dart';
+import 'package:at_utils/at_utils.dart';
+
+/// Class responsible for returning the appropriate [VerbBuilder] for given [AtKey]
+class LookUpBuilderManager {
+  ///Returns a [VerbBuilder] for the given AtKey instance
+  static VerbBuilder get(AtKey atKey, String currentAtSign) {
+    // If isPublic is true in metadata, the key is a public key, return PLookupVerbHandler.
+    if (atKey.sharedBy != currentAtSign &&
+        (atKey.metadata != null &&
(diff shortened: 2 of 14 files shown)


at_client/lib/src/client/local_secondary.dart
```diff
@@ -120,6 +120,9 @@ class LocalSecondary implements Secondary {
       if (builder.isCached) {
         llookupKey += 'cached:';
       }
+      if(builder.isPublic){
+        llookupKey += 'public:';
+      }
       if (builder.sharedWith != null) {
         llookupKey += '${AtUtils.formatAtSign(builder.sharedWith!)}:';
       }
```
gkc, line 128: Are cached, public and sharedWith all mutually exclusive? If so, perhaps have some defensive code in here to check that at most *one* of these three booleans is true?

at_client/lib/src/util/at_client_validation.dart
```diff
@@ -52,4 +52,18 @@ class AtClientValidation {
...
   }
+
+  static void validateAtKey(AtKey atKey) {
+    // validates the atKey
+    validateKey(atKey.key);
+    // validates the metadata
+    validateMetadata(atKey.metadata);
+    // verifies if the sharedWith atSign exists.
+    if (atKey.sharedWith != null) {
+      isAtSignExists(
```
gkc, line 65: If disconnected from the network, isAtSignExists will throw an exception. In this context (validation), will the exception message make sense to a user if for example the app displays the message on screen?

at_client/lib/src/client/verb_builder_manager.dart
```diff
@@ -0,0 +1,80 @@
...
+
+class SecondaryManager {
+  static Secondary getSecondary(VerbBuilder verbBuilder) {
+    if (verbBuilder is LookupVerbBuilder ||
+        verbBuilder is PLookupVerbBuilder ||
+        verbBuilder is NotifyVerbBuilder ||
+        verbBuilder is StatsVerbBuilder) {
+      return AtClientManager.getInstance().atClient.getRemoteSecondary()!;
+    }
+    return AtClientManager.getInstance().atClient.getLocalSecondary()!;
```
gkc, line 81: Is there an explanation somewhere of which verbs need to be executed on local secondary vs remote secondary?

gkc (commented):
LGTM, apart from a few comments / requests

At some point in the future we need to come back and think about this whole `AtClientManager.getInstance()` business again :-)

at_client/lib/src/client/local_secondary.dart
sitaram-kalluri: public and sharedWith are mutually exclusive(not cached). However there can be a key without any of these(for exmaple self keys and encryption keys).

gkc (approved):

at_client/lib/src/util/at_client_validation.dart
sitaram-kalluri: Updated the code to perform this check only network connection is available.

at_client/lib/src/client/verb_builder_manager.dart
sitaram-kalluri: Updated the dart docs

PR #395, opened 2022-02-08
refactor put method and update functional test

sitaram-kalluri:
**- What I did** Refactor the put method and update the functional tests

**- How I did it**

**- How to verify it**


gkc (approved):
Looks great from a design and simplification perspective but the amount of change here makes it difficult to review for correctness - we're dependent on how good our tests are. So my vote is let's merge, alert everyone that we are committing this major change to trunk and would like the app developers (geekyants etc) to start using the trunk version so we can find any more potential bugs

PR #396, opened 2022-02-08
Switch atsign bug

### LINKED ISSUE #379: Not receiving notification after switching atsign
Lead: @kalluriramkumar 

**Describe the bug**
If we have onboarded with @ a , we receive notification.
On switching to @ b , we still receive notifications.
If we switch back to @ a , we stop receiving notification.

**To Reproduce**
1. onboard with any atsign (@ a)
2. switch to second atsign (@ b)
3. switch back to @ a and we stopped receiving notifications.

**Expected behavior**
we should receive notifications at @ a .

  - @ mosphere pro desktop

**Additional information**
On notification sender's side , we can see status as delivered.

sitaram-kalluri:
**- What I did** 1. Replace isEncrypted metadata with checking if key contains public: to decrypt the value.
2. Fix switch atsign bug -

**- How I did it**
2 - Switch atsign bug
Notification service impl was missing call to add changeListener and removing the instance of notification service of previous atsign. Fixed code to correct these
**- How to verify it**


gkc:
LGTM

murali-shris (approved):

gkc (approved):
LGTM

PR #399, opened 2022-02-10
sync progress listener

murali-shris:
**- What I did**
- Implemented sync progress callback

**- How I did it**
-  added an abstract call SyncProgressListener. UI widgets should implement this and add onSync implementation behavior
**- How to verify it**
- run the functional test atclient_sync_callback_test


at_client/lib/src/service/sync_service_impl.dart
```diff
@@ -160,9 +172,20 @@ class SyncServiceImpl implements SyncService, AtSignChangeListener {
     final serverCommitId = await _getServerCommitId();
     await _sync(serverCommitId, syncRequest);
     _syncComplete(syncRequest);
+    _updateProgress(SyncStatus.success, serverCommitId: serverCommitId);
     _syncInProgress = false;
   }
 
+  void _updateProgress(SyncStatus status, {int? serverCommitId}) {
```
VJag: informSyncProgress

at_client/lib/src/service/sync_service_impl.dart
```diff
@@ -30,6 +33,7 @@ class SyncServiceImpl implements SyncService, AtSignChangeListener {
   late final AtClient _atClient;
   late final RemoteSecondary _remoteSecondary;
   late final NotificationServiceImpl _statsNotificationListener;
+  final List<SyncProgressListener> _syncProgressListeners = [];
```
VJag: Let's make it a set to prevent duplicate listeners being added.

at_client/lib/src/service/sync_service_impl.dart
```diff
@@ -138,31 +157,53 @@ class SyncServiceImpl implements SyncService, AtSignChangeListener {
...
           ..syncStatus = SyncStatus.success
           ..lastSyncedOn = DateTime.now().toUtc()
           ..dataChange = false;
         _syncComplete(syncRequest);
         _syncInProgress = false;
+        syncProgress.syncStatus = SyncStatus.success;
+        _informSyncProgress(syncProgress);
         return;
       }
     } on AtLookUpException catch (e) {
```
VJag: Should we catch AtClientException or may be even a generic Exception?

at_client/lib/src/service/sync_service_impl.dart
murali-shris: done

at_client/lib/src/service/sync_service_impl.dart
murali-shris: https://github.com/atsign-foundation/at_client_sdk/commit/6161fc676f72723739bf4ffbff37e06b21876076
please check whether this is fine

at_client/lib/src/listener/sync_progress_listener.dart
```diff
@@ -0,0 +1,5 @@
+import 'package:at_client/src/service/sync/sync_status.dart';
+
+abstract class SyncProgressListener {
+  void onSync(SyncProgress syncProgress);
```
gkc: Please rename SyncProgressListener's `onSync` method to `onSyncProgressEvent`

gkc (changes requested):
Looks really good with just one request - please rename SyncProgressListener's `onSync` method to `onSyncProgressEvent`

at_client/lib/src/listener/sync_progress_listener.dart
murali-shris: done Gary.

VJag (approved):

at_client/lib/src/service/sync_service_impl.dart
```diff
@@ -30,6 +33,9 @@ class SyncServiceImpl implements SyncService, AtSignChangeListener {
   late final AtClient _atClient;
   late final RemoteSecondary _remoteSecondary;
   late final NotificationServiceImpl _statsNotificationListener;
+
+  /// static because once listeners are added, they should be agnostic to switch atsign event
+  static final Set<SyncProgressListener> _syncProgressListeners = HashSet();
```
gkc, line 38: Why is this static? the listeners wish to subscribe to a specific sync service

What is expected behaviour of sync progress listeners when atsign is switched?

gkc (changes requested):
Noticed static set of listeners in sync service interface

at_client/lib/src/service/sync_service_impl.dart
VJag: @gkc I have requested that change for the following reasons:

If we don't make it static we need:

1. Every switch @sign is followed by a call to SyncServiceImpl.addListener() too. Which needs some education for the app dev's.

2. Generally state is not specific to an instance of SyncServiceImpl

Please let me know your thoughts.

murali-shris: e.g alice registered a sync progress listener..app switches to bob. Now alice's listener is retained due to static nature.Otherwise caller has to register the listener again or we have to handle to logic to remember the registered listeners per atsign, when atsign is switched.

gkc: So the contract is that if I subscribe to sync progress events, I will get them for all atSigns - ok that makes sense (needs to be made clear)

How do I, the subscriber, know which atSign the event was for? Maybe we should add atSign into SyncProgressEvent ?

murali-shris: I have made the changes.

gkc: Thanks Murali, LGTM

gkc (approved):

PR #405, opened 2022-02-16
fix for isInSync bug

murali-shris:
**- What I did**
    Fixed a bug in sync where out of order responses was received on RemoteSecondary due to which format exception was thrown in SyncUtil.getLatestServerCommitId(...)

**- How I did it**
- added a new method _isInSync which will use the existing RemoteSecondary
- Made a change in current isInSync which apps call. Dedicated RemoteSecondary will be created and closed.

**- How to verify it**
- run isInSync in a loop with few seconds delay between calls in main method. No exception should be thrown and sync should complete successfully.
- functional test. TODO

### DIFF
--- a/at_client/lib/src/service/sync_service_impl.dart
+++ b/at_client/lib/src/service/sync_service_impl.dart
@@ -158,7 +158,7 @@ class SyncServiceImpl implements SyncService, AtSignChangeListener {
     }
     final syncRequest = _getSyncRequest();
     try {
-      if (await isInSync()) {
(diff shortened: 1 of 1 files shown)


gkc:
This PR really needs some description explaining what the problem was and why these changes fix it.

murali-shris:
> This PR really needs some description explaining what the problem was and why these changes fix it.

done Gary

gkc:
thanks Murali

sitaram-kalluri (approved):

PR #408, opened 2022-02-18
Added race condition test & runLocal.sh to help with local running

gkc:
* Added at_lookup_race_test.dart to prevent regression of https://github.com/atsign-foundation/at_libraries/issues/140
* Updated docker-compose.yaml to comply with version 2 docker-compose schema
* added runLocal.sh to run the functional tests locally

Added fix to prevent regression of https://github.com/atsign-foundation/at_libraries/issues/140

### DIFF
--- a/at_functional_test/runLocal.sh
+++ b/at_functional_test/runLocal.sh
@@ -1,0 +1,16 @@
+#!/bin/bash
+
+echo "Getting dependencies" && dart pub get
+
+cd test || exit
+
+echo "Starting docker-compose" && docker-compose up -d
+
+cd ..
+echo "Checking docker readiness" && dart run test/check_docker_readiness.dart
+
+echo "Executing pkamLoad" && docker exec test_virtualenv_1 supervisorctl start pkamLoad
+
+echo "Checking test environment" && dart run test/check_test_env.dart
+
+echo "Running tests" && dart test --concurrency=1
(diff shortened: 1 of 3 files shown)


at_functional_test/runLocal.sh
```diff
@@ -0,0 +1,16 @@
+#!/bin/bash
+
+echo "Getting dependencies" && dart pub get
+
+cd test || exit
+
+echo "Starting docker-compose" && docker-compose up -d
```
cpswan: Assumes user has permission to run `docker`. It might be worth having a test for the user being able to run docker. Or use sudo.

at_functional_test/test/at_lookup_race_test.dart
```diff
@@ -0,0 +1,167 @@
...
+    var atClient = atClientManager.atClient;
+
+    // To setup encryption keys
+    await setEncryptionKeys(atsign, preference);
+
+    Random random = Random();
+
+    bool syncComplete = false;
+    void onSyncDone(syncResult) {
+      logger.shout("******* HELLO!!!!! ********");
```
cpswan: Can we shout something more meaningful?

at_functional_test/runLocal.sh
```diff
@@ -0,0 +1,16 @@
+#!/bin/bash
+
+echo "Getting dependencies" && dart pub get
+
+cd test || exit
```
cpswan: Rather than cding around could we just use something like `docker compose -f test/docker-compose.yml up`?

cpswan (approved):

PR #419, opened 2022-02-25
Fixes git issue#418. Sync keys to remote secondary.

### LINKED ISSUE #418: Automatic sync not working on at_client: 3.0.10 / at_utils 3.0.6 onwards
**Describe the bug**
When setting a AtKey and putting a value using at_client: 3.0.10 / at_utils 3.0.6 onwards no client sync is happening so no notification is sent and the receiving @ sign does not know about the shared information

**To Reproduce**
1. Using the MWC demo code use the previous version of the libs and everything works as it should
2. Then I updated to the newer versions and everything stops working and with weird sync results
3.  Moving back cures the issues
4.  The MWC demo and Snackbar use auto notiifications and hence if someone builds them with the latest libs they will not work!

**Expected behavior**
Auto Synch should work as it did in previous builds

**Video**

https://user-images.githubusercontent.com/6131216/155606320-d148726f-9206-4df8-818c-65746050bdaf.mp4

 - Dart SDK version: 2.16.1 (stable) (Tue Feb 8 12:02:33 2022 +0100) on "windows_x64"

  - mwc demo

**Additional context**

sitaram-kalluri:
**- What I did** Fix keys not sync'ed from local to remote secondary #418 

**- How I did it** Refactor shouldSkipSync method to shouldSync.

**- How to verify it** The keys should be sync'ed to remote secondary.

### DIFF
--- a/at_client/lib/src/client/at_client_impl.dart
+++ b/at_client/lib/src/client/at_client_impl.dart
@@ -408,13 +408,8 @@ class AtClientImpl implements AtClient {
       ..dataSignature = metadata.dataSignature
       ..operation = UPDATE_META;
 
-    var isSyncRequired = true;
-    if (SyncUtil.shouldSync(updateKey!)) {
-      isSyncRequired = false;
-    }
-
     var updateMetaResult =
-        await getSecondary().executeVerb(builder, sync: isSyncRequired);
+        await getSecondary().executeVerb(builder, sync: SyncUtil.shouldSync(updateKey!));
     return updateMetaResult != null;
   }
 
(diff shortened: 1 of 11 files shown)


murali-shris:
Please add a functional test to put few entries and then verify localCommitId

gkc:
+1 re functional test

Also need an e2e test in at_client_sdk to verify also that shared keys are
received by the other at sign client

sitaram-kalluri:
@gkc and @murali-shris : Added functional tests that would verify sync behavior on local ahead and server ahead scenarios.

murali-shris (approved):

PR #420, opened 2022-02-25
Update encryption condition

sitaram-kalluri:
**- What I did** 

Fixed the issue of when a shared key is created using AtKey static factory method and sharedWith is populated with currentAtSign, then shared key encryption is performed instead of selk key encryption.

**- How I did it**

Added an IF condition to check if sharedWith atsign is currentAtSign when AtKey is instance of SharedKey.

**- How to verify it**


gkc (approved):
LGTM

sitaram-kalluri:
Due to Git performance issue, the tests didnt run. Closing the PR and opening a new one.

PR #428, opened 2022-03-02
rootDomain changed to prod

Intiserahmed:
**- What I did**

As We want developers to develop in prod environment I set the  default rootDomain in AtClientPreference "root.atsign.org" which was previously setted to "root.atsign.wtf" AKA Staging environtment.

**- How I did it**
Just changed the default line code 

**- How to verify it**


gkc (changes requested):
Looks good. However, please remove .DS_Store files and edit .gitignore to exclude them  as they shouldn't be checked in

gkc (approved):

PR #445, opened 2022-03-17
End2end tests

sitaram-kalluri:
**- What I did**

Add end 2 end tests in at_client_sdk which runs on PR and on merge to trunk branch.

Add end to end tests


at_end2end_test/config/config.yaml
```diff
@@ -0,0 +1,9 @@
+# The @Protocol root server configuration
+root_server:
+  # The url to connect to root server
+  url: 'root.atsign.wtf'
+
+# cicd atSign details
+atSign:
+  firstAtSign: '@cicd1'
+  secondAtSign: '@cicd2'
```
gkc, line 9: Remember that as soon as they are available we want to use different at signs than these ones which are also used for e2e test in at_server

sitaram-kalluri: Sure Gary, Once we have the new atsign, will replace these with the new ones.

sitaram-kalluri:
The build check doesn't seem to restart. Closing this PR and raising a new one.

PR #446, opened 2022-03-17
End2end tests

sitaram-kalluri:
**- What I did**
Add end to end tests.

**- How I did it**

**- How to verify it**


murali-shris (approved):

cpswan:
@kalluriramkumar can you explain the plan please (as it's not obvious to me from the code here so far)?

Do we need some new cicd atsigns?

sitaram-kalluri:
> @kalluriramkumar can you explain the plan please (as it's not obvious to me from the code here so far)?
> 
> Do we need some new cicd atsigns?

@cpswan : The plan is test the at_client methods on real atSign's to cover end2end functionality. For the time being, used cicd1 and cicd2 atSign's to ensure the correctness of the tests. Yes, we need new cicd atSign's which will replace the cicd1 and cicd2.

PR #459, opened 2022-04-01
rethrow decryption exception

murali-shris:
**- What I did**
- fixed returning NULL value from decrypt method when there is exception or error
**- How I did it**
- catch error and exception in decryption impl and rethrow

### DIFF
--- a/at_client/lib/src/decryption_service/local_key_decryption.dart
+++ b/at_client/lib/src/decryption_service/local_key_decryption.dart
@@ -26,12 +26,15 @@ class LocalKeyDecryption implements AtKeyDecryption {
     }
     try {
       return EncryptionUtil.decryptValue(encryptedValue, sharedKey);
-    } on Exception catch (e) {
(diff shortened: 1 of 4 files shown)


at_client/lib/src/decryption_service/local_key_decryption.dart
```diff
@@ -24,6 +24,14 @@ class LocalKeyDecryption implements AtKeyDecryption {
       _logger.severe('Decryption failed. SharedKey is null');
       throw KeyNotFoundException('Decryption failed. SharedKey is null');
     }
-    return EncryptionUtil.decryptValue(encryptedValue, sharedKey);
+    try {
+      return EncryptionUtil.decryptValue(encryptedValue, sharedKey);
```
VJag: Can the try catches be moved to "EncryptionUtil.decryptValue()"? If we can then the change will be in one place.

VJag (approved):

PR #472, opened 2022-04-06
Deprecate filebin methods

murali-shris:
**- What I did**
- deprecated filebin methods
**- How I did it**
- added deprecated annotation to uploadFile, downloadFile and reuploadFiles in at_client_spec

### DIFF
--- a/at_client/CHANGELOG.md
+++ b/at_client/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 3.0.16
+- Decrypt notification value in SDK
+- Support for shared key and public key checksum in notify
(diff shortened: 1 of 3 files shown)


sachins-geekyants (commented):
we can also mark shareFiles method as Deprecated.
we have moved it to app side.

murali-shris:
> we can also mark shareFiles method as Deprecated. we have moved it to app side.

thanks Sachin.Done

sitaram-kalluri (approved):

gkc:
Was there a ticket which is closed by this PR?

PR #476, opened 2022-04-07
resolving issue while using .env file with at_client_sdk example

### LINKED ISSUE #439: Cannot run at_client_mobile example app
**Describe the bug**
The at_client_mobile example app is not currently working due to it's missing env file.
Screenshot 2022-03-08 at 7 40 50 PM

**To Reproduce**
Need to add the .env file .

srieteja:
… add the .env file through README.md

closes https://github.com/atsign-foundation/at_client_sdk/issues/439

**- What I did**
Tweaked versions of kotlin, build tools and other dependencies to get the at_client_mobile example app working. Added instructions to create an env file through readme.md

### DIFF
--- a/at_client_mobile/README.md
+++ b/at_client_mobile/README.md
@@ -122,7 +122,7 @@ bool isOnboarded = atClientServiceMap.containsKey(atsign);
 print(isOnboarded); // Prints true if onboarded.
 ```
 
-- Creating the .env file
+- Format of the .env file
 
 ```yaml
 NAMESPACE = "at_client_demo"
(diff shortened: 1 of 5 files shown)


gkc (changes requested):
I would like more than just adding to the readme, if possible - ideally something that during build will alert the developer that they need a .env file, and point them to the readme

srieteja:
I have upgraded the versions of kotlin, build tools and at_app_flutter. The app now works smoothly with an env file present. @Intiserahmed has also reported the example app to be working properly now.  @gkc does the example app still need to assert the presence of an env file?

sitaram-kalluri (approved):

gkc:
> I have upgraded the versions of kotlin, build tools and at_app_flutter. The app now works smoothly with an env file present. @Intiserahmed has also reported the example app to be working properly now. @gkc does the example app still need to assert the presence of an env file?

If @Intiserahmed is happy with it then I think this is good enough - thanks @srieteja

gkc (approved):

PR #481, opened 2022-04-11
Monitor heartbeats and unit tests.

gkc:
**What I did**
* Monitor now sends keep-alive heartbeat messages
* Added Monitor unit tests, using mocktail.

**How I did it**
Monitor now sends heartbeats (noop:0 commands) periodically. Default interval between heartbeats is 10 seconds. Heartbeat response (normally 'ok\n') is expected within a third of the heartbeat interval. If a timely heartbeat response isn't received then the connection is closed, status set to stopped, and the retryCallback is called. (retryCallback is passed to the monitor at constructor time.)

First heartbeat is scheduled if the monitor is successfully started.
Subsequent heartbeats are scheduled every time a heartbeat is sent.
Heartbeats are only sent (and the next one scheduled) if the monitor status is still 'started'
Every time a heartbeat is sent, we also schedule the check to verify that a timely response has been received.

The _handleResponse function, which handles responses from the server, has had a couple of lines of code added to detect the heartbeat response.

Added optional parameter monitorHeartbeatInterval to the Monitor constructor. If not supplied, it uses the value from AtClientPreference.monitorHeartbeatInterval. Added unit tests for both cases.

Finally, removed call to remoteSecondary.isAvailable() as it is superfluous - basically all it does is open a socket to the secondary, which is what the Monitor is going to do anyway. No point in checking if we can open a socket in order to then ... open a socket.

Added unit tests to cover the heartbeat behaviour

**How to verify it**
All tests should pass

**Description for the changelog**
Notifications monitor now uses heartbeat messages to detect when the socket goes awol

### DIFF
--- a/at_client/CHANGELOG.md
+++ b/at_client/CHANGELOG.md
@@ -1,3 +1,8 @@
+## 3.0.17
+- Fix self encryption key not found
+- Fix for _getLastNotificationTime method returning null
+- Added heartbeats to Notifications Monitor to detect and recover from
+  dead socket. Heartbeat interval is customizable via AtClientPreference
 ## 3.0.16
 - Decrypt notification value in SDK
 - Support for shared key and public key checksum in notify
--- a/at_client/CHANGELOG.md
+++ b/at_client/CHANGELOG.md
@@ -3,6 +3,8 @@
 - Fix for _getLastNotificationTime method returning null
 - Added heartbeats to Notifications Monitor to detect and recover from
   dead socket. Heartbeat interval is customizable via AtClientPreference
+- Fix for os write permission issue: give app option to pass the path where
+  the encrypted file will be saved on disk
 ## 3.0.16
 - Decrypt notification value in SDK
 - Support for shared key and public key checksum in notify
--- a/at_client/lib/src/client/remote_secondary.dart
+++ b/at_client/lib/src/client/remote_secondary.dart
@@ -98,24 +98,30 @@ class RemoteSecondary implements Secondary {
     }, restartCallBack: _restartCallBack);
   }
 
+  Future<String?> findSecondaryUrl() async {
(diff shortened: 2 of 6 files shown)


gkc:
Hi @VJag and @kalluriramkumar - please review first thing. I've only done unit testing on this change, as I've run out of time today. (However the unit tests are pretty comprehensive) It would be really great if you could add a functional test to verify that it actually really works against the server - i.e. does not do anything weird, at least. Once you're reasonably happy that the change is good, then please let @sachins-geekyants know so he can build a version of the app against this branch (monitor_heartbeats_and_unit_tests)

gkc:
@kalluriramkumar even better, maybe you could build the app against this branch and see if it is working ok. (And then ping Sachin) You may wish to change the defaultHeartbeatInterval to 30 seconds, with response therefore expected within 10 seconds. That seems generous enough even for a very busy device talking over a slow network with very high latency

sitaram-kalluri:
@gkc : The server return the response with the prefix `data:`. Hence added `data:` prefix in handleMonitorResponse.

gkc:
Tested and confirmed working as intended by @kalluriramkumar and @VJag 
Tested and confirmed working as intended by @sachins-geekyants in https://github.com/atsign-foundation/apps/issues/574#issuecomment-1096882958

nickelskevin (approved):
as requested

gkc:
Thanks Kevin. Will bump version & edit change log  before merging, will then publish new version of the package. While I'm here I'll change the strict dependencies to "any after" for the various at_ packages this package depends on

gkc:
When I updated pubspec.yaml to make some dependencies "any after" rather than specific version, found a breaking change in at_commons. Filed https://github.com/atsign-foundation/at_tools/issues/179. Reverted to having strict dependencies in at_client for now.

PR #486, opened 2022-04-18
fix for issue#483

### LINKED ISSUE #483: end2end_tests in at_client_sdk are flaky
**Describe the bug**
Sometimes the end2end_tests fail. Sometimes they succeed.

**To Reproduce**
They're flaky so it's not consistent. However - noticing that end2end_tests had failed for #482, I re-ran them and they failed again. I re-ran them again, and they worked.

**Expected behavior**
The tests should either succeed or fail consistently.

purnimavenkatasubbu:
**- What I did**
Fixed the issue #483 
**- How I did it**
 Generating a new UUID for every test
**- How to verify it**
e2e tests should pass

### DIFF
--- a/at_end2end_test/test/sharing_key_test.dart
+++ b/at_end2end_test/test/sharing_key_test.dart
@@ -1,5 +1,5 @@
 import 'dart:io';
-import 'dart:math';
+import 'package:uuid/uuid.dart';
 
 import 'package:at_client/at_client.dart';
 import 'package:at_client/src/encryption_service/encryption_manager.dart';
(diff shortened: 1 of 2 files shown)


at_end2end_test/test/sharing_key_test.dart
```diff
@@ -51,20 +51,22 @@ void main() {
...
+    var value = '+91 901920192';
     // Setting currentAtSign atClient instance to context.
     await AtClientManager.getInstance().setCurrentAtSign(
         currentAtSign, namespace, TestUtils.getPreference(currentAtSign));
     var putResult =
-        await currentAtSignClientManager?.atClient.put(phoneKey, value);
+        await currentAtSignClientManager?.atClient.put(phoneNumberKey, value);
     expect(putResult, true);
     var isSyncInProgress = true;
     currentAtSignClientManager?.syncService.sync(onDone: (syncResult) {
```
VJag, line 72: A call to sync is not required for this test. Probably not harmful but not required.

VJag: The reason I mention this is that app devs are not expected to call sync at all. Sync should happen irrespectively.

VJag (approved):

VJag:
This PR is good, however, let's remove calls to sync when the expected result can come from the local secondary across any such test.

gkc (approved):

PR #492, opened 2022-04-22
At key impl

sitaram-kalluri:
**- What I did**
1. Uptake the new at_key validations to at_client_impl put method.

**- How I did it**
1. Added two new method for putText and putBinary.
2. Inside put method calls to respective put basing on value is deligated.
**- How to verify it**


at_client/lib/src/util/at_client_validation.dart
```diff
@@ -62,10 +62,32 @@ class AtClientValidation {
     validateMetadata(atKey.metadata);
     // verifies if the sharedWith atSign exists.
     if (atKey.sharedWith != null && await NetworkUtil.isNetworkAvailable()) {
-      isAtSignExists(
+      await isAtSignExists(
```
VJag, line 68: The validity of an atsign needs to be cached.

VJag: Also while notifying we are verifying "await AtClientUtil.findSecondary(atSign, rootDomain, rootPort);" This needs be cached too.

at_client/lib/src/util/at_client_validation.dart
```diff
@@ -62,10 +62,32 @@ class AtClientValidation {
...
           AtClientManager.getInstance().atClient.getPreferences()!.rootDomain,
           AtClientManager.getInstance().atClient.getPreferences()!.rootPort);
     }
   }
+
+  /// Performs the validations on the PutRequest
+  static void validatePutRequest(
+      AtKey atKey, dynamic value, AtClientPreference atClientPreference) {
+    // If length of value exceeds maxDataSize, throw AtClientException
+    if (value.length > atClientPreference.maxDataSize) {
```
VJag, line 79: We need to have this covered under our validation framework i.e make them extend an interface that can either return an error code or throw an exception.

at_client/lib/src/client/at_client_impl.dart
```diff
@@ -308,8 +310,69 @@ class AtClientImpl implements AtClient {
...
+    atKey.metadata ??= Metadata();
+    // Setting metadata.isBinary to true for putBinary
+    atKey.metadata!.isBinary = true;
+    // Base2e15.encode method converts the List<int> type to String.
+    return await _putInternal(atKey, Base2e15.encode(value));
+  }
+
+  Future<AtResponse> _putInternal(AtKey atKey, dynamic value) async {
+    // Performs the put request validations.
+    AtClientValidation.validatePutRequest(atKey, value, preference!);
```
VJag, line 355: I was thinking of making this a validator

VJag: 1. Who has the knowledge of throwing the right exceptions for the scenarios we have documented. (atClient or atValidator or Somethingelse)
2. Though encryption and decryption do not have any correlation to the validators, however, they do throw the exceptions. Is there an abstraction we can add around the throwing of exceptions

VJag (approved):

PR #512, opened 2022-05-13
Use dedicated atSigns for client e2e test

### LINKED ISSUE #511: Use the new dedicated at-signs which have been set up for at_client_sdk e2e tests
**Describe the bug**
New dedicated secondary servers have been set up for at_client_sdk end-to-end tests. See https://github.com/atsign-foundation/at_client_sdk/issues/447

However, those new at-signs are not yet being used by the at_client_sdk end-to-end tests.

**Expected behavior**
The new at-signs should be being used by the at_client_sdk end-to-end tests

gkc:
**- What I did**
Fixes https://github.com/atsign-foundation/at_client_sdk/issues/511

**- How I did it**
Updated e2e tests' config.yaml

**- How to verify it**
e2e tests should successfully run against the dedicated atSigns

Use dedicated atSigns for client e2e test


at_end2end_test/config/config.yaml
```diff
@@ -5,5 +5,9 @@ root_server:
...
 # cicd atSign details
 atSign:
-  firstAtSign: '@cicd1'
-  secondAtSign: '@cicd2'
+  # @ce2e1 is running latest at_secondary_server trunk
+  # @ce2e3 is running the latest canary release
+  # See https://github.com/atsign-foundation/at_client_sdk/issues/447#issuecomment-1075290415 for
+  # full details.
+  firstAtSign: '@ce2e1'
+  secondAtSign: '@ce2e3'
```
cpswan: This atsign is running on the same VM as firstAtSign, which isn't necessarily a problem, but past convention has been to test between VMs

gkc: OK ... chose 1 & 3 to get different secondary versions. What would you suggest the best pairings would be? How about 1 & 4, and 3 & 2?

cpswan: Yes, 1&4 and 3&2 will give a cross test of old vs new. I'm not quite sure why we're not using the same 3 way matrix as at_server, but this is what @kalluriramkumar asked for.

gkc: Paired ce2e1 and ce2e4

at_end2end_test/config/config.yaml
```diff
@@ -5,5 +5,9 @@ root_server:
 
 # cicd atSign details
 atSign:
-  firstAtSign: '@cicd1'
-  secondAtSign: '@cicd2'
+  # @ce2e1 is running latest at_secondary_server trunk
+  # @ce2e3 is running the latest canary release
```
cpswan: Just need to update this comment line to match

cpswan (changes requested):

cpswan (approved):

PR #513, opened 2022-05-14
Stack exceptions on get

sitaram-kalluri:
**- What I did**
1. Stack all the exceptions to form a chained exception for get method.

**- How I did it**
1. Add exceptions to a stack.

NOTE: Uncommented dependency overrides to enable functional tests to run successfully. Once the packages are published, will replace dependency overrides with hosted versions.


gkc (commented):
LGTM. However given that there are changes here not just to exception handling but also to code related to core get() and put() methods, and recently some of the exception handling code was reverted from trunk via commit, I think it's important for @murali-shris to review this PR.

Additionally, please pair with someone when you get to resolving any merge conflicts.

Finally I'll reiterate the comment I made on a related PR, that this will need extensive testing with existing apps once merged to trunk.

at_client/lib/src/client/at_client_impl.dart
```diff
@@ -308,19 +325,68 @@ class AtClientImpl implements AtClient {
...
+    atKey.metadata!.isBinary = true;
+    // Base2e15.encode method converts the List<int> type to String.
+    return await _putInternal(atKey, Base2e15.encode(value));
+  }
+
+  Future<AtResponse> _putInternal(AtKey atKey, dynamic value) async {
+    // Performs the put request validations.
+    AtClientValidation.validatePutRequest(atKey, value, preference!);
     // Set sharedBy to currentAtSign if not set.
     if (atKey.sharedBy == null || atKey.sharedBy!.isEmpty) {
```
murali-shris: we can have a private method to check a string is null or empty..we have this check in multiple places

at_client/lib/src/client/at_client_impl.dart
```diff
@@ -231,25 +236,37 @@ class AtClientImpl implements AtClient {
...
+      }
+      // Send AtKey and AtResponse to transform the response to AtValue.
+      var getResponseTuple = Tuple<AtKey, String>()
+        ..one = atKey
+        ..two = (getResponse);
+      // Transform the response and return
+      var atValue = await GetResponseTransformer().transform(getResponseTuple);
+      return atValue;
+    } on AtException catch (e) {
+      var exceptionScenario = ExceptionScenario.localVerbExecutionFailed;
```
murali-shris: can we simplify these 3 lines using ternary operator?

at_client/lib/src/client/at_client_impl.dart
```diff
@@ -308,19 +325,68 @@ class AtClientImpl implements AtClient {
   @override
   Future<bool> put(AtKey atKey, dynamic value,
       {bool isDedicated = false}) async {
-    // Perform atKey validations.
-    await AtClientValidation.validateAtKey(atKey);
+    AtResponse atResponse = AtResponse();
+    // If the value is neither String nor List<int> throw exception
+    if (value is! String && value is! List<int>) {
```
murali-shris, line 329: @VJag in put method spec..we allow dynamic value..but in the implementation we check the type to be either string or binary. Is this fine? should the spec be modified to mention only string and binary values are allowed in put?

sitaram-kalluri: @murali-shris : My reasoning on checking if the value is String or List<int>: Though the older implementation of put method has dynamic, i believe is to support String and List<int> only. The encryption/decryption logic takes input of type only String (For binary data we are encoding to change list<int> to String and vice versa). We do not have parsing logic to convert other data types to String. Further, the data field in AtData object is of type String. Hence restricted the putText to String and putBinary to List<int>. Please let me know your thoughts on this.

at_client/lib/src/client/at_client_impl.dart
sitaram-kalluri: Implemented isNull method to verify if a field is null or empty

at_client/lib/src/client/at_client_impl.dart
sitaram-kalluri: Updated the code to use ternary operators

murali-shris (approved):

Issue #508, opened 2022-05-12
_getLastNotificationTime throws FormatException


sachins-geekyants:
**Describe the bug**
Data stored in _latestNotificationId is not in encrypted form, which throws formatException while onboarding.
This issue was seen on 0living atsign.

**To Reproduce**
1. onboard with 0living in wavi
2. when sync completes restart the app.
3. we should get an exception and onboarding will fail

**Screenshots**
<img width="1076" alt="Screenshot 2022-05-10 at 11 41 38 AM" src="https://user-images.githubusercontent.com/74917871/168003216-2d7c8630-0d08-45c9-906e-723e1c64e776.png">

**Smartphone (please complete the following information):**
 - macOS

**Were you using an @‎application when the bug was found?**
  - wavi desktop

**Additional context**
0living atKey and .hive files have been shared with Murali.

murali-shris:
fixed the issue in fix_lastNotificationId_bug branch. this branch also has some of the changes that have to be reverted from trunk that were merged for exception handling

murali-shris:
Some of the changes related to exception handling and at key validation had to  be reverted. Purnima verified the changes for 0living atsign. Fix is working fine. Have requested Sachin to verify the fix from his end as well.

sachins-geekyants:
@murali-shris I tried running with `fix_lastNotificationId_bug` at_client branch with 0living, the data in _latestNotificationId key was updated with some encrypted value.

But, I again got FormatException in jsonDecode
> if (atValue != null && atValue.value != null) {
        _logger.finer('json from hive: ${atValue.value}');
        return jsonDecode(atValue.value)['epochMillis'];
      }

I looked at the value in AtValue, it was in an encrypted format which failed at jsonDecode.

gkc:
@murali-shris is your branch ready for @sachins-geekyants to re-test with?

murali-shris:
> @murali-shris is your branch ready for @sachins-geekyants to re-test with?

I have just passed the  branch - last_notificationid_bug to Sachin. I have renamed the latest notification id key in my latest fix

sachins-geekyants:
Working fine with the latest fix.

murali-shris:
Published at_client 3.0.23 and at_client_mobile 3.1.15

PR #514, opened 2022-05-16
Last notificationid bug

murali-shris:
**- What I did**
- Fixed lastNotificationTime issue that happens while trying to decrypt old data
- Reverted the changes related to AtKey Validations and depricating AtClientException changes from trunk

**- How I did it**
- in notification_service_impl --> _getLastNotificationTime..handle the exception that happens while decrypting plain data. Get the unencrypted value from secondary server, encrypt and save it back to local secondary
- Revert the commits related to At Key validations and deprecating AtClientException changes.

**- How to verify it**
- onboard any atsign in wavi or atmosphere pro which is affected by issue https://github.com/atsign-foundation/at_client_sdk/issues/508
- atsign should be onboarded successfully without crashing the app


gkc (approved):
_latestNotificationIdv2 change looks good. I have not checked all of the reverting of the atKey validation and exceptions - so the branch will require full re-testing of apps to ensure that no other problems have been caused by that.

murali-shris:
> _latestNotificationIdv2 change looks good. I have not checked all of the reverting of the atKey validation and exceptions - so the branch will require full re-testing of apps to ensure that no other problems have been caused by that.

sanity check has been done on this branch on buzz, wavi and atmosphere pro

PR #524, opened 2022-05-19
Full Regular Expression matching for notificationService.subscribe

### LINKED ISSUE #523: NotificationService does not match regular expressions
**Describe the bug**
Notification Service uses String.allMatches().isNotEmpty which only checks for a match against the exact string.
Something like `.*` will not work with `notificationService.subscribe(regex: '.*')`

**To Reproduce**
1. Call `notificationService.subscribe(regex: '.*${AtEnv.appNamespace}.*')` in an app
2. Send a notification to that atsign
3. And check to see if notifications are received (they aren't)

**Expected behavior**
That regular expressions work for the regex parameter.

**Screenshots**
N/A

N/A

N/A

**Additional context**
Opening a PR with a fix using `RegExp().hasMatch()`

xavierchanth:
closes #523 

**- What I did**

- Replaced exact string match (`String.allMatches().isNotEmpty`) with `RegExp().hasMatch()`

**- How I did it**

- Local Changes

**- How to verify it**

Full Regular Expression matching for notificationService.subscribe

### DIFF
--- a/at_client/lib/src/service/notification_service_impl.dart
+++ b/at_client/lib/src/service/notification_service_impl.dart
@@ -10,6 +10,7 @@ import 'package:at_client/src/preference/monitor_preference.dart';
 import 'package:at_client/src/response/default_response_parser.dart';
 import 'package:at_client/src/response/notification_response_parser.dart';
 import 'package:at_client/src/service/notification_service.dart';
+import 'package:at_client/src/util/regex_match_util.dart';
 import 'package:at_commons/at_builders.dart';
 import 'package:at_commons/at_commons.dart';
 import 'package:at_utils/at_logger.dart';
@@ -139,8 +140,7 @@ class NotificationServiceImpl
                 await _getDecryptedNotifications(atNotification);
           }
           if (notificationConfig.regex != emptyRegex) {
-            if (RegExp(notificationConfig.regex)
-                .hasMatch(atNotification.key)) {
+            if (hasRegexMatch(atNotification.key, notificationConfig.regex)) {
               streamController.add(atNotification);
             }
           } else {
--- a/at_client/lib/src/service/notification_service_impl.dart
+++ b/at_client/lib/src/service/notification_service_impl.dart
@@ -139,9 +139,8 @@ class NotificationServiceImpl
                 await _getDecryptedNotifications(atNotification);
           }
           if (notificationConfig.regex != emptyRegex) {
-            if (notificationConfig.regex
-                .allMatches(atNotification.key)
-                .isNotEmpty) {
+            if (RegExp(notificationConfig.regex)
+                .hasMatch(atNotification.key)) {
               streamController.add(atNotification);
             }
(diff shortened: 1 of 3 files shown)


XavierChanth:
Waiting for the test secondaries to be updated by shepherd before rerunning the tests.

gkc:
Our cicd test images may be a bit borked, hence test failures. We are investigating.

gkc (approved):
LGTM. As with all such changes though, some current usages may well depend on the existing incorrect behaviour. Please test accordingly 😀

xavierchanth:
Is something like this suitable proof:
```dart
void main() async {
  String key = '@interiorwolverine:phone.streamprovider@snowboardingbeaver87';
  String regex;

  regex = '@interiorwolverine:phone.streamprovider@snowboardingbeaver87';
  print(originalQuery(regex, key)); // TRUE
  print(newQuery(regex, key)); // TRUE

  regex = 'phone';
  print(originalQuery(regex, key)); // TRUE
  print(newQuery(regex, key)); // TRUE

  regex = 'phone.streamprovider';
  print(originalQuery(regex, key)); // TRUE
  print(newQuery(regex, key)); // TRUE

  regex = '.*phone';
  print(originalQuery(regex, key)); // FALSE
  print(newQuery(regex, key)); // TRUE
}

bool originalQuery(String regex, String key) {
  return regex.allMatches(key).isNotEmpty;
}

bool newQuery(String regex, String key) {
  return RegExp(regex).hasMatch(key);
}
```

The previous only matched for exact string values, which this still does.

Should I implement a test for it somewhere?

gkc:
Ah OK, yeah this is fine. Yes ideally you should add a test somewhere. You could add a few simple unit tests just for this by breaking out the two relevant lines of code into their own method in NotificationServiceImpl and then just writing tests for that method. Or something else - up to you!

gkc:
👍👍 nice

gkc:
@XavierChanth please merge the latest changes from atsign-foundation/at_client_sdk/trunk into your branch ... those changes should resolve the failing checks

xavierchanth:
@gkc Do the CICD credentials need to be updated? It looks like they were 4 days ago.

gkc:
@XavierChanth yep. the move to using a new set of atSigns for testing dedicated to this repo has been a bit painful. I'm just merging those changes into trunk as well now. I'll ping you when done.

gkc:
@XavierChanth all right ... #512 has been merged to trunk. Please merge from this trunk into your trunk again - thanks!

xavierchanth:
Is this an issue with secrets @gkc ?

gkc:
@XavierChanth yes it's a secrets issue. We need to configure the actions so we only run the 'unit tests' check (which also actually includes the functional tests, we need to split it up into two checks) for pull requests from external branches.

xavierchanth:
I see, I could first push to a base branch, then pr to trunk.

PR #530, opened 2022-05-23
Full Regular Expression matching for notificationService.subscribe

### LINKED ISSUE #523: NotificationService does not match regular expressions

xavierchanth:
closes #523 

**- What I did**

- Replaced exact string match (`String.allMatches().isNotEmpty`) with `RegExp().hasMatch()`

**- How I did it**

- Local Changes

**- How to verify it**

Full Regular Expression matching for notificationService.subscribe

### DIFF
--- a/at_client/lib/src/service/notification_service_impl.dart
+++ b/at_client/lib/src/service/notification_service_impl.dart
@@ -10,6 +10,7 @@ import 'package:at_client/src/preference/monitor_preference.dart';
 import 'package:at_client/src/response/default_response_parser.dart';
 import 'package:at_client/src/response/notification_response_parser.dart';
 import 'package:at_client/src/service/notification_service.dart';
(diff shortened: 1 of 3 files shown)


xavierchanth:
@gkc did a PR from base repo instead, same as #524.

gkc (approved):
LGTM

PR #532, opened 2022-05-23
Fix modification of AtKey objects when calling notify with namespace aware keys.

### LINKED ISSUE #527: Namespace is appended to AtKey.key when using notify
**Describe the bug**
NotificationParams uses the original reference to AtKey.
In AtClientImpl.notifyChange, the original reference is modified which means the key cannot be reused after notify has been called.

**To Reproduce**
1. Create a Key
2. Notify using the notificationService.notify
3. Wait for the notify to occur
4. Check the key

**Expected behavior**
The AtKey instance should be unmodified after notify.

This could be achieved by:

- Creating a new AtKey in notifyChange to pass into notify
- Creating a clone of the AtKey in the NotificationParams constructor (the simpler solution)

**Additional context**
A temporary solution is to create a deep clone of the AtKey instance to notify with:
image

xavierchanth:
fixes #527 

**- What I did**

Created a copy of notificationParams.atKey.key which can be safely modified and passed to the NotifyVerbBuilder.

**- How I did it**

Local changes.

**- How to verify it**

See the notify tests.

Fix modification of AtKey objects when calling notify with namespace aware keys.

### DIFF
--- a/at_client/lib/src/client/at_client_impl.dart
+++ b/at_client/lib/src/client/at_client_impl.dart
@@ -704,6 +704,8 @@ class AtClientImpl implements AtClient {
 
   @override
   Future<String?> notifyChange(NotificationParams notificationParams) async {
+    String? notifyKey = notificationParams.atKey.key;
+
     // Check for internet. Since notify invoke remote secondary directly, network connection
     // is mandatory.
     if (!await NetworkUtil.isNetworkAvailable()) {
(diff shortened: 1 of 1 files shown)


xavierchanth:
Creating a copy constructor for AtKey or manually creating a deep copy of the AtKey is an alternative approach, however only copying the string should consume less memory, hence why I went with the latter approach.

gkc (approved):
LGTM

PR #533, opened 2022-05-26
handle empty sync response

murali-shris:
**- What I did**
- handle empty sync responses in at client
**- How I did it**
- in sync service impl, when server returns no response then break the loop to sync from local to remote.
**- How to verify it**
- have an atsign using two different apps with namespaces e.g wavi and atmospherepro
- run few updates in atmospherepro.. commit id on server gets incremented
- login to wavi app. Now new commits will be tried to sync to the app. but since the namespace doesn't match empty response will be sent by the server. After the fix , atsign should be able to login successfully
- handle empty sync responses from server in sync service

### DIFF
--- a/at_client/lib/src/service/sync_service_impl.dart
+++ b/at_client/lib/src/service/sync_service_impl.dart
@@ -274,7 +274,8 @@ class SyncServiceImpl implements SyncService, AtSignChangeListener {
         atSign: _atClient.getCurrentAtSign()!);
     var localCommitId = await _getLocalCommitId();
     if (serverCommitId > localCommitId) {
-      _logger.finer('syncing to local');
+      _logger.finer(
+          'syncing to local: localCommitId $localCommitId serverCommitId $serverCommitId');
       await _syncFromServer(serverCommitId, localCommitId);
     }
     if (unCommittedEntries.isNotEmpty) {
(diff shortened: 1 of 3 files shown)


VJag (approved):

at_client/lib/src/service/sync_service_impl.dart
```diff
@@ -339,9 +340,22 @@ class SyncServiceImpl implements SyncService, AtSignChangeListener {
           .parse(await _remoteSecondary.executeVerb(syncBuilder))
           .response);
       _logger.finest('** syncResponse $syncResponseJson');
+
+      if (syncResponseJson == null || syncResponseJson.isEmpty) {
+        _logger.finer(
+            'sync response is empty: local commitID: $localCommitId server commitID: $serverCommitId');
+        break;
```
gkc, line 347: When this occurs, I guess the local commit id remains at a lower value than server commit id. I don't think this is a problem but can you confirm please?

gkc (changes requested):

at_client/lib/src/service/sync_service_impl.dart
murali-shris: yes..two possibilities after this
1. subsequent update on app /server will trigger the next sync.
2. If nothing happens on app or server, this condition will be true until the next update happens.

gkc: OK... so when we get the stats notification via the monitor, that will trigger another sync every time (because the server will always be 'ahead'). This isn't necessarily a problem, just will get some technically unnecessary syncs happening. Is this OK?

murali-shris: yes Gary..This shouldn't be a problem I guess..another alternative to solve the issue is remembering the lastServer commitID and then trigger _syncFromServer only when new server commit id is greater than last server commit Id

gkc: I think a little extra traffic is no big deal. Let's go with this as it is for now. Thanks Murali

gkc (approved):

PR #541, opened 2022-06-03
latest commit id - handle no response or exception from server

murali-shris:
**- What I did**
- when server returns exception or no response, throw exception
**- How I did it**
- added Exception block to SyncUtil.getLatestServerCommitId. rethrow AtClientException
**- How to verify it**
- run sync from at client when server is not reachable
- client should run gracefully by calling the sync error call back.


at_client/lib/src/util/sync_util.dart
```diff
@@ -88,6 +91,10 @@ class SyncUtil {
     } on AtClientException catch (e) {
       logger.severe(
           'Exception occurred in processing stats verb ${e.errorCode} - ${e.errorMessage}');
+      rethrow;
+    } on Exception catch (e) {
+      throw AtClientException(atClientErrorCodes['AtClientException'],
```
VJag: Let's add a log statement.

VJag (approved):

sitaram-kalluri (approved):

at_client/lib/src/util/sync_util.dart
murali-shris: done

PR #543, opened 2022-06-06
throw AtPublicKeyNotFound when current atsign public key is not found…

### LINKED ISSUE #542: Incorrect exception thrown for missing public key of current atsign on decryption side
**Describe the bug**
The exception message is incorrect for the missing public key of the current atsign on the decryption side( receiver's end).

**To Reproduce**
1. log in to atsign A and atsign B
2. Then share a key from atsign B to atsign A
3. And then delete the public key on the atsignA

**Expected behavior**
An apparent exception message as "current atsigns's public key not found" should be shown

sitaram-kalluri:
…. Added unit tests

**- What I did**
1. Fixed the issue [542](https://github.com/atsign-foundation/at_client_sdk/issues/542) and added unit test
2. Added unit test for AtPrivateKeyNotFound Exception
3. Moved the unit tests from at_exception_test file to decryption_service_test.dart file.

**- How I did it**
Catch `KeyNotFoundException` and throw `AtPublicKeyNotFoundException`.

**- How to verify it**
Remove the encryption public key from the current atsign and run a get. While decrypting AtPublicKeyNotFoundException is thrown.


at_client/lib/src/decryption_service/shared_key_decryption.dart
```diff
@@ -31,10 +31,18 @@ class SharedKeyDecryption implements AtKeyDecryption {
...
-              .getEncryptionPublicKey(atClient!.getCurrentAtSign()!))
-          ?.trim();
+      String? currentAtSignPublicKey;
+      try {
+        currentAtSignPublicKey = (await atClient
+                ?.getLocalSecondary()!
+                .getEncryptionPublicKey(atClient!.getCurrentAtSign()!))
+            ?.trim();
+      } on KeyNotFoundException {
+        throw AtPublicKeyNotFoundException(
```
murali-shris, line 41: @kalluriramkumar @VJag 
should we try to look up public key from server before throwing AtPublicKeyNotFoundException ?

sitaram-kalluri: @murali-shris : The encryptionPublicKey is saved to localSecondary from .keys file and then it is sync'ed to remote secondary. So it is highly unlikely that key exist in remote secondary without its presence in local secondary.

VJag (approved):
LGTM

PR #545, opened 2022-06-08
Fix sync issue on atsign reset

sitaram-kalluri:
**- What I did**
1. Sync the local changes to server when server is reset.
2. Refactor the sync_service_impl.dart file to write unit test
3. Write unit test for server reset scenario.

NOTE: **sync_service_impl supersedes the sync_manager.dart. The changes to sync_manager are made only to avoid compile time issue. sync_manager.dart is not in use.

**- How I did it**
If localCommitId is greater than serverCommitId indicates server is reset (In other conditions localCommitId is either equal or less than serverCommitId). then set localSeqNumber to serverCommitId to get all the changes to sync to server.

**- How to verify it**
Pair the atsign in mobile app and reset the server. Upon activation the data should be sync'ed back to server.

### DIFF
--- a/at_client/lib/src/manager/at_client_manager.dart
+++ b/at_client/lib/src/manager/at_client_manager.dart
@@ -1,7 +1,6 @@
 import 'package:at_client/at_client.dart';
 import 'package:at_client/src/listener/at_sign_change_listener.dart';
 import 'package:at_client/src/listener/switch_at_sign_event.dart';
-import 'package:at_client/src/service/notification_service.dart';
 import 'package:at_client/src/service/notification_service_impl.dart';
 import 'package:at_client/src/service/sync_service.dart';
 import 'package:at_client/src/service/sync_service_impl.dart';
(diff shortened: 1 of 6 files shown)


gkc:
I don't think we should proceed with this approach.
1) We can have a backup automatically taken of a server's data as part of the deletion process, which can then be automatically recovered
2) The approach in this PR will not work for multiple clients each with fragments of the data due to commit ID clashes. We could try to have the server mediate all of this and go into a 'recovery' mode but this is very very complex
3) If we really want to have automatic reconstruction of server data from clients (e.g. in the event of catastrophic data loss on the server followed by recovery from backup up to 24 hours old) then that is _far_ more easily done with a sync-via-timestamp rather than sync-via-CommitID approach

gkc (changes requested):
See [this comment](https://github.com/atsign-foundation/at_client_sdk/pull/545#issuecomment-1149686341)

sitaram-kalluri:
@gkc : Can you please let me know if we can close this PR?

gkc:
@kalluriramkumar : Please close whenever you wish. @VJag has started a document to talk about desired behaviour in all of the different possible reset / data loss scenarios. That document should include a reference to this PR.

sitaram-kalluri:
> @kalluriramkumar : Please close whenever you wish. @VJag has started a document to talk about desired behaviour in all of the different possible reset / data loss scenarios. That document should include a reference to this PR.

Sure Gary, Thank you.

PR #549, opened 2022-06-13
Add exception chaining to self key encryption

sitaram-kalluri:
**- What I did**
1. Add exception chain when SelfEncryptionKey not found.
2. Add unit test for positive and negative scenario of Self Key Encryption.
3. Add NullCheck extension on String to check if String is null or empty.

**- How I did it**

**- How to verify it**

### DIFF
--- a/at_client/lib/src/encryption_service/self_key_encryption.dart
+++ b/at_client/lib/src/encryption_service/self_key_encryption.dart
@@ -1,6 +1,8 @@
+import 'package:at_client/src/client/local_secondary.dart';
 import 'package:at_client/src/manager/at_client_manager.dart';
 import 'package:at_client/src/encryption_service/encryption.dart';
 import 'package:at_client/src/response/default_response_parser.dart';
+import 'package:at_client/src/util/at_client_util.dart';
 import 'package:at_client/src/util/encryption_util.dart';
 import 'package:at_commons/at_commons.dart';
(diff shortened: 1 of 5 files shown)


at_client/lib/src/encryption_service/self_key_encryption.dart
```diff
@@ -9,35 +11,45 @@ import 'package:at_utils/at_logger.dart';
...
     try {
-      // Get AES key for current atSign
-      var selfEncryptionKey = await _getSelfEncryptionKey();
-      if (selfEncryptionKey == null ||
-          selfEncryptionKey.isEmpty ||
-          selfEncryptionKey == 'data:null') {
-        throw KeyNotFoundException(
-            'Self encryption key is not set for atSign ${atKey.sharedBy}');
+      selfEncryptionKey = await localSecondary.getEncryptionSelfKey();
+      if (selfEncryptionKey.isNull || selfEncryptionKey == 'data:null') {
```
murali-shris: local secondary getEncryptionSelfKey has a call to keystore.get..keystore.get will not return data:null.. do we need the check?

sitaram-kalluri: Yeah, the 'data:null' check can be ignored. I will remove the condition and push the changes.

murali-shris (approved):

PR #555, opened 2022-06-14
fix: add unit test for sign public key data and getKeyWithNamespace

sitaram-kalluri:
**- What I did**
1. Add unit test for sign public key data
2. Add unit tests for appending namespace to the key

**- How I did it**

**- How to verify it**

### DIFF
--- a/at_client/lib/src/util/at_client_util.dart
+++ b/at_client/lib/src/util/at_client_util.dart
@@ -113,26 +113,23 @@ class AtClientUtil {
   /// Accepts [AtKey] and returns [AtKey.key] with namespace appended
   /// Appends namespace if [atKey.metadata.namespaceAware] is set to true,
   /// else namespace is not appended
-  static String getKeyWithNameSpace(AtKey atKey) {
+  static String getKeyWithNameSpace(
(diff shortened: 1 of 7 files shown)


murali-shris (approved):

at_client/lib/src/util/at_client_util.dart
```diff
@@ -113,26 +113,23 @@ class AtClientUtil {
   /// Accepts [AtKey] and returns [AtKey.key] with namespace appended
   /// Appends namespace if [atKey.metadata.namespaceAware] is set to true,
   /// else namespace is not appended
-  static String getKeyWithNameSpace(AtKey atKey) {
+  static String getKeyWithNameSpace(
+      AtKey atKey, AtClientPreference atClientPreference) {
```
murali-shris, line 117: since are using only namespace can we replace atClientPreference with namespace?

sitaram-kalluri: @murali-shris : Since the namespace in AtClientPreference is `String?`, instead of checking isNull in all the caller methods (though namespace in AtClientPreference going is most unlikely), i have passed AtClientPreferences and checking for isNull here.

PR #590, opened 2022-07-06
feat:support for optional decryptPackets parameter

### LINKED ISSUE #189: Update TLS connection to optionally output TLS keys to file
Update TLS connection to optionally output TLS keys to file. This allows you to "see" inside the TLS packets using WireShark and diagnose issues.

**Describe the solution you'd like**
Open to suggestions on implementation but it would be nice just to include a dev low level library that includes this feature. So testing can be done by including a dev library in pubspec.yaml. This would dump the TLS keys in the directory where the binary is being run.

The additional lines of code required to do this are :- 

In at_lookup_impl.dart

```
      var secConConnect = SecurityContext();
       var keyfile = File('keysfile');
       secConConnect.setTrustedCertificates('caroot/rootcacert.pem');
      var secureSocket = await SecureSocket.connect(host, int.parse(port), context: secConConnect, keyLog: (line) => keyfile.writeAsStringSync(line, mode: FileMode.append));

```

And in monitor_client.dart

```
var secConConnect = SecurityContext();
       var keyfile = File('keysfile');
       secConConnect.setTrustedCertificates('caroot/rootcacert.pem');
      var secureSocket = await SecureSocket.connect(host, int.parse(port), context: secConConnect, keyLog: (line) => keyfile.writeAsStringSync(line, mode: FileMode.append));
```

replacing the secureSocket connection with no SecurityContent()

It would be nice to abstract the SecureSocket.connect so only one change would effect both lines of code and then that abstraction could be used in the secondary server code as well perhaps.

**Describe alternatives you've considered**
I did consider pushing all the way through via command line options or by adding a method options but that I think holds the danger of leaving it in place before going to a prod build.. But open to them or other ideas.. 

**Additional context**
Screen shot of the resulting Wireshark diagnostics

image

srieteja:
closes https://github.com/atsign-foundation/at_libraries/issues/189
**- What I did**
changes to support optional decryptPackets parameter
**- How I did it**
added new param in at_client_preference called "decryptPackets" that will enable this feature.
using the SecureSocketUtil from at_lookup to create a new outbound connection
if decryptPackets is enabled then the secure socket logs keys into a file
**- How to verify it**


srieteja:
Debugged the cause of failing tests. Need to modify some unit tests that mock MonitorOutboundConnectionFactory class to accommodate the new approach to creating Secure Sockets.

cconstab (approved):
Looks good to me

PR #591, opened 2022-07-07
feat: bypass cache option in get method

murali-shris:
**- What I did**
 * added a feature to bypass cache in at client get method
**- How I did it**
* introduced a new class GetRequestOptions with byPassCache attribute
* Added GetRequestOptions as optional param to at client get method
* modified GetRequestTransformer and PutRequestTransformer
* introduced new abstract class RequestTransformer which extends Transformer and adds an optional RequestOptions
**- How to verify it**
* Run the group of  tests - "A group of tests to check bypass cache flag" in verb_builder_test.dart


murali-shris:
checking the test failures

sitaram-kalluri (approved):

murali-shris:
> checking the test failures



> checking the test failures



> checking the test failures

fixed

gkc (changes requested):
This is a bit nit-picky, but please change all variable names which have 'byPass' to 'bypass' as bypass is the actual word

e.g. in GetRequestOptions, the bool should be `bypassCache` rather than `byPassCache`

murali-shris:
> This is a bit nit-picky, but please change all variable names which have 'byPass' to 'bypass' as bypass is the actual word
> 
> e.g. in GetRequestOptions, the bool should be `bypassCache` rather than `byPassCache`

changes done

gkc (approved):

PR #602, opened 2022-07-13
fix: do not sync cached keys to cloud secondary

sitaram-kalluri:
**- What I did**
1. Modify the sync logic to skip the cached keys being sync'ed to cloud secondary.

**- How I did it**
1. Add an IF condition what checks if key starts with `cached:`. If yes, ignores the key to sync-batch request that is sent to cloud secondary.

**- How to verify it**

### DIFF
--- a/at_client/lib/src/service/sync_service_impl.dart
+++ b/at_client/lib/src/service/sync_service_impl.dart
@@ -325,7 +325,7 @@ class SyncServiceImpl implements SyncService, AtSignChangeListener {
             var commitEntry = unCommittedEntryList.elementAt(batchId - 1);
             if (commitId == -1) {
               _logger.severe(
-                  'update/delete for key ${commitEntry.atKey} failed. Error code ${responseObject.errorCode} error message ${responseObject.errorMessage}');
+                  '${commitEntry.operation} for key ${commitEntry.atKey} failed. Error code ${responseObject.errorCode} error message ${responseObject.errorMessage}');
             }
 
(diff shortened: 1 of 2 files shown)


at_client/lib/src/service/sync_service_impl.dart
```diff
@@ -395,6 +395,12 @@ class SyncServiceImpl implements SyncService, AtSignChangeListener {
     var batchId = 1;
     for (var entry in uncommittedEntries) {
       String command;
+      //Skipping the cached keys to sync to cloud secondary.
```
gkc, line 398: 1) Why are we skipping here rather than when we are getting the list of uncommitted entries?
2) What happens to the uncommitted entries which are for "cached:"
3) Should we actively prevent updates via LocalSecondary to entries which are "cached:" ?

gkc (changes requested):

at_client/lib/src/service/sync_service_impl.dart
gkc: After discussion:

1: It's fine to skip here
2: Handled by logic around lastSyncedEntryCacheMap and SyncUtil.getLastSyncedEntry; no issue here, we won't get these uncommitted entries again

3. We should add logic to AtClientImpl.put() which checks key structure (starts with 'cached:') as well as what it currently does, which is check metadata.isCached

sitaram-kalluri:
The changes are merged in PR 603

PR #616, opened 2022-07-20
feat: Refraining syncing of stats notification id

### LINKED ISSUE #596: Refrain from syncing the stats notification id to cloud storage
The stats notification id gets sync'ed to the cloud secondary. Enhance the code to prevent the stats notification id being sync'ed to cloud secondary.

**Describe alternatives you've considered**

**Additional context**

purnimavenkatasubbu:
**- What I did**
Made latestnotificationId as a public key

**- How I did it**
By setting the isPublic(metadata) to true

**- How to verify it**
stats_notification_sync_test.dart should pass

### DIFF
--- a/at_client/lib/src/service/notification_service_impl.dart
+++ b/at_client/lib/src/service/notification_service_impl.dart
@@ -92,7 +92,7 @@ class NotificationServiceImpl
 
   Future<int?> _getLastNotificationTime() async {
     var lastNotificationKeyStr =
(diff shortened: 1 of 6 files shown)


gkc (changes requested):
I don't think this is the correct approach. I would expect instead for the stats notification id keys to be marked as internal /reserved in some way, similar to other keys that we don't sync from client to server

We had a major backwards compatibility issue recently re last notification id

purnimavenkatasubbu:
Closing this PR as I'm going to raise a new PR with only the latest fix.

PR #623, opened 2022-07-27
feat: encode new line characters in public-key value

sitaram-kalluri:
**- What I did**

- If the public data contains \n character, do a base64Encoding  and decode when the data is fetched

**- How I did it**
- In `PutRequestTransformer` check if the value contains \n character. If yes, encode the data using base64 and set the encoding attribute to type of encoding used.

- In `GetResponseTransformer`, if the key is public or cached:public: key the data is decoded based on the metadata.encoding which represents the type of encoding used.

- In sync_service_impl, add encoding attribute in metadata to sync the field between cloud and local secondaries and if metadata.encoding is not null, use updata:json version of regex to sync the data to cloud secondary.

- Add unit tests to test the PutRequestTransformer and GetResponseTransformer

* TODO: Add functional tests and e2e tests when the server changes are deployed to canary servers.
 
**- How to verify it**
- Create a public key with new line characters and the data should be base64Encoded.

- Encode the new line characters in public keys


at_client/lib/src/service/sync_service_impl.dart
```diff
@@ -491,28 +494,37 @@ class SyncServiceImpl implements SyncService, AtSignChangeListener {
...
-        var metaData =
-            await _atClient.getLocalSecondary()!.keyStore!.getMeta(key);
+        AtData value = await _atClient.getLocalSecondary()!.keyStore!.get(key);
         var keyGen = '';
-        if (metaData != null) {
-          keyGen = _metadataToString(metaData);
+        // If metadata.encoding is not empty, the value is encoded.
+        // When value is encoded, use update:json variant in update regex to
+        // send command to cloud secondary
+        if (value.metaData != null && value.metaData?.encoding != null) {
```
gkc: Should this be `value.metaData!.encoding`?

at_client/lib/src/service/sync_service_impl.dart
```diff
@@ -491,28 +494,37 @@ class SyncServiceImpl implements SyncService, AtSignChangeListener {
...
-        var value = await _atClient.getLocalSecondary()!.keyStore!.get(key);
-        var metaData =
-            await _atClient.getLocalSecondary()!.keyStore!.getMeta(key);
+        AtData value = await _atClient.getLocalSecondary()!.keyStore!.get(key);
         var keyGen = '';
-        if (metaData != null) {
-          keyGen = _metadataToString(metaData);
+        // If metadata.encoding is not empty, the value is encoded.
+        // When value is encoded, use update:json variant in update regex to
+        // send command to cloud secondary
```
gkc: I do like this approach in the client. However, we need to ensure that the non-json update verb syntax also supports `encoding`

at_client/lib/src/service/sync_service_impl.dart
```diff
@@ -712,6 +727,9 @@ class SyncServiceImpl implements SyncService, AtSignChangeListener {
       if (metaData[SHARED_WITH_PUBLIC_KEY_CHECK_SUM] != null) {
         builder.pubKeyChecksum = metaData[SHARED_WITH_PUBLIC_KEY_CHECK_SUM];
       }
+      if (metaData['encoding'] != null) {
```
gkc: replace all use of literal `'encoding'` with a constant, declared in the same place as `SHARED_KEY_ENCRYPTED` etc

at_client/lib/src/transformer/request_transformer/put_request_transformer.dart
```diff
@@ -44,6 +49,14 @@ class PutRequestTransformer
       }
       updateVerbBuilder.dataSignature = await SignInPublicData.signInData(
           updateVerbBuilder.value, encryptionPrivateKey!);
+      // Encode the public data if it contains new line characters
+      if (updateVerbBuilder.value.contains('\n')) {
+        updateVerbBuilder.value =
+            AtEncoderImpl().encodeData(updateVerbBuilder.value, encodingType);
+        updateVerbBuilder.encoding = encodingType.toString();
+        // When data is encode, use update:json in update regex. Hence setting isJson to true.
```
gkc: Typo - should read `When data is encoded` not `is encode`

gkc (changes requested):

at_client/lib/src/service/sync_service_impl.dart
sitaram-kalluri: Replaced with constant.

at_client/lib/src/service/sync_service_impl.dart
sitaram-kalluri: Removed the JSON approach and made changes to use to update regex.

at_client/lib/src/service/sync_service_impl.dart
sitaram-kalluri: As a part of reverting JSON approach, removed the above lines.

gkc (changes requested):
Tests are failing

sitaram-kalluri:
> Tests are failing

Fixed the failing tests.

at_client/lib/src/transformer/request_transformer/put_request_transformer.dart
```diff
@@ -44,6 +48,12 @@ class PutRequestTransformer
       }
       updateVerbBuilder.dataSignature = await SignInPublicData.signInData(
           updateVerbBuilder.value, encryptionPrivateKey!);
+      // Encode the public data if it contains new line characters
+      if (updateVerbBuilder.value.contains('\n')) {
+        updateVerbBuilder.value =
+            AtEncoderImpl().encodeData(updateVerbBuilder.value, encodingType);
+        updateVerbBuilder.encoding = encodingType.toString().split('.')[1];
```
murali-shris: please check how to get name from enum in a cleaner way. 
we can replace with encodingType.name using extensions
https://stackoverflow.com/questions/29567236/dart-how-to-get-the-name-of-an-enum-as-a-string

at_client/lib/src/util/at_client_util.dart
```diff
@@ -133,6 +135,14 @@ class AtClientUtil {
     }
     return atKey.key!;
   }
+
+  /// Converts the String representation of encodingType to [EncodingType]
+  EncodingType atDecoderStringToEnum(String encodingType) {
```
murali-shris: once we have extension on enum..this method may not be required

at_client/lib/src/transformer/request_transformer/put_request_transformer.dart
sitaram-kalluri: This change is implemented.

at_client/lib/src/util/at_client_util.dart
sitaram-kalluri: Removed the above method.

murali-shris:
will approve this PR once server and persistence changes are released

gkc (commented):
Will approve once server changes are in prod

sitaram-kalluri:
> Will approve once server changes are in prod

Sure Gary, Thank you.

gkc (approved):

PR #625, opened 2022-07-29
fix: Sync conflict bug

murali-shris:
**- What I did**
- fixed an issue in sync service impl. if there is an issue syncing a key, do not throw exception
**- How I did it**
- removed return statement in _syncToRemote

### DIFF
--- a/at_client/lib/src/service/sync_service_impl.dart
+++ b/at_client/lib/src/service/sync_service_impl.dart
@@ -386,7 +386,6 @@ class SyncServiceImpl implements SyncService, AtSignChangeListener {
         } on Exception catch (e) {
           _logger.severe(
               'exception syncing entry to local $serverCommitEntry - ${e.toString()}');
-          return keyInfoList;
         }
       }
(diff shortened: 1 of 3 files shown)


at_client/lib/src/service/sync_service_impl.dart
```diff
@@ -339,7 +339,6 @@ class SyncServiceImpl implements SyncService, AtSignChangeListener {
           } on Exception catch (e) {
             _logger.severe(
                 'exception while updating commit entry for entry:$entry ${e.toString()}');
-            return keyInfoList;
```
gkc, line 342: need to also remove the return statement at line 348 (now line 347)

gkc (changes requested):

at_client/lib/src/service/sync_service_impl.dart
murali-shris: done

gkc (approved):
LGTM

We will follow up with #627

PR #646, opened 2022-08-09
fix: revert deleted function

sonle-geekyants:
**- What I did**
Because we lost some functions in at_client_mobile when upgrading the data structure, I bring all of them back for a patch release with a deprecated warning.

- `getValue` and `putValue` are not supported with 3.2.0. So, we are throwing **UnimplementedError** from them with a Depraceted warning. Also, these methods should not be exposed to app side and onboarding package should have the control over updating keychain data.

**- How I did it**
Add deleted functions and make it compatible with the now at_client_mobile version.

**- How to verify it**
Should review my code and try out on your app

### DIFF
--- a/at_client_mobile/lib/src/keychain_manager.dart
+++ b/at_client_mobile/lib/src/keychain_manager.dart
@@ -354,15 +354,15 @@ class KeyChainManager {
     return atsigns?.secret;
   }
 
-  /// Use [getValue]
-  @Deprecated("Use getValue")
+  /// Use [readAtsign]
+  @Deprecated("Use readAtsign function to get AtsignKey")
   Future<String?> getPrivateKeyFromKeyChain(String atsign) async {
     final atsigns = await readAtsign(name: atsign);
     return atsigns?.pkamPrivateKey;
   }
 
-  /// Use [getValue]
-  @Deprecated("Use getValue")
+  /// Use [readAtsign]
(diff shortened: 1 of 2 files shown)


at_client_mobile/lib/src/keychain_manager.dart
```diff
@@ -570,6 +570,43 @@ class KeyChainManager {
     }
   }
 
+  /// Fetches the list of onboarded atsign saved in map datatype
+  Future<Map<String, bool?>> checkForValuesInFlutterKeychain() async {
```
xavierchanth, line 576: Mark this as deprecated and recommend to use `getAtsignsWithStatus()` instead.

xavierchanth (changes requested):
These changes also need to be made to ensure backwards compatibility:

`Future<bool> resetAtSignFromKeychain({required String atsign})`
needs to be changed to positional parameter format
`Future<bool> resetAtSignFromKeychain(String atsign)`

`Future<String?> getSecretFromKeychain(String atsign)`
previously returned a `Future<String>`
...maybe an empty string is an appropriate fallback value for when it is null?

xavierchanth (approved):
LGTM, thanks @sonle-geekyants @sachins-geekyants !

murali-shris (approved):

gkc (approved):

PR #651, opened 2022-08-11
fix: Ensure _syncFromServer, when it encounters an exception and having handled the exception chaining, then rethrows the exception 

gkc:
**- What I did**
Fixed two issues
* Recent changes to SyncServiceImpl to handle chained exceptions required exceptions to be caught in _syncFromServer ... however the exception wasn't then being rethrown
* test for chained exception caused by sync timeout was not actually doing any checks

**- How I did it**
* rethrow exception in the recently-introduced catch block `SyncServiceImpl._syncFromServer`
* finish implementation of the unit test case

**- How to verify it**
Tests pass

* fix: Ensure _syncFromServer, when it encounters an exception and having handled the exception chaining, then rethrows the exception

### DIFF
--- a/at_client/lib/src/service/sync_service_impl.dart
+++ b/at_client/lib/src/service/sync_service_impl.dart
@@ -93,8 +93,7 @@ class SyncServiceImpl implements SyncService, AtSignChangeListener {
   void _scheduleSyncRun() {
     _cron = Cron();
 
-    _cron.schedule(Schedule.parse('*/$_syncRunIntervalSeconds * * * * *'),
-        () async {
+    _cron.schedule(Schedule.parse('*/$_syncRunIntervalSeconds * * * * *'), () async {
       try {
         await processSyncRequests();
       } on Exception catch (e, trace) {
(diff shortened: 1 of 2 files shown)


at_client/lib/src/service/sync_service_impl.dart
```diff
@@ -396,6 +395,7 @@ class SyncServiceImpl implements SyncService, AtSignChangeListener {
             ExceptionScenario.remoteVerbExecutionFailed, e.message));
         _logger.severe(
             'Exception occurred in fetching sync response : ${e.getTraceMessage()}');
+        rethrow;
```
murali-shris, line 398: @gkc 
assume commitIds 100-150 from server has to be synced in batches of 10 commitIds each.
If there is an exception at commitId 120..121-150 will not be synced from server. is this behavior fine?

gkc: Up until Jul 29, the exception was not being caught - i.e. if an exception was encountered, it was thrown. I am simply reverting to that behaviour. I don't believe it is correct to continue in this case when an exception has occurred (note that the exception is occurring **when getting the response to the sync: command** and not when processing results)

murali-shris (approved):

PR #652, opened 2022-08-11
feat: Add enforceNamespace (default value true) to AtClientPreference.

gkc:
**- What I did**
* As a temporary measure allow app developers to disable the enforcement of namespace presence during AtKey validation for `put` requests **if they really need to do so** for backwards compatibility.

**- How I did it**
* Added `enforceNamespace` to AtClientPreference, default true
* Marked it as Deprecated as this is intended as a temporary measure only and will be removed in version 4
* Added code to respect value of AtClientPreference.enforceNamespace when doing validation of the AtKey during `put` requests

**- How to verify it**
Tests pass

feat: Add enforceNamespace (default value true) to AtClientPreference.

### DIFF
--- a/at_client/lib/src/client/at_client_impl.dart
+++ b/at_client/lib/src/client/at_client_impl.dart
@@ -384,13 +384,18 @@ class AtClientImpl implements AtClient {
       atKey.namespace ??= preference?.namespace;
     }
     // validate the atKey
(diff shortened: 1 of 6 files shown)


murali-shris (dismissed):

gkc:
@murali-shris I removed the dependency overrides and updated some outdated dependencies, otherwise unchanged since your earlier approval

murali-shris (approved):

PR #666, opened 2022-08-24
fix: validate atKeys in the authenticate method.

### LINKED ISSUE #640: authenticate returns true for invalid atKeys file
**Describe the bug**
If we reset an atSign and activate it again we will have a new atKeys file and old atkeys file will become invalid.
`AtClientService.authenticate` returns **true** for old atkeys file as well.

**To Reproduce**
1. activate an atSign and get the atKeys file.
2. reset the atSign.
3. activate again and get new atKeys file.
4. If we upload old atKeys file to authenticate it returns `true`.

**Expected behavior**
`AtClientService.authenticate` should return false for invalid atKeys.

 - macOS
 
  - atmospherePro

sitaram-kalluri:
**- What I did**
- As of now, when a .atKeys file is provided, the keys are directly added to the key-chain manager. So, if an incorrect key file is provided, the keys are added to key-chain manager but the authentication fails at a later stage.
- So, to fix this before adding keys to key-chain manager, performing PKAM authentication to validate it atKeys provided. If valid keys are added to key-chain manager. else UnAuthenticatedException is thrown.

**- How I did it**
- In AtClientService.authenticate method, after the decoding the JSON, a new method `_validateAtKeys`
 will be called. If keys are valid, stores keys to key-chain manager else throws exception

**- How to verify it**
- Provided invalid atkeys and exception should be thrown.

### DIFF
--- a/at_client_mobile/lib/src/at_client_service.dart
+++ b/at_client_mobile/lib/src/at_client_service.dart
@@ -139,22 +139,48 @@ class AtClientService {
   Future<bool> authenticate(
       String atsign, AtClientPreference atClientPreference,
       {OnboardingStatus? status, String? jsonData, String? decryptKey}) async {
-    // If cramSecret is null, PKAMAuth is completed,
-    // decode the json data and add keys to KeyChainManager
+    /**
+     * The authentication is performed either by CRAM authentication or PKAM authentication
+     * 1. If AtClientPreference.cramSecret is populated, then atSign is considered as new atSign.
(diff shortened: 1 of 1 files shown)


gkc:
I'm going to remove myself as a reviewer on this PR. I'd like to see all of the apps tested against this branch

sachins-geekyants (approved):

sachins-geekyants:
Tested this branch with atmopsherePro and the issue was not reproducible.

PR #667, opened 2022-08-24
feat: add clientConfig to from verb

sitaram-kalluri:
**- What I did**
- Send at_client version to server

**- How I did it**
- Pass the client version via the clientConfig in AtLookupImpl in the RemoteSecondary.

**- How to verify it**

### DIFF
--- a/at_client/lib/src/preference/at_client_config.dart
+++ b/at_client/lib/src/preference/at_client_config.dart
@@ -1,0 +1,13 @@
+/// The class contains all the client configurations.
+class AtClientConfig {
+  static final AtClientConfig _singleton = AtClientConfig._internal();
+
+  AtClientConfig._internal();
+
+  factory AtClientConfig.getInstance() {
+    return _singleton;
+  }
+
+  /// Represents the at_client version.
+  final String atClientVersion = '3.0.36';
+}
--- a/at_client/lib/src/util/at_client_util.dart
+++ b/at_client/lib/src/util/at_client_util.dart
@@ -132,13 +132,6 @@ class AtClientUtil {
     }
     return atKey.key!;
   }
-
-  static String getVersion() {
-    var applicationConfiguration = ApplicationConfiguration('pubspec.yaml');
-    var yamlMap = applicationConfiguration.getYaml();
-    var version = yamlMap![VERSION];
-    return version;
-  }
 }
 
 class Tuple<T1, T2> {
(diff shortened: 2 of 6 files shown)


at_client/lib/src/util/at_client_util.dart
```diff
@@ -135,6 +134,13 @@ class AtClientUtil {
     }
     return atKey.key!;
   }
+
+  static String getVersion() {
+    var applicationConfiguration = ApplicationConfiguration('pubspec.yaml');
```
gkc: This is dependent on what the "current working directory" is ...

Suggest you use this approach instead: https://stackoverflow.com/a/72486732

gkc (changes requested):

at_client/lib/src/preference/at_client_config.dart
```diff
@@ -0,0 +1,13 @@
...
+  static final AtClientConfig _singleton = AtClientConfig._internal();
+
+  AtClientConfig._internal();
+
+  factory AtClientConfig.getInstance() {
+    return _singleton;
+  }
+
+  /// Represents the at_client version.
+  final String atClientVersion = '3.0.36';
```
gkc, line 12: Maintenance of this needs to be automated - maybe through using build_runner. At a minimum, you need to add a VERY VISIBLE comment to pubpsec.yaml stating that when the package version is increasing, then this constant needs to be changed accordingly.

gkc (changes requested):

at_client/lib/src/util/at_client_util.dart
sitaram-kalluri: Explored the above, the rootBundle is a flutter dependent package. Hence unable to use it in a Dart package.

at_client/lib/src/preference/at_client_config.dart
sitaram-kalluri: @gkc: For now, to get going, will add a comment in pubspec.yaml. And keep exploring the alternatives to automate the capturing the at_client version. Does this sound's good?

gkc: 👍 thanks Sita

sitaram-kalluri:
@gkc: Updated the pubspec.yaml with a comment. and changed the e2e test atSign from `ce2e4` to `ce2e2` because `ce2e4` runs on a prod version of at_secondary_server and hence the e2e tests are failing because the changes related to clientConfig are not yet released into Prod. I will a new PR having tests guarded with version checking to support cross version testing.

gkc (dismissed):
OK for now

gkc (approved):

murali-shris (approved):

PR #672, opened 2022-08-29
fix: NotificationService.subscribe fails to return existing listener on same regex.

### LINKED ISSUE #684: NotificationService.subscribe fails to return existing listener on same regex
**Describe the bug**
When trying to reproduce the following issue: https://github.com/atsign-foundation/at_client_sdk/issues/661, Found that the notification service-> subscribe method fails to return the existing listener when already present. Instead creates a new StreamController.

The root cause of the issue is we are check on String (regex) instead of comparing on the NotificationConfig object. Below is the code snippet.

```dart
if (_streamListeners.containsKey(regex)) {
```

Fix:
Modify the streamListeners from Map to HashMap and overrides the equal and hashcode methods to compare the NotificationConfig Object.

sitaram-kalluri:
**- What I did**
- Fix the issue of single notification being sent for multiple times.

  **-Root cause**
  At the moment, NotificationService.subscribe returns a Stream.broadcast which supports multiple listeners. From the app,
 `NotificationService.subscribe` method is being invoked multiple time's with same regex which leads to the same notification being fetched in the listen callback.

  **-Fix**
  - Replace the Stream.broadcast with simple Streams which supports only single listener. If subscribed for the second time, 
  throws a `Bad State. stream was already listened to`(This will **NOT** cause the app to crash). 
  - Ideally, for a given regex, having a single listener would suffice and i do not see a scenario which would require multiple 
  listeners for the same regex. Please share your thoughts on this? 
 
**- How I did it**
- Replace the Stream.broadcast with simple Streams.
- Replace the Map with HaspMap and override the equals and hashcode methods to compare the NotificationConfig object on regexes.

**- How to verify it**
- The Notification should be sent to app only once.
- If a NotificationService.subscribe is invoked for more than once, then a `Bad State` erorr would be thrown.

### DIFF
--- a/at_client/test/notification_service_test.dart
+++ b/at_client/test/notification_service_test.dart
@@ -381,4 +381,48 @@ void main() {
           NotificationStatusEnum.undelivered);
     });
   });
+
+  group('A group of tests to validate notification subscribe method', () {
+    test('NotificationService subscribe returns a new stream for a new regex', () async {
+      var notificationServiceImpl = await NotificationServiceImpl.create(
+          mockAtClientImpl,
+          atClientManager: mockAtClientManager,
+          monitor: mockMonitor) as NotificationServiceImpl;
+
+      var notificationStream = notificationServiceImpl.subscribe(
+          regex: '.wavi', shouldDecrypt: false);
+      notificationStream.listen((event) async {
+        print(event);
+      });
+
+      var notificationStream1 = notificationServiceImpl.subscribe(
+          regex: '.buzz', shouldDecrypt: false);
+      notificationStream1.listen((event) {
+        print(event);
+      });
+      expect(notificationServiceImpl.getStreamListenersCount(), 2);
+      notificationServiceImpl.stopAllSubscriptions();
+    });
+
+    test('NotificationService subscribe returns an existing stream for a same regex', () async {
+      var notificationServiceImpl = await NotificationServiceImpl.create(
+          mockAtClientImpl,
+          atClientManager: mockAtClientManager,
+          monitor: mockMonitor) as NotificationServiceImpl;
+
+      var notificationStream = notificationServiceImpl.subscribe(
+          regex: '.wavi', shouldDecrypt: false);
+      notificationStream.listen((event) async {
+        print(event);
+      });
+
+      var notificationStream1 = notificationServiceImpl.subscribe(
+          regex: '.wavi', shouldDecrypt: true);
+      notificationStream1.listen((event) {
+        print(event);
+      });
+      expect(notificationServiceImpl.getStreamListenersCount(), 1);
+      notificationServiceImpl.stopAllSubscriptions();
+    });
+  });
 }
--- a/at_client/lib/src/service/notification_service_impl.dart
+++ b/at_client/lib/src/service/notification_service_impl.dart
@@ -1,4 +1,5 @@
 import 'dart:async';
+import 'dart:collection';
 import 'dart:convert';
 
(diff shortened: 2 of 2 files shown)


sitaram-kalluri:
The e2e tests have failed because KeyStream tests have multiple subscriptions causing Bad State: Stream is already listened to. Basing on the dicsussion in the internal stand-up will refactor the changes.

gkc:
@kalluriramkumar checks are failing ...

gkc (changes requested):
Checks are failing

sitaram-kalluri:
> @kalluriramkumar checks are failing ...

Yes Gary, At the moment, the KeyStream tests are have multiple listeners, hence the tests the are failing. Wanted to check if the removing the broadcast (Changing Stream.broadcast to simple stream) is fine. If yes, will refactor the tests.

sitaram-kalluri:
@XavierChanth : I am not sure on which apps are using KeyStreams. Request your feedback/thoughts on changing the Stream.broadcast to simple streams.

xavierchanth:
For root cause:
Is the subscription being invoked multiple times in our client code, or is it an app developer mistake causing the problem?
If client code, then I think we should seek to fix the problem of multiple invocations, here's why:

Keep in mind the following from dart docs:

> Single-subscription streams are generally used for streaming chunks of larger contiguous data, like file I/O.
> 
> A broadcast stream allows any number of listeners, and it fires its events when they are ready, whether there are listeners or not.

Notifications are event driven, and ideally should be using broadcast streams.
From a Flutter development perspective, it is useful to be able to subscribe from multiple places, to reduce the need for unnecessary re-renders of certain widgets.

Despite all of this, if we still decide to opt into single-subscription streams:
KeyStream will need to be refactored so that it does not need to subscribe to the notification service. This will most likely require coupling notification service and key stream, which is less ideal.

sitaram-kalluri:
> For root cause: Is the subscription being invoked multiple times in our client code, or is it an app developer mistake causing the problem? If client code, then I think we should seek to fix the problem of multiple invocations, here's why:
> 
> Keep in mind the following from dart docs:
> 
> > Single-subscription streams are generally used for streaming chunks of larger contiguous data, like file I/O.
> > A broadcast stream allows any number of listeners, and it fires its events when they are ready, whether there are listeners or not.
> 
> Notifications are event driven, and ideally should be using broadcast streams. From a Flutter development perspective, it is useful to be able to subscribe from multiple places, to reduce the need for unnecessary re-renders of certain widgets.
> 
> Despite all of this, if we still decide to opt into single-subscription streams: KeyStream will need to be refactored so that it does not need to subscribe to the notification service. This will most likely require coupling notification service and key stream, which is less ideal.

@XavierChanth : It is from the app's code the multiple requests coming in and communicated to the app developer. Putting Stream.broadcast back in-place.

at_client/test/notification_service_test.dart
```diff
@@ -381,4 +381,26 @@ void main() {
           NotificationStatusEnum.undelivered);
     });
   });
+
+  group('A group of tests to validate notification subscribe method', () {
+    test('A subscribe returns a new stream for a new regex', () async {
```
gkc: Tests should have explicit assertions / expectations.

gkc (changes requested):

at_client/test/notification_service_test.dart
sitaram-kalluri: Modified the tests to assert on the size of the streamListeners map to verify if a new instances is being created or not.

gkc (approved):

PR #700, opened 2022-09-15
fix: key not found exception not being handled

srieteja:
**- What I did**
- add exception handling for KeyNotFoundException in EncryptionService._getAESKeyForEncryption()
- remove duplicate code in the same method


sitaram-kalluri:
@srieteja : The encryption_service.dart is obsolete. Currently, for encryption, we have the logic placed in encryption_service folder. Following is the location to files: https://github.com/atsign-foundation/at_client_sdk/tree/trunk/at_client/lib/src/encryption_service

srieteja:
Thanks @kalluriramkumar for informing me. I'll try migrating to the new service

srieteja:
Closing this as it's not relevant

PR #704, opened 2022-09-16
fix: expose notification params

sitaram-kalluri:
**- What I did**
- Expose the following notifications params - priority, strategy, notifier, latestN and notificationExpiry.
- Added setUp method in "atclient_notify_test.dart" to set currentAtSign before each test.
**- How I did it**

**- How to verify it**
- Added unit tests to verify priority, strategy, notifier, latestN and notificationExpiry can be set via the NotificationParams.
- Added functional test and end-to-end test to verify notification expiry.

* Expose new attributes in NotificationParams.forUpdate method


sitaram-kalluri:
The Changes are completed, but pending on writing on functional and end2end tests. Since, fetching notifications via the monitor is not feasible, we have introduced notify fetch with which we can fetch a notification using the notification-id. Waiting the notify fetch changes for writing functional and e2e tests.

gkc:
@sitaram-kalluri Looks like there are useful changes in this PR. Can you bring it up to date with trunk & resolve conflicts please? Let's get it tested, reviewed, merged and published

sitaram-kalluri:
> @sitaram-kalluri Looks like there are useful changes in this PR. Can you bring it up to date with trunk & resolve conflicts please? Let's get it tested, reviewed, merged and published

Sure Gary.

packages/at_client/lib/src/service/notification_service.dart
```diff
@@ -151,10 +151,23 @@ class NotificationParams {
...
+  StrategyEnum strategy = StrategyEnum.all;
+
+  /// Represents the most recent 'N' notifications to be delivered. Defaults to 1.
+  int latestN = 1;
+
+  /// Groups the notifications
+  String notifier = SYSTEM;
+
+  /// Represents the time for the notification to expire. Defaults to 15 minute
+  Duration notificationExpiry = Duration(minutes: 15);
```
gkc: 15 minutes? Why 15 minutes? Seems like a very short duration given that, as I understand it, our current default is 24 hours?

packages/at_client/lib/src/service/notification_service.dart
```diff
@@ -166,24 +179,14 @@ class NotificationParams {
...
-  PriorityEnum get priority => _priority;
-
-  StrategyEnum get strategy => _strategy;
-
-  String get notifier => _notifier;
-
-  int get latestN => _latestN;
-
   /// Returns [NotificationParams] to send an update notification.
   static NotificationParams forUpdate(AtKey atKey, {String? value}) {
```
gkc: Shouldn't we add all of the NotificationParams parameters as optional pararmeters to the forUpdate and forText methods? Seems easier to me to be able to set the properties when calling forUpdate/forText rather than first calling forUpdate / forText, and then setting the properties on the NotificationParams they return

gkc (changes requested):

packages/at_client/lib/src/service/notification_service.dart
sitaram-kalluri: Yes Gary, Will add optional parameters.

packages/at_client/lib/src/service/notification_service.dart
sitaram-kalluri: Will update the default time out duration.

gkc: Hasn't changed ?

packages/at_client/lib/src/service/notification_service.dart
gkc: Parameters haven't been added?

gkc (changes requested):

packages/at_client/lib/src/service/notification_service.dart
sitaram-kalluri: @gkc: Apologies, Missed committing and pushing the changes. Added optional parameters

packages/at_client/lib/src/service/notification_service.dart
sitaram-kalluri: @gkc: Apologies, Updated the duration to 24 hours and pushed the changes

gkc (approved):

gkc (changes requested):
Oops didn't finish review. Please don't merge yet.

packages/at_client/lib/src/transformer/request_transformer/notify_request_transformer.dart
```diff
@@ -50,7 +50,8 @@ class NotificationRequestTransformer
       ..priority = notificationParams.priority
       ..strategy = notificationParams.strategy
       ..latestN = notificationParams.latestN
-      ..notifier = notificationParams.notifier;
+      ..notifier = notificationParams.notifier
+      ..ttln = notificationParams.notificationExpiry.inMilliseconds;
```
gkc, line 54: Do we have tests which exercise the ttln functionality, and can we perhaps write a functional test to verify this (since we've just added it to the transformer?)

gkc (changes requested):

packages/at_client/lib/src/transformer/request_transformer/notify_request_transformer.dart
sitaram-kalluri: @gkc: Updated a functional test to verify ttln

sitaram-kalluri:
Added dependency overrides to consume the changes in https://github.com/atsign-foundation/at_tools/pull/290. Will update with the hosted version once the changes are published.

packages/at_client/lib/src/service/notification_service.dart
```diff
@@ -175,16 +176,40 @@ class NotificationParams {
...
+  /// * priority: Represents the priority of the notification. The notification marked [PriorityEnum.high] takes precedence over other notifications.
+  ///
+  /// * strategy: When notifications marked with [StrategyEnum.all], all the notifications are sent. For [StrategyEnum.latest], only the [latestN] of a notifier are sent
+  ///
+  /// * latestN: Represents the count of notifications to store that belong to a particular [notifier].
+  ///
+  /// * notifier: Groups the notifications that has the same notifier.
+  ///
+  /// * notificationExpiry: Refers to the amount of time the notification is
+  /// available in the KeyStore. Beyond which the notification is removed from the KeyStore.
```
gkc, line 194: Excellent documentation! 🎉

packages/at_client/lib/src/transformer/request_transformer/notify_request_transformer.dart
gkc: Thank you @sitaram-kalluri

gkc (dismissed):
LGTM. Nice work @sitaram-kalluri

gkc:
Ran dart format and committed those changes so that the format check passes.

gkc (dismissed):
LGTM. Nice work @sitaram-kalluri

gkc:
Hi @sitaram-kalluri  ... not really sure what is going on here; your notification expiry test is failing in functional test; but I copied it into end2end tests and it works fine there against the various `@ce2e` atServers. It also works fine when I run end to end tests locally against atServers running from latest atServer trunk; and also works fine when I run them locally against atServers which are running version 3.0.28 (current prod)

I don't have my local setup to run functional tests against the docker image - can you do that on Friday morning please and see if you can figure out what is going on?

gkc:
@sitaram-kalluri further to my last comment - I added a print to the test so it prints out what it fetches from the server - here's what we get:
```
Fetched notification AtNotification{id: 575bd0bb-760d-46d4-b4aa-cc985b3f5766, key: , from: , to: , epochMillis: 0, value: null, operation: null, status: NotificationStatus.expired, expiresAt: null}
```

So .. it has expired? Makes no sense!  Especially when, as I mentioned, everything works fine when running end to end tests.

Hope you have more luck than I did in tracking it down

gkc:
Progress: added some more print statements; here's what the output is:
```
Testing notification expiry - test run #0
  notificationResult: id: 0d29a84a-913a-4bf5-bebd-b9374cc6cef4 status: NotificationStatusEnum.undelivered
  notificationResult.atClientException: Exception: @alice🛠 is not authorized to send notification as @bob🛠
  Fetched notification AtNotification{id: 0d29a84a-913a-4bf5-bebd-b9374cc6cef4, key: , from: , to: , epochMillis: 0, value: null, operation: null, status: NotificationStatus.expired, expiresAt: null}
  Null check operator used on a null value
  test/atclient_notify_test.dart 217:48  main.<fn>
```

What I don't understand now is that the "is not authorized to send notification" message is only added in your server PR https://github.com/atsign-foundation/at_server/pull/1257 ... which has not yet been merged to trunk.

gkc:
OK two things:

1. Checks are passing now, after I changed that test to explicitly call atClientManager.setCurrentAtSign. Problem here is that the current atSign is changed by another test, and setUpAll only runs ONCE, not for every test **=> this needs to be fixed**
2. Second issue - why is the "is not authorized" message appearing ... is a mystery to me. I can't figure out how that message can have made its way into the dev_env docker image, since the server PR https://github.com/atsign-foundation/at_server/pull/1257 which added that code has not yet been approved or merged to trunk. Any ideas?

sitaram-kalluri:
> OK two things:
> 
> 1. Checks are passing now, after I changed that test to explicitly call atClientManager.setCurrentAtSign. Problem here is that the current atSign is changed by another test, and setUpAll only runs ONCE, not for every test **=> this needs to be fixed**
> 2. Second issue - why is the "is not authorized" message appearing ... is a mystery to me. I can't figure out how that message can have made its way into the dev_env docker image, since the server PR [fix: Check if sharedBy atSign is authorized to send notifications at_server#1257](https://github.com/atsign-foundation/at_server/pull/1257) which added that code has not yet been approved or merged to trunk. Any ideas?

On # 2, The changes of validating if atSign is authorized to send notifications are done in the at_client side and are already merged to trunk in the following https://github.com/atsign-foundation/at_client_sdk/pull/961. Perhaps the message might be from the at_client SDK only. Not from the server.

gkc:
That would make sense indeed.

gkc (approved):
LGTM 🎉

PR #715, opened 2022-09-20
fix: write shared_key to remote secondary

sitaram-kalluri:
**- What I did**
- Write the shared_key to the remote secondary to expedite the sync process. The reason behind expediting is because when notification is sent before the shared_key is sync'ed the decryption fails on the receiver's side.

**- How I did it**
- Write shared key to remote secondary.

**- How to verify it**
- Initial notifications on a new atsign should not fail with decryption error on the receiver side.


sitaram-kalluri:
The changes are obsolete. Hence closing the PR.

PR #738, opened 2022-10-03
feat: uptake notify fetch changes

### LINKED ISSUE #931: Retrieve the AtNotification using the notification-id
- Feature request to provide capability to retrieve the AtNotification object using the notification-id. 

- At the moment, only way to retrieve the notifications is query using the `notify:list` which returns a list of notifications; but user cannot retrieve a specific notification using the notification id.

**Describe the solution you'd like**
- Provide a new variant of notify which enables to retrieve the AtNotification object using the notification id

sitaram-kalluri:
**- What I did**
- Add 'fetch' method to NotificationService class and fetches the AtNotification from the cloud secondary using the notification id

**- How I did it**
- Files Modified:
- * at_notification.dart - When a notification is expired, the notification is removed from the keystore and hence details cannot be fetched to populate the fields in AtNotification class (except the id). So, removed the late modifier and initialized the fields to empty string and epochMillis to '0 (Zero)' to avoid LateInitialization error when accessing the AtNotification.
- * Since `AtNotification()` has arguments that are available when notification is expired, added a named constructor `AtNotification.empty` and populate the fields.
- * Added unit tests for the scenario's
        - when notification is available in keystore
        - when notification is not available in keystore
        
- * Removed the skip tag on end2end test and refactored to fetch the notification using `notify fetch` and added assertions

**- How to verify it**

### DIFF
--- a/at_client/lib/src/service/notification_service_impl.dart
+++ b/at_client/lib/src/service/notification_service_impl.dart
@@ -97,22 +97,23 @@ class NotificationServiceImpl
           'monitor is already started for ${_atClient.getCurrentAtSign()}');
       return;
     }
-    if (AtClientManager.getInstance()
-        .atClient
-        .getPreferences()!
-        .fetchOfflineNotifications) {
-      final lastNotificationTime = await _getLastNotificationTime();
-      await _monitor!.start(lastNotificationTime: lastNotificationTime);
-    } else {
-      await _monitor!.start();
-    }
+    await _monitor!.start(lastNotificationTime: await getLastNotificationTime());
 
     if (_monitor!.status == MonitorStatus.started) {
       _isMonitorPaused = false;
     }
   }
 
-  Future<int?> _getLastNotificationTime() async {
+  @visibleForTesting
+  Future<int?> getLastNotificationTime() async {
+    if (_atClientManager.atClient.getPreferences()!.fetchOfflineNotifications == false) {
+      // fetchOfflineNotifications == false means issue `monitor` command without a lastNotificationTime
+      // which will result in the server not sending any previously received notifications
+      return null;
+    }
+
+    // fetchOfflineNotifications == true (the default) means we want all notifications since the last one we received
+    // We keep track of the last notification id in the client-side key store
     var lastNotificationKeyStr =
         '$notificationIdKey.${_atClient.getPreferences()!.namespace}${_atClient.getCurrentAtSign()}';
     var atKey = AtKey.fromString(lastNotificationKeyStr);
@@ -198,7 +199,7 @@ class NotificationServiceImpl
     Future.delayed(
         Duration(seconds: 15),
         () async => _monitor!
-            .start(lastNotificationTime: await _getLastNotificationTime()));
+            .start(lastNotificationTime: await getLastNotificationTime()));
   }
 
   void _onMonitorError(Exception e) {
--- a/at_client/lib/src/service/notification_service_impl.dart
+++ b/at_client/lib/src/service/notification_service_impl.dart
@@ -12,6 +12,7 @@ import 'package:at_client/src/manager/monitor.dart';
 import 'package:at_client/src/preference/monitor_preference.dart';
 import 'package:at_client/src/response/default_response_parser.dart';
 import 'package:at_client/src/response/notification_response_parser.dart';
+import 'package:at_client/src/response/response.dart';
 import 'package:at_client/src/service/notification_service.dart';
 import 'package:at_client/src/transformer/request_transformer/notify_request_transformer.dart';
 import 'package:at_client/src/transformer/response_transformer/notification_response_transformer.dart';
@@ -19,7 +20,10 @@ import 'package:at_client/src/util/at_client_util.dart';
 import 'package:at_client/src/util/at_client_validation.dart';
 import 'package:at_client/src/util/regex_match_util.dart';
 import 'package:at_client/src/response/at_notification.dart';
+import 'package:at_commons/at_builders.dart';
 import 'package:at_commons/at_commons.dart';
+import 'package:at_persistence_secondary_server/at_persistence_secondary_server.dart'
+    as at_persistence_secondary_server;
 import 'package:at_utils/at_utils.dart';
 import 'package:meta/meta.dart';
 
@@ -407,5 +411,38 @@ class NotificationServiceImpl
   @visibleForTesting
   getStreamListenersCount() {
     return _streamListeners.length;
+  }
+
+  @override
+  Future<AtNotification> fetch(String notificationId) async {
+    var notifyFetchVerbBuilder = NotifyFetchVerbBuilder()
+      ..notificationId = notificationId;
+    var atNotificationStr = await _atClient
+        .getRemoteSecondary()
+        ?.executeVerb(notifyFetchVerbBuilder);
+    if (atNotificationStr == null) {
+      throw AtClientException.message('Failed to fetch the notification id',
+          intent: Intent.remoteVerbExecution,
+          exceptionScenario: ExceptionScenario.remoteVerbExecutionFailed);
+    }
+    AtResponse atResponse = DefaultResponseParser().parse(atNotificationStr);
+    var atNotificationMap = jsonDecode(atResponse.response);
+    if (atNotificationMap['notificationStatus'] ==
+        at_persistence_secondary_server.NotificationStatus.expired.toString()) {
+      return AtNotification.empty()
+        ..id = atNotificationMap['id']
+        ..status = atNotificationMap['notificationStatus'];
+    }
+    return AtNotification.empty()
+      ..id = atNotificationMap['id']
+      ..key = atNotificationMap['notification']
+      ..from = atNotificationMap['fromAtSign']
+      ..to = atNotificationMap['toAtSign']
+      ..epochMillis = DateTime.parse(atNotificationMap['notificationDateTime'])
+          .millisecondsSinceEpoch
+      ..status = atNotificationMap['notificationStatus']
+      ..value = atNotificationMap['atValue']
+      ..operation = atNotificationMap['opType']
+      ..messageType = atNotificationMap['messageType'];
   }
 }
--- a/at_client/lib/src/service/notification_service_impl.dart
+++ b/at_client/lib/src/service/notification_service_impl.dart
@@ -417,9 +417,14 @@ class NotificationServiceImpl
   Future<AtNotification> fetch(String notificationId) async {
     var notifyFetchVerbBuilder = NotifyFetchVerbBuilder()
       ..notificationId = notificationId;
-    var atNotificationStr = await _atClient
-        .getRemoteSecondary()
-        ?.executeVerb(notifyFetchVerbBuilder);
+    String? atNotificationStr;
+    try {
+      atNotificationStr = await _atClient
+          .getRemoteSecondary()
+          ?.executeVerb(notifyFetchVerbBuilder);
+    } on AtException catch (e) {
+      throw AtExceptionManager.createException(e);
+    }
     if (atNotificationStr == null) {
       throw AtClientException.message('Failed to fetch the notification id',
           intent: Intent.remoteVerbExecution,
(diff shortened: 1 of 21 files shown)


at_client/lib/src/service/notification_service_impl.dart
```diff
@@ -408,4 +412,37 @@ class NotificationServiceImpl
...
   }
+
+  @override
+  Future<AtNotification> fetch(String notificationId) async {
+    var notifyFetchVerbBuilder = NotifyFetchVerbBuilder()
+      ..notificationId = notificationId;
+    var atNotificationStr = await _atClient
+        .getRemoteSecondary()
+        ?.executeVerb(notifyFetchVerbBuilder);
+    if (atNotificationStr == null) {
```
gkc: Need some more clarity on the potential exceptions here. Could have
* notification with that ID does not exist
* server-side exception of some sort in the cloud secondary
* failed to connect to cloud secondary

Also - if the notification does not exist, is that really an exception? Should we maybe just return null if the verb executes correctly but the notification is not found? (Just asking the question but I think my personal preference would be to receive a null for not found, and an exception if there is an exception.)

at_client/lib/src/service/notification_service_impl.dart
```diff
@@ -408,4 +412,37 @@ class NotificationServiceImpl
...
+      throw AtClientException.message('Failed to fetch the notification id',
+          intent: Intent.remoteVerbExecution,
+          exceptionScenario: ExceptionScenario.remoteVerbExecutionFailed);
+    }
+    AtResponse atResponse = DefaultResponseParser().parse(atNotificationStr);
+    var atNotificationMap = jsonDecode(atResponse.response);
+    if (atNotificationMap['notificationStatus'] ==
+        at_persistence_secondary_server.NotificationStatus.expired.toString()) {
+      return AtNotification.empty()
```
gkc: Why the special handling for 'expired' status? i.e. why not just pass the full notification back regardless of status?

gkc: Oh I think I know the answer to my question - it's because the server only returns the id and the status in this case.

at_end2end_test/test/notify_test.dart
```diff
@@ -108,42 +105,29 @@ void main() {
...
+                .getInstance()
+                .notificationService
+                .fetch(notificationResult.notificationID);
+            atNotification.isEncrypted = input.atKey.metadata!.isEncrypted;
+            await NotificationResponseTransformer().transform(Tuple()
+              ..one = atNotification
+              ..two = (NotificationConfig()
+                ..shouldDecrypt = input.atKey.metadata!.isEncrypted!));
+            expect(atNotification.id, notificationResult.notificationID);
+            expect(atNotification.key, '$whomToNotify:$notifyText');
```
gkc: We already know the expected output, please change to `expect(atNotification.key, expectedOutput);`

gkc (changes requested):
A few comments /. questions /. changes requested.

Couple of other things:
* please add to the functional test suite in at_functional_test/test/atclient_notify_test.dart
* please add functional and end to end tests for fetches on expired notifications
* please add functional and end to end tests for fetches on non-existent notifications
* please add unit tests for every exception scenario for fetch

at_client/lib/src/service/notification_service_impl.dart
sitaram-kalluri: In case of notification id does exist, servers return NotificationStatus.expired.

The above IF condition that checks for null is when remote secondary fails to respond back due to network issue or secondary not reacheable.

at_end2end_test/test/notify_test.dart
sitaram-kalluri: Updated with the expectedOutput

sitaram-kalluri:
> A few comments /. questions /. changes requested.
> 
> Couple of other things:
> 
> * please add to the functional test suite in at_functional_test/test/atclient_notify_test.dart
> * please add functional and end to end tests for fetches on expired notifications
> * please add functional and end to end tests for fetches on non-existent notifications
> * please add unit tests for every exception scenario for fetch

* Added tests in at_functional_tests
* Added test in e2e for non-existent notifications
* Added unit tests for exception scenario
* At the moment, we do not have a way to set the expiry on the notification. We have a [expose-notification-params](https://github.com/atsign-foundation/at_client_sdk/pull/704)  PR which allows to add expiry to notifications. Can we add fetch notification on expiry in that PR?

gkc:
> * At the moment, we do not have a way to set the expiry on the notification. We have a [expose-notification-params](https://github.com/atsign-foundation/at_client_sdk/pull/704)  PR which allows to add expiry to notifications. Can we add fetch notification on expiry in that PR?

I don't understand ... we can set ttln on a notification. When time-to-live is set, then when the time-to-live duration has passed, the notification has expired, no?

sitaram-kalluri:
> > * At the moment, we do not have a way to set the expiry on the notification. We have a [expose-notification-params](https://github.com/atsign-foundation/at_client_sdk/pull/704)  PR which allows to add expiry to notifications. Can we add fetch notification on expiry in that PR?
> 
> I don't understand ... we can set ttln on a notification. When time-to-live is set, then when the time-to-live duration has passed, the notification has expired, no?

At the moment, in at_client SDK, we do not have a way to set ttln on the NotificationParams. We have an open [PR](https://github.com/atsign-foundation/at_client_sdk/pull/704) which exposes notificationExpiry.

gkc (dismissed):

at_client/lib/src/service/notification_service_impl.dart
```diff
@@ -409,4 +415,42 @@ class NotificationServiceImpl
...
     return _streamListeners.length;
   }
+
+  @override
+  Future<AtNotification> fetch(String notificationId) async {
+    var notifyFetchVerbBuilder = NotifyFetchVerbBuilder()
+      ..notificationId = notificationId;
+    String? atNotificationStr;
+    try {
+      atNotificationStr = await _atClient
```
murali-shris: did you test this logic and how exception is handled when there is no connectivity on client side?

sitaram-kalluri: Since all the verbs thats needs to run on remote secondary needs network connectivity, have added a check - that checks for internet connectivity, in executeVerb method in RemoteSecondary.dart file. 

@VJag and @gkc : Can you plz review the change and let me know if this approach is fine?

gkc: No. Checking for internet connectivity does not mean that the connection is still there by the time you execute the verb, the check is therefore pointless. 

I would echo what Murali asked which is, did you test the logic and how the exception is handled **when** there is no connectivity?

sitaram-kalluri: Removed the condition that check for internet connectivity in remoteSecondary-> executeVerb method.
Further, tested the scenario, where the network connection is down on client side for notify fetch and `AtTimeOutException` is thrown.

```
SEVERE|2022-10-17 12:12:23.332957|AtLookup|Exception in sending to server, Exception: Waited for 10000 millis. No response after 2022-10-17 12:12:13.315518  
```

When secondary is down, some times a AtTimeoutException is thrown with error message as above. But, sometimes i did observe a broken pipe error (pasting the error below). Need to debug further on from where the broken pipe error originates. Also, i observed the below when running from the samples, need to check how the app  behaves in the same situation.

```
Failed to load "/home/sitaram/IdeaProjects/atsign/core/at_client_sdk/packages/at_client/test/samples/remote_llookup.dart": Broken pipe (OS Error: Broken pipe, errno = 32), address = vip.ve.atsign.zone, port = 42664
```

packages/at_client/lib/src/service/notification_service.dart
```diff
@@ -101,6 +101,9 @@ abstract class NotificationService {
 
   /// Returns the status of the given notificationId
   Future<NotificationResult> getStatus(String notificationId);
+
+  /// Returns the [AtNotification] of the given notificationId
```
VJag, line 121: Lets document various return types for various inputs and exceptions thrown.

Ex: 1. A notification with a given id exists
      2. When it does not exist
      3. Error in fetching the values etc...

packages/at_client/lib/src/service/notification_service.dart
```diff
@@ -101,6 +101,31 @@ abstract class NotificationService {
...
+  ///  operation: OperationType.update, status: NotificationStatus.delivered}
+  /// ```
+  ///
+  /// Returns [NotificationStatus.expired] when
+  /// * Notification id does not exist
+  /// * Notification is expired / delete
+  /// ```
+  ///  AtNotification{id: abc-123, status: NotificationStatus.expired}
+  /// ```
+  /// Throws [AtClientException] when server is not reachable or server timeout's to respond
```
VJag, line 143: What does it return for nonexistent notification?

sitaram-kalluri: For a non-existent notification also, we return `AtNotification{id: abc-123, status: NotificationStatus.expired}`.

Currently, we do not have a way to differentiate between if the notification is expired and deleted or if the notification does not exist. So, returning the same response for both.

Updated the same in the dart docs too.

packages/at_client/lib/src/service/notification_service.dart
```diff
@@ -103,6 +103,28 @@ abstract class NotificationService {
   Future<NotificationResult> getStatus(String notificationId);
 
   /// Returns the [AtNotification] of the given notificationId
+  ///
+  /// Usage:
+  ///
+  ///  To fetch a notification with id: abc-123
+  /// ```dart
+  /// var notification = NotificationServiceImpl(atClient!);
```
gkc: This isn't how we create a NotificationServiceImpl right now; we expose NotificationServiceImpl.create(...)

gkc: You could just write `NotificationService notifSvc = <code to get handle to notification service>` so that the docs stay current

packages/at_client/lib/src/service/notification_service.dart
gkc: I think this is OK

at_client/lib/src/service/notification_service_impl.dart
gkc: @kalluriramkumar Have you done the follow-up you mentioned in your comment above?

gkc (commented):
Need to complete the follow-up you mentioned in response to @murali-shris's questions about exception handling

at_client/lib/src/service/notification_service_impl.dart
kalluriramkumar: @gkc: Debugged on the issue mentioned above:
1. Observe the broken pipe issue during the data transfer when the connection is closed (shutdown secondary server). We are replicating the issue by placing the breakpoints.
2. In other scenarios getting either the successful message or the notification expiry message.

sitaram-kalluri: @gkc: Debugged the issue mentioned above:
1. The broken pipe issue occurs when the connection is closed during the data transfer (Shutdown of the secondary server).
We tested this by placing the breakpoints.
2. The other cases, either the successful message or the notification expiry message is received.

Tested scenario -1 with wavi app and the following are observations:
1. The app is stable and no sign's of an app crash
2. The app logs have an error message stating the connection is closed.

```
I/flutter ( 4334): Unable to fetch latest server commit id: Exception: HandshakeException: Connection terminated during handshake 
E/flutter ( 4334): [ERROR:flutter/runtime/dart_vm_initializer.cc(41)] Unhandled Exception: SocketException: Broken pipe (OS Error: Broken pipe, errno = 32), address = vip.ve.atsign.zone, port = 42338
E/flutter ( 4334): 
I/flutter ( 4334): INFO|2022-10-27 10:03:00.395937|AtLookup|Creating new connection 
```

packages/at_client/lib/src/service/notification_service.dart
sitaram-kalluri: @gkc: Updated the dart docs as mentioned above.

gkc (approved):
This LGTM. @VJag also had comments & questions, please don't merge until @VJag has also re-reviewed & approved

sitaram-kalluri:
> This LGTM. @VJag also had comments & questions, please don't merge until @VJag has also re-reviewed & approved

Sure Gary

VJag (approved):

PR #739, opened 2022-10-05
fix: Fix behaviour associated with the 'fetchOfflineNotifications' flag

gkc:
**- What I did**
Corrected behaviour of the recently-added 'fetchOfflineNotifications' flag. Expected behaviour was:
* Flag set to true - current behaviour - monitor verb requests the server for all notifications after locally-stored lastNotificationTime
* Flag set to false - monitor verb requests server for all notifications after _right now_
* Actual behaviour when set to false was that the monitor verb requests server for all previously-received notifications - i.e. the opposite of intended

Also made change so that the fetchOfflineNotifications flag is honoured in the event of the monitor reconnecting after disconnecting. (For example, let's say your phone loses network connectivity for 4 hours but the app is running for all of that time - if fetchOfflineNotifications flag is still false when the monitor reconnects, then we will again request server for all notifications after _right now_. If an app happens to have behaviour where it wants to sometimes have fetchOfflineNotifications be false and sometimes true, the app should have logic to change the flag.)

**- How I did it**
* In order to have the behaviour be consistent, moved the check of fetchOfflineNotifications into the `_getLastNotificationTime()` method

**- How to verify it**
* Tests pass

* Corrected behaviour when the recently-added 'fetchOfflineNotifications' flag is set to false

### DIFF
--- a/at_client/lib/src/service/notification_service_impl.dart
+++ b/at_client/lib/src/service/notification_service_impl.dart
@@ -101,6 +101,7 @@ class NotificationServiceImpl
     }
   }
 
+
   Future<int?> _getLastNotificationTime() async {
     if (AtClientManager.getInstance().atClient.getPreferences()!.fetchOfflineNotifications == false) {
       // Don't fetch notifications which were received by the server while this client was offline ...
--- a/at_client/lib/src/service/notification_service_impl.dart
+++ b/at_client/lib/src/service/notification_service_impl.dart
@@ -93,15 +93,8 @@ class NotificationServiceImpl
           'monitor is already started for ${_atClient.getCurrentAtSign()}');
       return;
     }
-    if (AtClientManager.getInstance()
-        .atClient
-        .getPreferences()!
-        .fetchOfflineNotifications) {
-      final lastNotificationTime = await _getLastNotificationTime();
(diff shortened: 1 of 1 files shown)


at_client/lib/src/service/notification_service_impl.dart
```diff
@@ -93,22 +93,24 @@ class NotificationServiceImpl
...
       _isMonitorPaused = false;
     }
   }
 
+
   Future<int?> _getLastNotificationTime() async {
+    if (AtClientManager.getInstance().atClient.getPreferences()!.fetchOfflineNotifications == false) {
+      // Don't fetch notifications which were received by the server while this client was offline ...
+      // ... means ask only for notifications received after RIGHT NOW
+      return DateTime.now().millisecondsSinceEpoch;
```
VJag, line 109: I was wondering if the difference in now() of client and server could be a problem. Not passing any value is understood as after now() by the server.

gkc: OK this whole PR is wrong then 🤡

Need to do a much deeper look at what's going on. What Colin & Kevin observed is that when they set this new flag to "false" then the client would not subsequently sync.

I'll close this PR and create another ticket.

PR #748, opened 2022-10-12
build: uptake persistence secondary server 3.0.39

sitaram-kalluri:
**- What I did**
- Uptake at_persistence_secondary_server to 3.0.38 version
- Add an addition condition that verifies if serverMetadata contains isEncrypted

**- How I did it**

**- How to verify it**

### DIFF
--- a/at_client/lib/src/service/sync_service_impl.dart
+++ b/at_client/lib/src/service/sync_service_impl.dart
@@ -460,7 +460,7 @@ class SyncServiceImpl implements SyncService, AtSignChangeListener {
       }
       final serverEncryptedValue = serverCommitEntry['value'];
       final serverMetaData = serverCommitEntry['metadata'];
-      if (serverMetaData != null && (serverMetaData[IS_ENCRYPTED] != null && serverMetaData[IS_ENCRYPTED] == "true")) {
+      if (serverMetaData != null && serverMetaData[IS_ENCRYPTED] == "true") {
         final decryptionManager =
             AtKeyDecryptionManager.get(atKey, _atClient.getCurrentAtSign()!);
         final serverDecryptedValue =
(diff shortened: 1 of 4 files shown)


murali-shris (dismissed):

at_client/lib/src/service/sync_service_impl.dart
```diff
@@ -460,7 +460,7 @@ class SyncServiceImpl implements SyncService, AtSignChangeListener {
       }
       final serverEncryptedValue = serverCommitEntry['value'];
       final serverMetaData = serverCommitEntry['metadata'];
-      if (serverMetaData != null && serverMetaData[IS_ENCRYPTED] == "true") {
+      if (serverMetaData != null && (serverMetaData[IS_ENCRYPTED] != null && serverMetaData[IS_ENCRYPTED] == "true")) {
```
gkc: You already had `serverMetaData[IS_ENCRYPTED] == "true"` - by definition, if those values are equal then `serverMetaData[IS_ENCRYPTED] != null` is also true. i.e. this change makes no difference. What problem are you actually trying to solve?

gkc (changes requested):

at_client/lib/src/service/sync_service_impl.dart
kalluriramkumar: If the metadata for the older keys does not have 'IS_ENCRYPTED', then serverMetaData[IS_ENCRYPTED] might result to a null pointer exception. So, having an additional check  to prevent a null pointer

sitaram-kalluri: Original intent we to solve the null pointer exception if serverMetadata does not have IS_ENCRYPTED. However; since, we do not further resolve on `serverMetadata[IS_ENCRYPTED]` even if its is null, still the condition fails but we will not encounter null pointer exception

gkc (approved):

Issue #742, opened 2022-10-06
Synchronisation error


ralfeus:
**Describe the bug**
After some moment, probably after the key is expired the errors are logged every 0.15 seconds:

```
SEVERE|2022-10-06 15:22:04.682472|SyncService|exception syncing entry to local {atKey: @baboon9blue:memberinviterequest.atkrypto@81bored7, value: null, metadata: null, commitId: 125, operation: -} - Failed to decryptData caused by
Decryption failed. Encrypted value is null 

SEVERE|2022-10-06 15:22:04.846217|SyncService|exception syncing entry to local {atKey: @baboon9blue:memberinviterequest.atkrypto@81bored7, value: null, metadata: null, commitId: 125, operation: -} - Failed to decryptData caused by
Decryption failed. Encrypted value is null 
```

When the key with same name is created again it's never synchronised to the remote secondary.

**To Reproduce**
1. Receive a key with TTL 60 seconds. The code to send a key:
```
  final Metadata _defaultMetadata = Metadata( 👍 
    ..isEncrypted = true
    ..isPublic = false
    ..namespaceAware = true
    ..ttl = 60000
    ..ttr = -1;
    
    Future<bool> send(String recipient, String key, dynamic value) async {
    var atKey = AtKey()
      ..key = key
      ..sharedBy = client.getCurrentAtSign()
      ..sharedWith = recipient
      ..metadata = _defaultMetadata;
    Logger.root.finest(atKey.key);
    final result = await client.put(atKey, value);
    return result;
  }
```  
2. Have the recipient to get the key. The receiving code:
```dart
  void listen() async {
    final atStream =
        AtClientManager.getInstance().notificationService.subscribe();
    await for (final notification in atStream) {
      // whatever
    }
  }
```
3. Wait the key to expire.
4. Observe errors above

**Expected behavior**
No errors upon key's expiration

gkc:
Thanks @ralfeus - an unexpected consequence of some fairly recent changes. We'll fix the bug and add some tests to cover this scenario.

ralfeus:
Not sure if it's related or a different issue - please advice. 
I put a key `memberInviteRequest` shared with `@81bored7` to a store. Then it's supposed to be synchronised and I see following in the log:

```
I/flutter (20529): INFO|2022-10-10 00:25:10.028272|AtLookup|Creating new connection
I/flutter (20529): INFO|2022-10-10 00:25:10.856429|AtLookup|auth success
[...]

PR #754, opened 2022-10-14
fix: sync conflict - check server value for null before decyption

murali-shris:
**- What I did**
- fix for exception caused when server value is null
**- How I did it**
- added a check in sync conflict to check the server value for null/empty before decrypting.
**- How to verify it**
- run the functional test [atclient_sync_conflict_test.dart](https://github.com/atsign-foundation/at_client_sdk/compare/trunk...sync_conflict_decryption_value_check#diff-da7a6aa9c49b0c15a4bae124b9490c92e993e2f1ec6156292f093bed40a4bc40)

Fixes the exception reported in bug https://github.com/atsign-foundation/at_client_sdk/issues/742


at_functional_test/test/atclient_sync_conflict_test.dart
```diff
@@ -33,6 +32,31 @@ void main() async {
     }
     await Future.delayed(Duration(seconds: 10));
   });
+
+  // server value is null, client has non value for same key.
```
gkc: should be `client has non-null value`

at_functional_test/test/atclient_sync_conflict_test.dart
```diff
@@ -33,6 +32,31 @@ void main() async {
...
+  // you should get a sync conflict populated with no exception thrown.
+  test('server value is null', () async {
```
gkc: The test is good. However, it did take me quite a few minutes of reading and re-reading the code before I could figure out what was going on. Please can you add some comments to explain what is going on?

gkc (changes requested):

at_functional_test/test/atclient_sync_conflict_test.dart
purnimavenkatasubbu: Updated the comments and pushed the changes.

at_functional_test/test/atclient_sync_conflict_test.dart
purnimavenkatasubbu: Done

gkc (approved):

gkc:
Nice work, @murali-shris and @purnimavenkatasubbu

Issue #713, opened 2022-09-19
Notifications get hung up as secondary syncs


cconstab:
**Describe the bug**
When sending frequent updates (say 2hertz) there is noticeable delay in notifications received in the client.

**To Reproduce**
1. First I send a lot of notifications to a secondary (I used the GPS Demo code)
2. Then I connect a client and subscribe to nofications
3. As the client is synching no notifications are being dealt with and so they get hung up until sync completes
4. 

**Expected behavior**
Notifications should not be held up by the sync process

**Screenshots**

https://user-images.githubusercontent.com/6131216/191070158-9d74f280-b9a7-418d-b13d-f1876a17af92.mp4

**Were you using an atApplication when the bug was found?**
  - GPS Demo is what I used

**Additional context**

cconstab:
Much improved by removing the old hive boxes on the receiver and changing the onboarding Pref to 

```
  //onboarding preference builder can be used to set onboardingService parameters
  AtOnboardingPreference atOnboardingConfig = AtOnboardingPreference()
    ..hiveStoragePath = '$homeDirectory/.$nameSpace/$receivingAtsign/$deviceName/storage'
    //..syncBatchSize = 1
    ..syncIntervalMins = 10
    ..syncRegex = "ai6bh"
    ..namespace = nameSpace
    ..downloadPath = '$homeDirectory/.$nameSpace/files'
    ..isLocalStoreRequired = true
    ..commitLogPath = '$homeDirectory/.$nameSpace/$receivingAtsign/$deviceName/storage/commitLog'
    ..fetchOfflineNotifications = false
    ..monitorHeartbeatInterval = Duration(seconds: 60)
    ..atKeysFilePath = atsignFile;
    ```

**Comment by cconstab** [comment]
After leaving over night the symptoms slowly come back, what I also observed is that the monitor can also die in this hang which leads to more delays.

```
WARNING|2022-09-20 13:55:38.408682|Monitor|Heartbeat response not received within expected duration. Heartbeat was sent at 1663682137536, it is now 1663682138408, last heartbeat response was received at 1663682114992. Will close connection, set status stopped, call retryCallback

INFO|2022-09-20 13:55:38.409651|Monitor|socket.listen onDone called. Will destroy socket, set status stopped, call retryCallback

INFO|2022-09-20 13:55:38.410344|Monitor|Another closeStopAndRetry operation is in progress

[...]

PR #760, opened 2022-10-16
feat: Introduce a local key for latestNotificationId so it is not sync'd to server

### LINKED ISSUE #750: Skip latestNotificationId to sync to server

At the moment, the latestNotificationId keys gets sync'ed the server. The key gets updated for every notification received and as a result the key gets continuously sync'ed to the server which created a lot of load on the server. Further, server does not need to know the latestNotificationId.

Proposed Solution:
-------------------------
1. Enhance the latestNotificationId as a local key.
2. In the at_persistence_secondary_server, in the commit method, modify the IF condition such that local keys are **NOT** added to commit log. 
3. In the at_client SDK, update the latestNotificationId as a local key. Fetch from the old 'latestNotificationId' and set the value as a default value to the latestNotificationId and store it to the keystore.
4. Do not update/fetch the data from the old 'latestNotificationId'. So, that old key becomes unusable and do not get sync'ed continuously.

Backward Compatibility:
--------------------------------
1. Since the older key is not refrained in the commit log, in case of old client and new server (or) new client and old server, the key gets still sync'ed without causing the sync imbalance.

sitaram-kalluri:
**- What I did**
- Skip latestNotificationId from the commit log to refrain the key to sync between client and server

**- How I did it**
- Introduce a new key `local:lastReceivedNotification` that do not sync between the cloud and local secondary. (Local keys do not sync between cloud and local secondaries)
- In getLastNotificationTime, check if the new key (local:lastReceivedNotification) is available. If key is available and value is not null, return the epochMillis. If key is not available, get the data from old key (_latestNotificationId) and update the new key with the data.

- Refactor the existing unit tests as follows
   * Moved isKeyExists from MockSecondaryKeyStore class to test : isKeyExists method needs to true/false differently for tests. So moved to tests from common place to get the desired behaviour in tests.
   * Moved get and put methods from MockAtClientImpl to tests: To achieve the desired behaviour of get and put methods specific to tests, moved from class to tests.
   * Moved ` mockPreferences.fetchOfflineNotifications = true;` to tests - To set the fetchOfflineNotifications to true/false for tests, removed from the class level
- Added the following unit tests to verify the getNotificationLastTime behaviour
  * Test 1: When the lastReceivedNotificationKey(new key) and _latestNotificationIdv2(old key) are not available, null value is returned
  * Test 2: When the lastReceivedNotificationKey(new key) is not available, last received notification date time is returned from the old key
  * Test 3: When the lastReceivedNotificationKey(new key) is available, last received notification date time is returned from the new key
  * Test 4: test to verify toString called on lastNotificationReceived key returns correct string representation
  * Test 5: test to verify fromString on lastNotificationReceived key returns correct AtKey instance

- Following files do not have code changes. Only dart formatting changes:
   * at_client_spec.dart
   * at_client_impl.dart
   * sync_service_impl.dart
   * sync_service_test.dart

**- How to verify it**
- Received notifications, latestNotificationId, should not be updated and synced to server.
- The local:lastReceivedNotification should be updated and not sync to the secondary

- Deprecate latestNotificationId. Introduce a local key - local:lastReceivedNotification to fetch/store the last received notification date time

### DIFF
--- a/packages/at_client/lib/src/service/notification_service_impl.dart
+++ b/packages/at_client/lib/src/service/notification_service_impl.dart
@@ -2,10 +2,8 @@ import 'dart:async';
 import 'dart:collection';
 import 'dart:convert';
 
-import 'package:at_client/src/client/at_client_spec.dart';
+import 'package:at_client/at_client.dart';
 import 'package:at_client/src/encryption_service/encryption_manager.dart';
-import 'package:at_client/src/manager/at_client_manager.dart';
-import 'package:at_client/src/listener/connectivity_listener.dart';
 import 'package:at_client/src/listener/at_sign_change_listener.dart';
 import 'package:at_client/src/listener/switch_at_sign_event.dart';
 import 'package:at_client/src/manager/monitor.dart';
@@ -13,15 +11,11 @@ import 'package:at_client/src/preference/monitor_preference.dart';
 import 'package:at_client/src/response/default_response_parser.dart';
 import 'package:at_client/src/response/notification_response_parser.dart';
 import 'package:at_client/src/response/response.dart';
-import 'package:at_client/src/service/notification_service.dart';
 import 'package:at_client/src/transformer/request_transformer/notify_request_transformer.dart';
 import 'package:at_client/src/transformer/response_transformer/notification_response_transformer.dart';
-import 'package:at_client/src/util/at_client_util.dart';
 import 'package:at_client/src/util/at_client_validation.dart';
 import 'package:at_client/src/util/regex_match_util.dart';
-import 'package:at_client/src/response/at_notification.dart';
 import 'package:at_commons/at_builders.dart';
-import 'package:at_commons/at_commons.dart';
 import 'package:at_persistence_secondary_server/at_persistence_secondary_server.dart'
     as at_persistence_secondary_server;
 import 'package:at_utils/at_utils.dart';
@@ -33,6 +27,7 @@ class NotificationServiceImpl
       HashMap(equals: _compareNotificationConfig, hashCode: _generateHashCode);
   final emptyRegex = '';
   static const notificationIdKey = '_latestNotificationIdv2';
+  static const lastReceivedNotificationKey = 'lastReceivedNotification';
   static final Map<String, NotificationService> _notificationServiceMap = {};
 
   final _logger = AtSignLogger('NotificationServiceImpl');
@@ -44,6 +39,9 @@ class NotificationServiceImpl
   late AtClientManager _atClientManager;
   AtClientValidation atClientValidation = AtClientValidation();
   AtKeyEncryptionManager atKeyEncryptionManager = AtKeyEncryptionManager();
+
+  @visibleForTesting
+  late AtKey lastReceivedNotificationAtKey;
 
   static Future<NotificationService> create(AtClient atClient,
       {required AtClientManager atClientManager, Monitor? monitor}) async {
@@ -70,6 +68,10 @@ class NotificationServiceImpl
             MonitorPreference()..keepAlive = true,
             _monitorRetry);
     _atClientManager.listenToAtSignChange(this);
+    lastReceivedNotificationAtKey = AtKey.local(lastReceivedNotificationKey,
+            _atClientManager.atClient.getCurrentAtSign()!,
+            namespace: _atClientManager.atClient.getPreferences()!.namespace)
+        .build();
   }
 
   Future<void> _init() async {
@@ -106,6 +108,12 @@ class NotificationServiceImpl
   }
 
   @visibleForTesting
+
+  /// Return the last received notification DateTime in epochMillis when
+  /// [AtClientPreference.fetchOfflineNotifications] is set true.
+  ///
+  /// Returns null when the key which holds the lastNotificationReceived
+  /// does not exist.
   Future<int?> getLastNotificationTime() async {
     if (_atClientManager.atClient.getPreferences()!.fetchOfflineNotifications ==
         false) {
@@ -116,24 +124,37 @@ class NotificationServiceImpl
 
     // fetchOfflineNotifications == true (the default) means we want all notifications since the last one we received
     // We keep track of the last notification id in the client-side key store
-    var lastNotificationKeyStr =
-        '$notificationIdKey.${_atClient.getPreferences()!.namespace}${_atClient.getCurrentAtSign()}';
-    var atKey = AtKey.fromString(lastNotificationKeyStr);
+    // Check if the new key (local:lastNotificationReceived@alice) is available in the keystore.
+    // If yes, fetch the value;
+    AtValue? atValue;
     if (_atClient
         .getLocalSecondary()!
         .keyStore!
-        .isKeyExists(lastNotificationKeyStr)) {
-      AtValue? atValue;
-      try {
-        atValue = await _atClient.get(atKey);
-      } on Exception catch (e) {
-        _logger
-            .severe('Exception in getting last notification id: ${e.toString}');
+        .isKeyExists(lastReceivedNotificationAtKey.toString())) {
+      atValue = await _atClient.get(lastReceivedNotificationAtKey);
+    }
+    // If new key does not exist or value is null, check for the old key (_latestNotificationIdv2@alice)
+    // If old key exist, fetch the value and update the new key with old key's value
+    if (atValue == null || atValue.value == null) {
+      var lastNotificationKeyStr =
+          '$notificationIdKey.${_atClient.getPreferences()!.namespace}${_atClient.getCurrentAtSign()}';
+      var atKey = AtKey.fromString(lastNotificationKeyStr);
+      if (_atClient
+          .getLocalSecondary()!
+          .keyStore!
+          .isKeyExists(lastNotificationKeyStr)) {
+        try {
+          atValue = await _atClient.get(atKey);
+          await _atClient.put(lastReceivedNotificationAtKey, atValue.value);
+        } on Exception catch (e) {
+          _logger.severe(
+              'Exception in getting last notification id: ${e.toString}');
+        }
       }
-      if (atValue != null && atValue.value != null) {
-        _logger.finer('json from hive: ${atValue.value}');
-        return jsonDecode(atValue.value)['epochMillis'];
-      }
+    }
+    if (atValue?.value != null) {
+      _logger.finer('json from hive: ${atValue?.value}');
+      return jsonDecode(atValue?.value)['epochMillis'];
     }
     return null;
   }
@@ -157,7 +178,7 @@ class NotificationServiceImpl
       for (var atNotification in atNotifications) {
         // Saves latest notification id to the keys if its not a stats notification.
         if (atNotification.id != '-1') {
-          await _atClient.put(AtKey()..key = notificationIdKey,
+          await _atClient.put(lastReceivedNotificationAtKey,
               jsonEncode(atNotification.toJson()));
         }
         _streamListeners.forEach((notificationConfig, streamController) async {
--- a/packages/at_client/lib/src/service/sync_service_impl.dart
+++ b/packages/at_client/lib/src/service/sync_service_impl.dart
@@ -699,7 +699,7 @@ class SyncServiceImpl implements SyncService, AtSignChangeListener {
       case '#':
       case '*':
         var builder = UpdateVerbBuilder()
-          ..atKey = serverCommitEntry['atKey']
+          ..atKeyObj = AtKey.fromString(serverCommitEntry['atKey'])
           ..value = serverCommitEntry['value'];
         builder.operation = UPDATE_ALL;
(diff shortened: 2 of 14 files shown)


packages/at_client/lib/src/util/at_client_util.dart
```diff
@@ -11,6 +11,9 @@ import 'package:crypton/crypton.dart';
 class AtClientUtil {
   static String buildKey(UpdateVerbBuilder builder) {
     var updateKey = '';
+    if(builder.isLocal){
```
murali-shris: dart format missing

packages/at_client/lib/src/service/notification_service_impl.dart
```diff
@@ -102,33 +112,47 @@ class NotificationServiceImpl
...
-        _logger
-            .severe('Exception in getting last notification id: ${e.toString}');
-      }
-      if (atValue != null && atValue.value != null) {
-        _logger.finer('json from hive: ${atValue.value}');
-        return jsonDecode(atValue.value)['epochMillis'];
+        .isKeyExists(lastReceivedNotification.toString())) {
+      atValue = await _atClient.get(lastReceivedNotification);
+    }
+    // If new key does not exist or value is null, check for the old key (_latestNotificationId@alice)
```
murali-shris: in the comment..should be _latestNotificationIdv2@alice ?

packages/at_client/lib/src/service/notification_service_impl.dart
```diff
@@ -102,33 +112,47 @@ class NotificationServiceImpl
...
+        try {
+          atValue = await _atClient.get(atKey);
+          await _atClient.put(lastReceivedNotification, atValue.value);
+        } on Exception catch (e) {
+          _logger.severe(
+              'Exception in getting last notification id: ${e.toString}');
+        }
       }
     }
+    if (atValue != null && atValue.value != null) {
```
murali-shris: you can replace the condition with atValue?.value != null

sitaram-kalluri: Updated the condition.

packages/at_client/lib/src/service/notification_service_impl.dart
sitaram-kalluri: Updated the comment.

packages/at_client/lib/src/util/at_client_util.dart
sitaram-kalluri: Ran dart format and pushed the changes

packages/at_client/lib/src/util/at_client_util.dart
```diff
@@ -11,6 +11,9 @@ import 'package:crypton/crypton.dart';
 class AtClientUtil {
   static String buildKey(UpdateVerbBuilder builder) {
     var updateKey = '';
+    if (builder.isLocal) {
```
VJag: Where does the key being both local, public, and shared prevented?

packages/at_client/lib/src/transformer/request_transformer/put_request_transformer.dart
```diff
@@ -72,9 +72,9 @@ class PutRequestTransformer
       ..isEncrypted = (atKey.metadata?.isEncrypted != null)
           ? atKey.metadata?.isEncrypted!
           : false
-      ..isBinary = (atKey.metadata?.isBinary != null)
-          ? atKey.metadata?.isBinary!
-          : false;
+      ..isBinary =
```
VJag, line 75: I did not get this line, why isBinary and isLocal mixed in the statement.

packages/at_client/lib/src/service/notification_service_impl.dart
```diff
@@ -102,33 +112,47 @@ class NotificationServiceImpl
...
-      try {
-        atValue = await _atClient.get(atKey);
-      } on Exception catch (e) {
-        _logger
-            .severe('Exception in getting last notification id: ${e.toString}');
-      }
-      if (atValue != null && atValue.value != null) {
-        _logger.finer('json from hive: ${atValue.value}');
-        return jsonDecode(atValue.value)['epochMillis'];
+        .isKeyExists(lastReceivedNotification.toString())) {
```
VJag: Can you add some tests around toString behavior of lastReceivedNotification key

packages/at_client/lib/src/service/notification_service_impl.dart
```diff
@@ -102,33 +112,47 @@ class NotificationServiceImpl
...
+          _logger.severe(
+              'Exception in getting last notification id: ${e.toString}');
+        }
       }
     }
+    if (atValue?.value != null) {
+      _logger.finer('json from hive: ${atValue?.value}');
+      return jsonDecode(atValue?.value)['epochMillis'];
+    }
     return null;
```
VJag, line 159: Just curious, how does the caller make use of a null return value?

VJag (changes requested):
Hi SIta. Pls check my comments.

sitaram-kalluri:
> Hi SIta. Pls check my comments.

Implemented the review comments and pushed the changes

packages/at_client/lib/src/util/at_client_util.dart
gkc: @VJag There is some guarding in AtKey but we could do with more.

Currently there is a single guard in the setter for isLocal in AtKey
```dart
  set isLocal(bool isLocal) {
    if (isLocal == true && sharedWith != null) {
      throw InvalidAtKeyException(
          'sharedWith should be empty when isLocal is set to true');
    }
    _isLocal = isLocal;
  }
```

We should have similar guard on the sharedWith setter.

Additional follow-up question: how do we ensure consistency not just isLocal and sharedWith but also of isPublic and isEncrypted?

gkc (changes requested):

gkc:
@kalluriramkumar Sorry I can't remember and I can't find it - is there a PR in at_persistence_secondary_server which implements "do not sync 'local' keys"?

gkc:
@kalluriramkumar ignore my last question - that doesn't belong in at_oersustence - belongs in the sync code on the client.

sitaram-kalluri:
@gkc/ @VJag : To ensure the consistency between isLocal, isPublic, isCached and sharedWith, we added validations in the verb builders - llookup_verb_builder and update_verb_builder (to start with).

Following is the draft PR: https://github.com/atsign-foundation/at_tools/pull/245

Please review and share your thoughts on this.

If the changes look good will push the corresponding in the at_client_impl - a minor change in local_secondary.dart which reduces the code redundancy to build the key (e.g: @‎alice:phone@‎bob) from the verb builders.

packages/at_client/test/notification_service_test.dart
```diff
@@ -521,55 +478,192 @@ void main() {
           .thenAnswer((_) => mockAtClientImpl);
     });
 
-    test('getLastNotificationTime() returns null if checkOfflineNotifications is set to false',
-            () async {
-          var notificationServiceImpl = await NotificationServiceImpl.create(
-              mockAtClientImpl,
-              atClientManager: mockAtClientManager,
-              monitor: mockMonitor) as NotificationServiceImpl;
+    test(
```
VJag: Just in words can we document what tests were written to test the code?

Ex:

Test 1: Sync when local is ahead and check that the right commit id is sent to the server
Test 2: Sync when the cloud is ahead and verify that the right commit id is sent to the server
Test 3: Repeat a put and delete 100 times and verify the commit id

PS: Tests are just an example

Copied from comment I made on Murali's PR. Just an example.

gkc:
@kalluriramkumar Where is the code which actually excludes 'local' keys from syncing?

sitaram-kalluri:
> @kalluriramkumar Where is the code which actually excludes 'local' keys from syncing?

@gkc: That is in at_persistence_secondary_server package, in at_commit_log.dart file. Following is the code snippet that omits the local keys from adding to the commit log.

```dart
if (!key.startsWith('public:__') &&
        (key.startsWith(RegExp('private:|privatekey:|public:_|local:')))) {
      return -1;
    }
```

cconstab:
What is the ETA on this merge ? The GPS Demo is still stalling as a result of this I think

https://github.com/atsign-foundation/at_client_sdk/issues/713

sitaram-kalluri:
> What is the ETA on this merge ? The GPS Demo is still stalling as a result of this I think
> 
> #713

@cconstab: We will target merging the changes to the trunk branch by early next week.

sitaram-kalluri:
Tested the changes with the prod atsigns on atmosphere pro app @snowboating37 and @81modern

The following are the observations:
1. The `lastReceivedNotification` is not sync'ed to the cloud secondary

```
@snowboating37@scan lastReceivedNotification
data:[]
```
2. Able to fetch the notification with the new key

<details>
<summary> screenshot of the test </summary>


Screenshot from 2022-11-09 12-28-40

</details>

3. Upgraded the at_client package to the feature branch version
    * Initially fetched the last notification date time from the old key and the data is inserted with the new key.
    * Subsequently fetched the last notification date time from the new key.

<details>
<summary>Screenshots of the test </summary>

upgrade-1

upgrade-2

</details>

packages/at_client/lib/src/util/at_client_util.dart
```diff
@@ -5,23 +5,16 @@ import 'package:at_client/at_client.dart';
...
 import 'package:crypton/crypton.dart';
 
 class AtClientUtil {
   static String buildKey(UpdateVerbBuilder builder) {
-    var updateKey = '';
-    if (builder.isPublic) {
-      updateKey += 'public:';
+    // If the key starts with privatekey:, it represents the internal key
+    // do not perform validation.
+    if (builder.atKey!.startsWith('privatekey:')) {
```
gkc: Three things:
(1) The builder's buildKey() should handle this case
(2) Why does this method exist at all? Why can't we deprecate this method and instead just use builder.buildKey()? The only place this method is used, as best I can tell, is in `LocalSecondary`'s `_update(...)` method where currently it does this `var updateKey = AtClientUtil.buildKey(builder);` and instead could do this `var updateKey = builder.buildKey();`

(3) The other thing I'm curious about, now that you've got an AbstractVerbBuilder which has an AtKeyObj, is why you can't have buildKey just return AtKeyObj.toString() ?

packages/at_client/lib/src/service/notification_service_impl.dart
```diff
@@ -45,6 +41,9 @@ class NotificationServiceImpl
   AtClientValidation atClientValidation = AtClientValidation();
   AtKeyEncryptionManager atKeyEncryptionManager = AtKeyEncryptionManager();
 
+  @visibleForTesting
+  static late AtKey lastReceivedNotification;
```
gkc: This should not be static

gkc: also please rename to 'lastReceivedNotificationAtKey'

packages/at_client/lib/src/util/sync_util.dart
```diff
@@ -117,7 +118,7 @@ class SyncUtil {
     if (key.startsWith(AT_PKAM_PRIVATE_KEY) ||
         key.startsWith(AT_PKAM_PUBLIC_KEY) ||
         key.startsWith(AT_ENCRYPTION_PRIVATE_KEY) ||
-        key.startsWith(statsNotificationId)) {
+        key.startsWith(NotificationServiceImpl.lastReceivedNotificationKey)) {
```
gkc, line 122: Just in case, please exclude the old key name also, i.e. change this to 
```
        key.startsWith(NotificationServiceImpl.notificationIdKey) ||
        key.startsWith(NotificationServiceImpl.lastReceivedNotificationKey)) {
```

gkc (changes requested):
Overall close to LGTM but there are a couple of important changes required, and I'd like to follow up with discussion of the questions I've asked re AtClientUtil.buildKey and UpdateVerbBuilder.buildKey

packages/at_client/lib/src/service/sync_service_impl.dart
```diff
@@ -699,7 +699,7 @@ class SyncServiceImpl implements SyncService, AtSignChangeListener {
       case '#':
       case '*':
         var builder = UpdateVerbBuilder()
-          ..atKey = serverCommitEntry['atKey']
+          ..atKeyObj = AtKey.fromString(serverCommitEntry['atKey'])
```
gkc, line 702: Just looking at this again ... would there ever be situations in which the server commit entry might contain an `atKey` which is invalid?

gkc (dismissed):

VJag (dismissed):

VJag (approved):

gkc:
Woo hoo! This is a big one - not just the fix but that it is using the "local" (client-side-only) keys feature which can be used for all sorts of similar client-side-only data in the future. Also using improved and better-tested verb handlers, better-tested AtKey validations

@cconstab FYI

@kalluriramkumar Let's aim to publish tomorrow

sitaram-kalluri:
The changes are published in at_client 3.0.44

PR #768, opened 2022-10-25
fix:  handle keystore delete changes<do not merge>

murali-shris:
**- What I did**
- Made changes in atclient to handle KeyNotFoundException thrown from remove method
**- How I did it**
- Modified _delete method in local_secondary.dart to return null when KeyNotFoundException is encountered
**- How to verify it**
- run unit tests in local_secondary_test.dart
**- Test scenarios covered
test 1: put a key using at_client and then delete the key. Delete should return true
test 2: put a key at_client, delete the key. Now try deleting the same key again. First delete should return true, second delete should return false.

### DIFF
--- a/packages/at_client/lib/src/service/sync_service_impl.dart
+++ b/packages/at_client/lib/src/service/sync_service_impl.dart
@@ -375,7 +375,8 @@ class SyncServiceImpl implements SyncService, AtSignChangeListener {
       List<CommitEntry> uncommittedEntries) async {
     // Iterates until serverCommitId is greater than localCommitId are equal.
     List<KeyInfo> keyInfoList = [];
-    while (serverCommitId > localCommitId) {
+    int lastReceivedServerCommitId = localCommitId;
+    while (serverCommitId > lastReceivedServerCommitId) {
       var syncBuilder = SyncVerbBuilder()
         ..commitId = localCommitId
         ..regex = _atClient.getPreferences()!.syncRegex
@@ -403,6 +404,7 @@ class SyncServiceImpl implements SyncService, AtSignChangeListener {
       }
       // Iterates over each commit
       for (dynamic serverCommitEntry in syncResponseJson) {
+        lastReceivedServerCommitId = serverCommitEntry['commitId'];
         try {
           final keyInfo =
               KeyInfo(serverCommitEntry['atKey'], SyncDirection.remoteToLocal);
@@ -420,9 +422,7 @@ class SyncServiceImpl implements SyncService, AtSignChangeListener {
               'error syncing entry to local $serverCommitEntry - ${e.toString()}');
         }
       }
-      // assigning the lastSynced local commit id.
-      localCommitId = await _getLocalCommitId();
-      _logger.finest('**localCommitId $localCommitId');
+      _logger.finest('**lastReceivedServerCommitId $lastReceivedServerCommitId');
     }
     return keyInfoList;
   }
--- a/packages/at_client/lib/src/service/sync_service_impl.dart
+++ b/packages/at_client/lib/src/service/sync_service_impl.dart
@@ -404,7 +404,12 @@ class SyncServiceImpl implements SyncService, AtSignChangeListener {
       }
       // Iterates over each commit
       for (dynamic serverCommitEntry in syncResponseJson) {
-        lastReceivedServerCommitId = serverCommitEntry['commitId'];
+        final receivedServerCommitId = serverCommitEntry['commitId'];
+        if (receivedServerCommitId is int) {
+          lastReceivedServerCommitId = receivedServerCommitId;
+        } else {
+          lastReceivedServerCommitId = int.parse(receivedServerCommitId);
+        }
         try {
           final keyInfo =
               KeyInfo(serverCommitEntry['atKey'], SyncDirection.remoteToLocal);
@@ -422,7 +427,8 @@ class SyncServiceImpl implements SyncService, AtSignChangeListener {
               'error syncing entry to local $serverCommitEntry - ${e.toString()}');
         }
       }
-      _logger.finest('**lastReceivedServerCommitId $lastReceivedServerCommitId');
+      _logger
+          .finest('**lastReceivedServerCommitId $lastReceivedServerCommitId');
     }
     return keyInfoList;
   }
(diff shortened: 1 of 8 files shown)


tests/at_functional_test/test/atclient_delete_test.dart
```diff
@@ -0,0 +1,69 @@
...
+  //   // To setup encryption keys
+  //   await setEncryptionKeys(atsign, preference);
+  //   // phone.me@alice🛠
+  //   var phoneKey = AtKey()
+  //     ..key = 'location'
+  //     ..metadata = (Metadata()..isPublic = true)..sharedBy=atsign;
+  //   var value = 'California';
+  //   var putResult = await atClient.put(phoneKey, value);
+  //   expect(putResult, true);
+  //   var getResult = await atClient.delete(phoneKey);
```
gkc: 1. A naming thing - should name this variable 'deleteResult' instead of 'getResult'
2. You can check server version and assert greater than x.y.z instead of commenting out the test. See here for existing examples of usage
```
find . -name "*.dart" -exec egrep -l 'getVersion *\(\)' {} \;                 
./at_server/tests/at_end2end_test/test/update_verb_test.dart
./at_server/tests/at_end2end_test/test/e2e_test_utils.dart
./at_server/tests/at_end2end_test/test/plookup_verb_test.dart
./at_server/tests/at_end2end_test/test/notify_verb_test.dart
```

tests/at_functional_test/test/atclient_delete_test.dart
```diff
@@ -0,0 +1,69 @@
...
+    await setEncryptionKeys(atsign, preference);
+    // public:phone.wavi@alice🛠
+    var phoneKey = AtKey()
+      ..key = 'location'
+      ..metadata = (Metadata()..isPublic = true)
+      ..sharedBy = atsign;
+    var value = 'California';
+    var putResult = await atClient.put(phoneKey, value);
+    expect(putResult, true);
+    var getResult = await atClient.delete(phoneKey);
```
gkc: Please name this variable 'deleteResult' not 'getResult'

gkc (changes requested):
Nicely done! Couple of suggested small changes.

tests/at_functional_test/test/atclient_delete_test.dart
murali-shris: done

tests/at_functional_test/test/atclient_delete_test.dart
murali-shris: i have implemented the test. Ran into this issue while running the tests
 https://github.com/atsign-foundation/at_server/issues/1005
 Had a discussion with @kalluriramkumar and tried  a fix. Have to verify the fix with Chris/Xavier.

murali-shris: issue is fixed

packages/at_client/lib/src/service/sync_service_impl.dart
```diff
@@ -403,6 +404,12 @@ class SyncServiceImpl implements SyncService, AtSignChangeListener {
       }
       // Iterates over each commit
       for (dynamic serverCommitEntry in syncResponseJson) {
+        final receivedServerCommitId = serverCommitEntry['commitId'];
```
VJag, line 412: Pls add dart docs the decisions here

packages/at_client/lib/src/service/sync_service_impl.dart
```diff
@@ -375,9 +375,10 @@ class SyncServiceImpl implements SyncService, AtSignChangeListener {
       List<CommitEntry> uncommittedEntries) async {
     // Iterates until serverCommitId is greater than localCommitId are equal.
     List<KeyInfo> keyInfoList = [];
-    while (serverCommitId > localCommitId) {
+    int lastReceivedServerCommitId = localCommitId;
+    while (serverCommitId > lastReceivedServerCommitId) {
       var syncBuilder = SyncVerbBuilder()
-        ..commitId = localCommitId
+        ..commitId = lastReceivedServerCommitId
```
VJag, line 386: Pls document what should be the right value of a commitId need to be

packages/at_client/test/local_secondary_test.dart
```diff
@@ -206,6 +206,34 @@ void main() {
           throwsA(isA<KeyNotFoundException>()));
     });
 
+    test('test delete verb builder - delete non existent key', () async {
```
VJag, line 209: Just in words can we document what tests were written to test the code?

Ex: 

Test 1: Sync when local is ahead and check that the right commit id is sent to the server
Test 2: Sync when the cloud is ahead and verify that the right commit id is sent to the server
Test 3: Repeat a put and delete 100 times and verify the commit id

PS: Tests are just an example

murali-shris: test 1: put a key using at_client and then delete the key. Delete should return true
test 2: put a key at_client, delete the key. Now try deleting the same key again. First delete should return true, second delete should return false.

packages/at_client/lib/src/service/sync_service_impl.dart
murali-shris: done

murali-shris:
@gkc encountered sync infinite loop issue on the client side in end2end tests. I have fixed that as well in this PR

packages/at_client/test/local_secondary_test.dart
VJag: @murali-shris tests in words as part of PR itself.

murali-shris:
@gkc I had a discussion with @VJag regarding infinite loop issue with the latest changes.
Scenario that causes infinite loop - https://docs.google.com/spreadsheets/d/1GKdrVnEReWHRIBAcQVsttCHJb9YOE_BM5LCDsHGrlEw/edit?usp=sharing

We discussed that if a delete comes from a server and key doesn't exist on app, app should honour the delete operation and update the commit id in local commit log instead of throwing KeyNotFoundException. This will eliminate the infinite loop issue.
We can decide on the final approach in the  architecture call tomorrow.

murali-shris:
changes in this PR are put on hold until sync behaviour is thoroughly documented and tested

gkc:
Closing as we decided against this approach

PR #774, opened 2022-11-01
feat: Refactor e2e tests

### LINKED ISSUE #732: end to end tests failing in at_client SDK
**Describe the bug**
The end2end tests are failing in at_client sdk and following are the observations

- The Hive keystore and commit log keystore are being removed and the end of test and being re-initialized for every test. As a result the initial sync happens before all the tests in a given file (in some cases before each test). which is time consuming and also blocking the server.

- The empty batch requests are being sent to secondary server: https://github.com/atsign-foundation/at_client_sdk/issues/731

Temporary workaround:
--------------------------------
Added skip for the failing tests. A permanent fix would be to refactor the tests.

purnimavenkatasubbu:
Fixes https://github.com/atsign-foundation/at_client_sdk/issues/732

**- What I did**
Refactored the client end-to-end tests in a way that the encryption keys are only set at the beginning of the tests and if the keys are already set, The tests will go ahead and run without setting the encryption keys again.

**- How I did it**
Added separate classes for the following,

**AtEncryptionKeysLoader**   
  _encryptionKeysCheckerMap -- has the info on whether the encryption keys have been set or not.
 If encryption keys are already set, it returns
 If keys are not already set, it goes ahead and sets the encryption keys

**SyncInitializer** - Responsible for sync and is used across all the end2end tests
**TestInitializer** - Responsible for the initial setup, which includes creating an atclientManager instance, and encryption keys setup.

Moved all the above classes to the lib/src folder and now the tests folder has only tests

**- How to verify it**
All the end2end tests should pass


murali-shris (approved):

gkc (approved):
Thanks!

PR #803, opened 2022-11-17
refactor: Use sync progress listener for KeyStream (WIP)

xavierchanth:
**- What I did**

KeyStream used to use a notification subscription for it's event trigger. Switching over to sync for reliability, and so that it works for self keys.

**- How I did it**

**- How to verify it**

refactor: Use sync progress listener for KeyStream


xavierchanth:
I don't have the understanding to complete this PR, what I don't understand is:
How to detect the difference between a created/updated key and a deleted key (in SyncProgress.keyInfo). I specifically need to be able to detect deleted keys so that I can remove them from the collection type KeyStreams (IterableKeyStream, ListKeyStream, etc.).

PR #807, opened 2022-11-21
fix: add sync unit test cases

sitaram-kalluri:
**- What I did**
1. Contains the sync service unit tests

**- How I did it**
1. Implemented unit tests to verify the sync service functionality

**- How to verify it**
1. The unit tests should cover the functionality of sync service


packages/at_client/test/sync_new_test.dart
```diff
@@ -0,0 +1,1922 @@
...
+      atData.data = encryptionPrivateKey;
+      //------------Operation---------------------------------
+      String atKey = (AtKey.self(AT_ENCRYPTION_PRIVATE_KEY, sharedBy: atsign))
+          .build()
+          .toString();
+      await keystore!.put(atKey, atData);
+      // -----------Assertions---------------------------------
+      // verifying the key in the key store
+      var keyStoreGetResult = await keystore.get(atKey);
+      expect(keyStoreGetResult!.data, encryptionPrivateKey);
```
VJag, line 785: There should also be an assertion that there no entry in the commit log

packages/at_client/test/sync_new_test.dart
```diff
@@ -0,0 +1,1922 @@
...
+    /// 1. There should be no entry for the same key in the key store
+    /// 2. There should be no entry for the same key in the commit log
+
+    // Operation
+    /// Put a local key
+
+    // Assertions
+    /// 1. Keystore should have the local key with the value inserted
+    /// 2. Assert the metadata of the key. "CreatedAt" should be populated with
+    /// DateTime which is less than DateTime.now()
```
VJag, line 605: Assert that the local key should not be present in the commit log. Pls, add it to the docs and code.

purnimavenkatasubbu: Added the assertion!

packages/at_client/test/sync_new_test.dart
```diff
@@ -0,0 +1,1944 @@
...
+      await keystore!.put(atKey, atData);
+      //-----------Assertions---------------------------------:
+      // verifying the key in the key store
+      var keyStoreGetResult = await keystore.get(atKey);
+      expect(keyStoreGetResult!.data, 'sample');
+      // verifiyng the createdAt time is less than DateTime.now()
+      expect(keyStoreGetResult.metaData!.updatedAt!.isBefore(DateTime.now()),
+          true);
+      // verifying the version of the key is 0
+      expect(keyStoreGetResult.metaData!.version, 0);
```
VJag, line 628: We are asserting after teardown?

purnimavenkatasubbu: Sorry!. Fixed.

packages/at_client/test/sync_new_test.dart
```diff
@@ -0,0 +1,2147 @@
...
+    /// Assertions:
+    /// 1. Key store should have the public key with the value inserted
+    /// 2. Assert the metadata of the key. "CreatedAt" should be populated with
+    /// DateTime which is less than DateTime.now()
+    /// 3. The version of the key should be set to 0
+    /// 4. CommitLog should have an entry for the new public key with commitOp.Update
+    /// and commitId is null
+    test('Verify uncommitted queue on creation of a public key', () async {
+      //------------Setup---------------------------------
+      await TestResources.setupLocalStorage(atsign);
```
sitaram-kalluri: If `await TestResources.setupLocalStorage(atsign)` and `await TestResources.tearDownLocalStorage()` need to run before and after each test, please consider moving the lines to setUp and tearDown methods respectively; Instead of repeating the lines in each test.

srieteja:
Many of the failing tests seem to pass independently. Investigating why they seem to fail when run as a group.

packages/at_client/test/sync_new_test.dart
purnimavenkatasubbu: Done

gkc (approved):
There are quite a few test cases here with code which has been commented out, and they are important test cases. I see comments on some which say they will be uncommented later once some changes have been merged to trunk (although the reason is not entirely clear to me).

There are several commented tests which mention that the assertions are not possible because we can no longer create invalid keys in the keystore. Let's not just give up on those tests, let's try to find ways to test those conditions because at some stage in the future we could have a situation where the server rejects keys that a client is sending because the client code is not yet preventing creation of those keys

Other than that the PR looks good, so I'll merge the changes to trunk, but we will need to follow up on the points I've mentioned above

srieteja:
The tests with the commented-out code are waiting for those features/behaviour to be implemented on the client side and will fail with the current client behaviour. When the client code matches the expected behaviour we can enable those tests, as per my knowledge these features are being worked on currently.

Reg the invalid-key-based tests, our Keystore is straight-out rejecting those keys making it impossible to insert those keys. One approach worth trying would be manually opening the hive boxes, inserting the keys and then asserting the conditions; which would bypass any validations. Will try working this approach out, or any other approach that I can think of or find.

PR #808, opened 2022-11-21
chore: consume lower_case enforcement changes in at_commons

srieteja:
**- What I did**
- enforce lower case check and conversion in at_client_impl
- show a warning if UPPERCASE keys are received and convert them to lower case

**- How I did it**
- log a warning that the key will be converted to lowercase 
- adjust tests to match the new key behaviour

chore: consume lowercase enforcement related changes in at_client


srieteja:
Migrating some of the implementations to at_commons. Will investigate the cause of failure after the migration.

srieteja:
Migrating some of the implementations to at_commons. Will investigate the cause of failure after the migration.

packages/at_client/lib/src/service/notification_service_impl.dart
```diff
@@ -26,8 +26,8 @@ class NotificationServiceImpl
   final Map<NotificationConfig, StreamController> _streamListeners =
       HashMap(equals: _compareNotificationConfig, hashCode: _generateHashCode);
   final emptyRegex = '';
-  static const notificationIdKey = '_latestNotificationIdv2';
-  static const lastReceivedNotificationKey = 'lastReceivedNotification';
+  static const notificationIdKey = '_latestnotificationidv2';
```
murali-shris: is there a need to change the constant ? since we are lower casing the key anyway before storing and in processing

srieteja: It's for the tests, we are checking for this key in the keystore and since the key will now be in lowercase the tests are failing. Alternative we can let the key be the same as before and use key.toLowerCase in the expect method of tests

gkc: imo Better to use key.toLowerCase in the expects in the tests

srieteja: I'll revert the variable name and expect key.lowercase(). Will push that change as part of the next commit

srieteja: This change has been done as part of ba098634f6a36c8e5d0b93eef775d50a3d111970

packages/at_client/test/notification_service_test.dart
```diff
@@ -994,9 +1022,9 @@ class StatsAtKeyMatcher extends Matcher {
 
   @override
   bool matches(item, Map matchState) {
-    if (item is AtKey && item.key!.contains('_latestNotificationIdv2')) {
+    if (item is AtKey && item.key!.contains('_latestNotificationIdV2')) {
```
murali-shris: why do we need upper casing for V ?

srieteja: this might have been changed while I was reverting the variable names to camelCase. The matcher class is not used anymore, so I  have removed the class in further commits. I believe that should solve this.

packages/at_client/lib/src/client/at_client_impl.dart
```diff
@@ -472,6 +472,10 @@ class AtClientImpl implements AtClient, AtSignChangeListener {
     if (atKey.metadata!.namespaceAware) {
       atKey.namespace ??= preference?.namespace;
     }
+    if (atKey.key!.contains(r'^[A-Z]') || atKey.key!.contains(r'^[A-Z]') ) {
```
gkc: Does this if condition work for a string like 'abcDefg'?

murali-shris (dismissed):

packages/at_client/lib/src/client/at_client_impl.dart
```diff
@@ -472,6 +472,10 @@ class AtClientImpl implements AtClient, AtSignChangeListener {
     if (atKey.metadata!.namespaceAware) {
       atKey.namespace ??= preference?.namespace;
     }
+    if (atKey.key!.contains(r'^[A-Z]') || atKey.namespace!.contains(r'^[A-Z]') ) {
+      _logger.info(
+          'AtKey: ${atKey.key}.${atKey.namespace} contains UPPER_CASE characters, converting the key to lower_case');
```
gkc: please use "upper case" and "lower case"

packages/at_client/test/notification_service_test.dart
```diff
@@ -162,6 +162,7 @@ void main() {
               AtClientPreference()..namespace = 'wavi',
               mockSharedKeyEncryptionImpl)
           .transform(notificationParams);
+      //UPPER_CASE in the forText() method will be preserved
```
gkc: Clarify the comment - maybe, "Upper case should be preserved when using forText notifications"

gkc (changes requested):
LGTM. Some small things to change in code comments

packages/at_client/lib/src/client/at_client_impl.dart
gkc: @srieteja Please address this question

srieteja: I just tested it and it's not working for a string like 'abcDefg'. Updating the regex to include strings like these.

srieteja: Update: the regex works fine but the way it's used might not be entirely correct. Updating the way regex is matched.

packages/at_client/lib/src/client/at_client_impl.dart
```diff
@@ -472,9 +472,13 @@ class AtClientImpl implements AtClient, AtSignChangeListener {
     if (atKey.metadata!.namespaceAware) {
       atKey.namespace ??= preference?.namespace;
     }
-    if (atKey.key!.contains(r'^[A-Z]') || atKey.namespace!.contains(r'^[A-Z]') ) {
+
```
gkc, line 476: This works :-)

Suggest making upperCaseRegex a static const

gkc (changes requested):

packages/at_client/lib/src/client/at_client_impl.dart
srieteja: sure. let me make that change too

srieteja: since it's not possible to use a static var in an instance method, I've made upperCaseRegex a class-level static variable in the latest commit.

packages/at_client/lib/src/client/at_client_impl.dart
```diff
@@ -61,6 +61,7 @@ class AtClientImpl implements AtClient, AtSignChangeListener {
   RemoteSecondary? _remoteSecondary;
   AtClientCommitLogCompaction? _atClientCommitLogCompaction;
   AtClientConfig? _atClientConfig;
+  static var upperCaseRegex = RegExp(r'[A-Z]');
```
gkc: static const please

gkc (changes requested):

packages/at_client/lib/src/client/at_client_impl.dart
srieteja: dart's not allowing me to make it a static const, it needs the RegExp constructor to be a const too(which it is not). Does making in static final work instead of static const ?

ps: the actual error being displayed by dart: 'The constructor being called isn't a const constructor.'

gkc: oh yeah I forgot that RegExp doesn't have a const constructor, because it can throw an exception

so just just static final

srieteja: upperCaseRegex is now static final as part of the latest commit

gkc (approved):
LGTM

packages/at_client/CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 - build: consume changes in at_commons v3.0.35 that enforce lowercase on AtKey
 - build: upgrade dependency at_persistence_secondary_server to v3.0.46
 ## 3.0.54
+- feat: Introduce CommitOp(CommitOperation) to the KeyInfo to describe key update or delete upon sync
```
gkc: This should be in the 3.0.55 section of the changelog
3.0.54 has already been published https://pub.dev/packages/at_client/changelog

gkc (changes requested):

packages/at_client/CHANGELOG.md
srieteja: This has been resolved in the latest commit

gkc (approved):

PR #809, opened 2022-11-21
feat: initial commit for at_chops uptake

murali-shris:
**- What I did**
- uptake at_chops package in at_client
- upgraded dependencies in pubspec for at_client
**- How I did it**
- created and injected an instance of atChops in persistKeys method of at_client_service.dart
- added a feature flag useAtChops to at_client_preference.dart. Defaults to false. Dependent clients have to set this flag to true in order to use at_chops. This flag will be defaulted to true once all packages have completed uptake of at_chops
- Modified implementation of packages/at_client/lib/src/decryption_service/shared_key_decryption.dart and packages/at_client/lib/src/encryption_service/abstract_atkey_encryption.dart to use current implementation and new atChops implementation based on feature flag
- Deprecated decryptKey method in encryption_util.dart which uses private key
- Modified unit tests to check results from current and new implementation
**- How to verify it**
 - run the unit tests in decryption_service_test.dart


packages/at_client/analysis_options.yaml
```diff
@@ -7,9 +7,12 @@ include: package:lints/recommended.yaml
 # Uncomment to specify additional rules.
 linter:
   rules:
```
gkc, line 9: Since you're editing this file anyway, can you add `unawaited_futures: true` as well please?

packages/at_client/lib/src/client/at_client_impl.dart
```diff
@@ -934,4 +937,43 @@ class AtClientImpl implements AtClient {
...
+  @override
+
+  /// Inject an instance of [AtChops]
+  /// if [AtChops] is not injected then pkam and encryption keys should be saved to local secondary before invoking this method.
+  void setAtChops({AtChops? atChops}) async {
+    if (atChops != null) {
+      _atChops = atChops;
+      return;
+    }
+    if (_atChops != null) {
```
gkc: Why are you not allowing it to be changed after it's been initially set?

packages/at_client/lib/src/client/at_client_impl.dart
```diff
@@ -68,7 +70,8 @@ class AtClientImpl implements AtClient {
       {AtClientManager? atClientManager,
       RemoteSecondary? remoteSecondary,
       EncryptionService? encryptionService,
-      SecondaryKeyStore? localSecondaryKeyStore}) async {
+      SecondaryKeyStore? localSecondaryKeyStore,
+      AtChops? atChops}) async {
```
gkc: You need to add atChops parameter to the private constructor, and pass this atChops parameter to it.

gkc (changes requested):
Looking good! I've made a few comments; FYI I've also [requested a change on the at_chops PR](https://github.com/atsign-foundation/at_libraries/pull/253#discussion_r1032798307) which will have impact here

packages/at_client/analysis_options.yaml
murali-shris: done

packages/at_client/lib/src/client/at_client_impl.dart
murali-shris: i have removed the check.

packages/at_client/lib/src/client/at_client_impl.dart
murali-shris: done

packages/at_client/lib/src/client/at_client_impl.dart
```diff
@@ -104,7 +109,8 @@ class AtClientImpl implements AtClient {
       AtClientPreference preference, AtClientManager atClientManager,
       {RemoteSecondary? remoteSecondary,
       EncryptionService? encryptionService,
-      SecondaryKeyStore? localSecondaryKeyStore}) {
+      SecondaryKeyStore? localSecondaryKeyStore,
+      AtChops? atChops}) {
```
gkc: You're not setting the _atChops instance variable if it's been passed?

packages/at_client/lib/src/client/at_client_impl.dart
```diff
@@ -967,4 +978,40 @@ class AtClientImpl implements AtClient {
     _preference = preference;
     _namespace = namespace;
   }
+
+  @override
+
+  /// Inject an instance of [AtChops]
+  /// if [AtChops] is not injected then pkam and encryption keys should be saved to local secondary before invoking this method.
+  void setAtChops({AtChops? atChops}) async {
```
gkc: Looks like this method is only called from one place in the codebase in at_client_service.dart, and when it is called from that place, it is called with an AtChopsImpl

So a few questions
(1) When do you envisage the fallback code here being executed? - i.e. when would we be calling setAtChops(null) ?
(2) Can we move the async code in this method somewhere else? i.e. can this be a standard setter?

gkc (changes requested):

packages/at_client/lib/src/client/at_client_impl.dart
murali-shris: (1 )
atChops object is mandatory for an at client instance because without at_chops encryption/decryption/data signing cannot be performed.
i put the fallback code in case the caller doesn't pass an instance of at_chops. There may be other packages apart from at_client_mobile that can create an instance of at client and call get/put etc.,
(2)
Will think about a solution for this.

gkc: I was thinking that at the time that _init is called, then the atClient could create its atChops instance if none was supplied.

murali-shris: I had the same idea and tried the change earlier today. When we are in the init method, there are two scenarios
1) local secondary is not available when isLocalStoreRequired=false
2) local secondary is available but encryption/pkam keys are not yet persisted.
For both 1 and 2, we will not be able to create an instance of at chops. in the latest commit, I changed the setter to dart convention and moved the async behavior to a new private method(currently not called)
I will uptake at chops in a package that uses at client without the onboarding flow and explore the options to create at chops instance.

gkc:
@murali-shris When do you think it will be ready to merge this to trunk?

murali-shris:
> @murali-shris When do you think it will be ready to merge this to trunk?

Final piece that has to solved is from where to call AtClientImpl._createAtChopsInstance() when atChops instance is not set for an at_client instance. Packages using at_client without at_client_mobile will break because _createAtChopsInstance() needs the encryption and pkam keys set in localSecondary.

Alternate approach is have useAtChops=false by default in at_client_preference and tackle one package at a time by setting useAtChops=true when we uptake at_chops for that  particular package.

murali-shris:
removed at_client_mobile related changes in this PR and moved to a new PR
https://github.com/atsign-foundation/at_client_sdk/pull/863

murali-shris:
TODOs before merging this PR
* merge atlookup PR - https://github.com/atsign-foundation/at_libraries/pull/276
* publish atlookup
* change pubspec in atclient, end2end and functional test

murali-shris:
> TODOs before merging this PR
> 
> * merge atlookup PR - [fix: fallback code for atchops at_libraries#276](https://github.com/atsign-foundation/at_libraries/pull/276)
> * publish atlookup
> * change pubspec in atclient, end2end and functional test

done

packages/at_client/lib/src/client/at_client_impl.dart
```diff
@@ -954,4 +973,27 @@ class AtClientImpl implements AtClient {
     _preference = preference;
     _namespace = namespace;
   }
-}
+
+  Future<AtChops> _createAtChopsInstance() async {
```
gkc: Don't think we need this method any more?

gkc (dismissed):

packages/at_client/lib/src/client/at_client_impl.dart
murali-shris: yeah.i will remove this.

gkc (approved):

sitaram-kalluri (approved):

PR #814, opened 2022-11-24
fix: sync running into loop on invalid keys

sitaram-kalluri:
**- What I did**
* Fix the sync infinite loop issue.

**- How I did it**
* At the moment, When the server is ahead of the local secondary, to ensure all the delta commits are received, we run a loop until the localCommitId is equal to serverCommitId.
* The problem with this approach is if an invalid key is not synced to local  secondary, the sync runs into an infinite loop.
* Hence we replaced the `while (serverCommitId > localCommitId)` with `while (serverCommitId > lastReceivedServerCommitId)` so that we run the loop until the lastReceivedServerCommitId is equal to serverCommitId

**- How to verify it**
* The end-to-end tests and the unit tests should pass.

* Fix the sync running into an infinite loop when invalid keys do not sync to the local secondary

### DIFF
--- a/packages/at_client/lib/src/service/sync_service_impl.dart
+++ b/packages/at_client/lib/src/service/sync_service_impl.dart
@@ -421,12 +421,11 @@ class SyncServiceImpl implements SyncService, AtSignChangeListener {
           await _syncLocal(serverCommitEntry);
           keyInfoList.add(keyInfo);
         } on Exception catch (e, stacktrace) {
-          var cause = (e is AtException) ? e.getTraceMessage() : e.toString();
           _logger.severe(
-              'exception syncing entry to local $serverCommitEntry - $cause - stacktrace: $stacktrace');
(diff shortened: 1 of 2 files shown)


packages/at_client/lib/src/service/sync_service_impl.dart
```diff
@@ -422,7 +431,8 @@ class SyncServiceImpl implements SyncService, AtSignChangeListener {
       }
```
gkc, line 430: LGTM. While we're changing the code here, can you also log the stack trace when we encounter exceptions or errors please

gkc (changes requested):

packages/at_client/lib/src/service/sync_service_impl.dart
sitaram-kalluri: Added stack trace to the exception and the error.

gkc (approved):

PR #823, opened 2022-11-28
fix: invalid pad block issue

sitaram-kalluri:
**- What I did**
1. Refactor the `getSharedKey` method in `AbstractAtKeyEncryption` to sync the `encryptedSharedKey` to the cloud secondary if not already synced before sending the notifications.
2. Write unit tests to assert the above functionality
3. Refactor the classes to initialize fields with mock objects when writing unit tests.

**- How I did it**
1. Introduce `isEncryptedSharedKeyInSync` which verifies if the `shared_key` is synced to the cloud secondary. If not synced, runs the update verb to store the shared_key in the cloud secondary.

2. Refactor the `_getSharedWithPublicKey` method which fetches the encryption public key of the sharedWith atSign. 
    - Earlier (prior to caching of public keys) we store to sharedWith atSign's encryption public key as `public.bob@alice` to store the encryption public key of @‎bob in @‎alice atSign. 
    - Now with caching of the public key, upon initial look up the encryption public key, a cached key will be created automatically (cached:public:publickey@‎bob).      
    -  So, instead of creating a new key (public.bob@alice), we can utilize the cached key(cached:public:publickey@bob). Therefore, removed the piece of code that creates the key (public.bob@alice) and refactor the method that fetches the encryption public key to first look for a cached public key, if not then run a lookup verb to sharedWith atSign.

3. Following are the changes to enable writing unit tests:

   1. Refactor AtKeyEncryptionManager and AtKeyDecryptionManager classes to accept the at_client object 
   2. Add a constructor to LocalKeyDecryption, SharedKeyEncryption and StreamEncryption classes that accept AtClient instance to enable passing mock objects in unit testing.
   3. Remove the static nature of the getSharedKey method and write unit tests around getting the shared key from the local secondary and remote secondary.
   4. Since, LocalKeyDecryption also needs a shared_key to decrypt the value, mark LocalKeyDecryption as a child of AbstractAtKeyEncryption to make the getSharedKey method available for LocalKeyDecryption.
   5. Refactor the existing unit tests to accommodate the above changes.

**- How to verify it**

Following are the unit tests written to assert the above changes:

Tests in encryption_service_test.dart

1. A group of tests related to fetching and decrypting the encrypted shared key
- A test to verify encrypted shared key is fetched from the local secondary and decrypted successfully
- A test to verify encrypted shared key is fetched from remote secondary and decrypted successfully
- A test to verify empty string is returned when the shared key is not found in local and remote secondary

2. A group of tests related to shared_key is synced to the cloud secondary
- A test to verify isEncryptedSharedKeyInSync method returns false when commit id null
- A test to verify isEncryptedSharedKeyInSync method returns true when commit is not null
- A test to verify isEncryptedSharedKeyInSync method returns false when two commit entries exist with commit id null and not null

3. A group of tests related to shared key encryption
- test to verify encryption when the shared key is available
- test to verify encryption when a new shared key is generated
- test to verify exception is thrown when update command fails to store shared_key to remote secondary

Test in notification_service_test.dart
1. A test to verify notification is not sent when the encryption service throws an exception

Fix initial notifications throwing pad block issue because "encryptedSharedKey" is not synced to cloud secondary.

ToDo's

- Initial plan was to fetch the sharedWith encryption public key only if a new shared_key is generated. If the shared_key is already available, then fetch the encrypted shared key from the @‎bob:shared_key@alice key. 

Below are items for discussion:

1. The checksum of encryption public key of sharedWith atSign is added to AtKey metadata. At the moment, the checksum data is not stored; needs to compute every time. For this, we need an encryption public key of the shared With atSign. So irrespective of IF the shared_key is existing or a new one, we might need to fetch the encryption public key of sharedWith atSign.

2. If the shared_key is generated, we need an additional IF block that will search for the shared_key (@‎bob:shared_key@‎alice) availability in the local and remote. So, since we anyways have to fetch the encryption public key for # 1, Can we directly encrypt the shared key each time or should it be fetched from the key (@‎bob:shared_key@‎alice)?

### DIFF
--- a/packages/at_client/lib/src/decryption_service/decryption_manager.dart
+++ b/packages/at_client/lib/src/decryption_service/decryption_manager.dart
@@ -1,13 +1,25 @@
+import 'package:at_client/at_client.dart';
 import 'package:at_client/src/decryption_service/decryption.dart';
 import 'package:at_client/src/decryption_service/local_key_decryption.dart';
 import 'package:at_client/src/decryption_service/self_key_decryption.dart';
 import 'package:at_client/src/decryption_service/shared_key_decryption.dart';
-import 'package:at_commons/at_commons.dart';
 
 ///The manager class for [AtKeyDecryption]
 class AtKeyDecryptionManager {
+  late AtClient _atClient;
+
+  static final AtKeyDecryptionManager _singleton =
+      AtKeyDecryptionManager._internal();
+
+  AtKeyDecryptionManager._internal();
+
+  factory AtKeyDecryptionManager.getInstance(AtClient atClient) {
+    _singleton._atClient = atClient;
+    return _singleton;
+  }
+
   /// Return the relevant instance of [AtKeyDecryption] for the given [AtKey]
-  static AtKeyDecryption get(AtKey atKey, String currentAtSign) {
+  AtKeyDecryption get(AtKey atKey, String currentAtSign) {
     // If sharedBy not equals currentAtSign, return SharedKeyDecryption
     // Eg: currentAtSign is @bob and key is phone@alice
     if (atKey.sharedBy != currentAtSign) {
@@ -23,6 +35,6 @@ class AtKeyDecryptionManager {
     // Returns LocalKeyDecryption to for the keys present in local storage
     // that are sharedWith other atSign's.
     // @alice:phone@bob
-    return LocalKeyDecryption();
+    return LocalKeyDecryption(_atClient);
   }
 }
--- a/packages/at_client/lib/src/decryption_service/decryption_manager.dart
+++ b/packages/at_client/lib/src/decryption_service/decryption_manager.dart
@@ -6,17 +6,9 @@ import 'package:at_client/src/decryption_service/shared_key_decryption.dart';
 
 ///The manager class for [AtKeyDecryption]
 class AtKeyDecryptionManager {
-  late AtClient _atClient;
+  late final AtClient _atClient;
 
-  static final AtKeyDecryptionManager _singleton =
-      AtKeyDecryptionManager._internal();
-
-  AtKeyDecryptionManager._internal();
-
-  factory AtKeyDecryptionManager.getInstance(AtClient atClient) {
-    _singleton._atClient = atClient;
-    return _singleton;
-  }
+  AtKeyDecryptionManager(this._atClient);
 
   /// Return the relevant instance of [AtKeyDecryption] for the given [AtKey]
   AtKeyDecryption get(AtKey atKey, String currentAtSign) {
--- a/packages/at_client/lib/src/encryption_service/abstract_atkey_encryption.dart
+++ b/packages/at_client/lib/src/encryption_service/abstract_atkey_encryption.dart
@@ -1,12 +1,16 @@
+import 'dart:collection';
+
 import 'package:at_client/at_client.dart';
 import 'package:at_client/src/client/secondary.dart';
 import 'package:at_client/src/encryption_service/encryption.dart';
 import 'package:at_client/src/encryption_service/shared_key_encryption.dart';
 import 'package:at_client/src/encryption_service/stream_encryption.dart';
 import 'package:at_client/src/response/default_response_parser.dart';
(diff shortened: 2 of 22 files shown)


packages/at_client/lib/src/decryption_service/decryption_manager.dart
```diff
@@ -1,13 +1,25 @@
...
 import 'package:at_client/src/decryption_service/local_key_decryption.dart';
 import 'package:at_client/src/decryption_service/self_key_decryption.dart';
 import 'package:at_client/src/decryption_service/shared_key_decryption.dart';
-import 'package:at_commons/at_commons.dart';
 
 ///The manager class for [AtKeyDecryption]
 class AtKeyDecryptionManager {
+  late AtClient _atClient;
+
+  static final AtKeyDecryptionManager _singleton =
```
gkc: Please find another way other than introducing a singleton, especially since calling the getInstance() can modify the singleton's state.

gkc (changes requested):

packages/at_client/lib/src/decryption_service/decryption_manager.dart
gkc: PS This refers to both AtKeyEncryptionManager and AtKeyDecryptionManager

sitaram-kalluri: Removed the singleton nature for AtKeyEncryptionManager and AtKeyDecryptionManager classes.

packages/at_client/lib/src/encryption_service/abstract_atkey_encryption.dart
```diff
@@ -61,177 +62,167 @@ abstract class AbstractAtKeyEncryption implements AtKeyEncryption {
...
+      ..atKey =
+          '$AT_ENCRYPTION_SHARED_KEY.${atKey.sharedWith?.replaceAll('@', '')}'
+      ..sharedBy = atKey.sharedBy;
+    return await secondary.executeVerb(llookupVerbBuilder);
+  }
+
+  /// Checks if the encryptedSharedKey is synced to the cloud secondary
+  ///
+  /// If Synced, returns true; else returns false.
+  Future<bool> isEncryptedSharedKeyInSync(AtKey atKey) async {
```
gkc, line 208: Since getLatestCommitEntry is an expensive operation (for now) - can you maintain a map of atKeys which have been previously checked and found to be in sync? So then, if the map contains the atKey, you can return true immediately.

packages/at_client/pubspec.yaml
```diff
@@ -37,6 +37,12 @@ dependencies:
   at_utils: ^3.0.11
   meta: ^1.8.0
 
+dependency_overrides:
+  at_persistence_secondary_server:
+    git:
+      url: https://github.com/atsign-foundation/at_server
+      path: packages/at_persistence_secondary_server
+      ref: refactor_commit_log
```
gkc: Please can you

either (1) implement getLatestCommitEntry on the current code structure - i.e. while holding off on the refactoring

Or (2) ensure the refactoring is refactoring only - i.e. does not introduce any breaking change

gkc (changes requested):

packages/at_client/lib/src/encryption_service/abstract_atkey_encryption.dart
sitaram-kalluri: Sure Gary.

packages/at_client/pubspec.yaml
sitaram-kalluri: @gkc: I have moved the logic to fetch the latest commit entry to `SyncUtil` class in AtClient package. So now, we do not have to touch the at_persistence_secondary_server

gkc: Perfect - thanks Sitaram

packages/at_client/lib/src/encryption_service/abstract_atkey_encryption.dart
gkc: Perfect - thanks Sitaram

gkc (approved):

PR #824, opened 2022-11-28
fix: Client can 'miss' initial notifications

### LINKED ISSUE #770: Clients will miss pending notifications upon startup if there is a delay between NotificationService creation and client calling NotificationService.subscribe
### Describe the bug

Clients will miss pending notifications upon startup if there is a delay between Monitor creation and client calling NotificationService.subscribe

### Steps to reproduce

Introduce a delay (for example by waiting for sync to complete) before calling subscribe on NotificationService. 

### Expected behavior

Clients should not miss notifications

### Screenshots

_No response_

### Smartphones

_No response_

_No response_

### Additional context

Reported by a customer, encountered by them when using some sample code I had supplied which waited for an initial sync to complete before calling NotificationService.subscribe()

1. I have implemented an initial patch which minimizes the risk of this happening and customer has verified. Will raise PR for that
2. Have filed an [enhancement request](https://github.com/atsign-foundation/at_client_sdk/issues/771) to fix this properly.

gkc:
**- What I did**
- Fixes #770 : "Clients will miss pending notifications upon startup if there is a delay between NotificationService creation and client calling NotificationService.subscribe"
- This is a temporary fix to tide us over until the proper fix: implementation of enhancements to the monitor verb which allow for multiple subscriptions, and changes to this service and to the Monitor to make use of that enhancement
- Fixed state management bug re isMonitorPaused
- Fixed state management bug re connectivity listener subscription

**- How I did it**
NotificationServiceImpl
- Previously we were initializing the notification service before there were any 'real' subscriptions and, because the notification service currently starts the monitor without any regex, the monitor immediately starts to stream all notifications
- As a result, if there was even a very short delay between when the notification service was created and when the app called 'subscribe', then the app would 'miss' any notifications already delivered from server.
- So - for now, if the subscription is 'statsNotification' (which is a subscription by the SyncService), then we will delay initialization of the notification service (and thus delay creation of the Monitor and its issuing of the monitor command) until when the app does a 'real' `subscribe` call. We will delay initialization for 30 seconds or until the app code calls `subscribe`, whichever is sooner.
- the `monitorIsPaused` instance variable was previously initially set to `false`. This was OK when we were immediately calling _init upon creation and therefore immediately starting the monitor. But now that we are not immediately calling init, monitorIsPaused is initially set to 'true'
* _startMonitor has always set monitorIsPaused to false, but only if the monitor status was 'started' after calling Monitor.start(). This was incorrect, since the intention is clearly that the monitor *should* be running after startMonitor is called, even if Monitor.start() failed
* Prior to this fix, if Monitor.start() failed then the retry mechanisms would still work because monitorIsPaused hadn't been set to 'true' by anything - but now, we are initializing monitorIsPaused to true
  * (Note that the retry mechanism would not have worked if we had had a _startMonitor followed by stopAllSubscriptions followed by a _startMonitor where the Monitor failed to start)
* So now, we are setting monitorIsPaused to false when _startMonitor is called, even if the monitor fails to start, so we are correctly reflecting the intended state
* With this fix, the monitorIsPaused variable will now always correctly reflect the intended state. It will be set to true at construction of the NotificationServiceImpl, will be set to false when _startMonitor is called, and will be set to true when stopAllSubscriptions is called. As a result, the retry mechanism will work as intended in all situations.
* stopAllSubscriptions now sets _connectivityListener to null in addition to cancelling connectivity listener subscription. As a result, a subsequent subscribe() will now lead to a new connectivity listener subscription being created. (Previously, a stopAllSubscriptions followed later by a new subscribe - would have meant we would have ended up without a connectivity listener subscription, since it is set up during _init() only if _connectivityListener is null.

SyncServiceImpl
- After the above changes to NotificationServiceImpl, if the app isn't explicitly calling 'sync', and hasn't called subscribe(), then there will now be a delay of 30 seconds before the monitor is started (and therefore a statsNotification message can be received, which will then queue a sync request, which will then eventually cause a sync to run)
- In order to compensate for that, the SyncServiceImpl itself now queues a sync request when it is initialized. This ensures that a sync will take place even if (1) the app hasn't explicitly called sync() and (2) app has not called subscribe so we've not yet received a statsNotification. This preserves behaviour similar to today where rapid connection of the Monitor followed by receipt of a statsNotification from the server would lead to a sync request being queued very soon after creation of the SyncService.

**- How to verify it**
- Tests all pass
- **Note** that the fixed behaviour (defer calling _init() until a 'real' subscription is made) has already been tested and verified by a customer. This PR contains that fix but also contains several more changes which address the side effects of the functionality / behaviour fix

- fix: Changes to ensure that notification stream from server is not set up until client code sets up its subscriptions.


gkc:
@VJag Hi Jagan - can you review this PR, please? As noted in the PR description above, the core fix has already been tested and verified by ralfeus in another branch of mine, and I've been as careful as I can be re side effects, but the more scrutiny the better.

gkc:
Created [a ticket](https://github.com/atsign-foundation/at_client_sdk/issues/827) for adding more unit tests to prevent regression of the other bugs uncovered during and fixed by this PR

packages/at_client/lib/src/service/notification_service_impl.dart
```diff
@@ -340,8 +374,8 @@ class NotificationServiceImpl
         return notificationResult;
       } else {
         // no wait? no await
-        _waitForAndHandleFinalNotificationSendStatus(
-            notificationParams, notificationResult, onSuccess, onError);
+        unawaited(_waitForAndHandleFinalNotificationSendStatus(
```
VJag, line 377: Found this explanation of 'unawaited' simple and effective:

https://github.com/vandadnp/flutter-tips-and-tricks/blob/main/tipsandtricks/unawaited-and-ignore-in-dart/unawaited-and-ignore-in-dart.md

VJag (approved):

PR #838, opened 2022-12-02
fix: Dependabot PRs skip LCOV steps in end2end

cpswan:
fixes #837 

**- What I did**

Added conditionals to skip steps that won't complete correctly due to missing earlier steps

**- How I did it**

Copy/paste/rework from existing conditionals

**- How to verify it**

Will need another Dependabot PR. Should be possible with #836 after a trunk merge following this being merged.

**- Description for the changelog**

fix: Dependabot PRs skip LCOV steps in end2end

PR #839, opened 2022-12-02
fix: Skip end2end_test_14 for Dependabot PRs

cpswan:
#838 means that end2end tests appear to succeed for Dependabot PRs, as they're `required`. But then end2end_test_14 runs as if the previous job completed correctly, when all the stuff that matters has been skipped. So the test might succeed (or fail) for reasons unrelated to the change at hand.

**- What I did**

New conditional to explicitly skip test block for Dependabot PRs

**- How I did it**

Based on similar conditionals in at_server workflow

**- How to verify it**

Next Dependabot PR should skip

fix: Skip end2end_test_14 for Dependabot PRs

### DIFF
--- a/.github/workflows/at_client_sdk.yaml
+++ b/.github/workflows/at_client_sdk.yaml
@@ -199,6 +199,7 @@ jobs:
   # The Job runs end-to-end tests between server code running on trunk branch and canary (production ready) server
   end2end_test_14:
     needs: [   end2end_tests ]
+    if: ${{ github.actor != 'dependabot[bot]' }}
     runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


gkc:
One of the Functional tests is flaking and occasionally failing. Not related to this PR. I'll raise a ticket for the flaking test and fix it later

gkc (approved):

gkc:
> One of the Functional tests is flaking and occasionally failing. Not related to this PR. I'll raise a ticket for the flaking test and fix it later

Ticket https://github.com/atsign-foundation/at_client_sdk/issues/840

PR #846, opened 2022-12-09
fix: namespace in atkey overrides the namespace in preference

### LINKED ISSUE #841: Atkey namespace overridden by AtClientPreference namespace
### Describe the bug

This happened when we tried deleting wavi image key from Buzz app.

In Buzz, we created following atKey with `wavi` as namespace
```
 atKey = AtKey()
      ..key = "image"
      ..sharedWith = null
      ..namespace = 'wavi'
      ..metadata = metaData;
```

when we called delete operation on it, the namespace was overridden by the one added in `AtClientPreference`.

Thus the actual key was not deleted.

### Steps to reproduce

1. Run Buzz app and try to delete image with above-mentioned code.

### Expected behavior

- the namespace added in atKey should not have been overridden.

### Screenshots

_No response_

### Smartphones

- iPhone 12 simulator

Buzz

### Additional context

_No response_

sitaram-kalluri:
**- What I did**
1. When atKey has a different namespace from the namespace in the AtClientPreference, set the namespace from the atKey to the key.
2. Update the CHANGELOG.md
3. Update the version in pubspec.yaml and AtClientConfig

**- How I did it**
1. In the atClient-> _delete method, use `AtClientUtil.getKeyWithNameSpace` instead of `_getKeyWithNamespace` in AtClientImpl

**- How to verify it**
1. Added unit test, functional test and an end2end test to verify the namespace from atKey overrides the namespace in the preference.

- fix: Enable AtKey.namespace overrides the namespace in AtClientPreference in AtClient delete method


gkc:
@kalluriramkumar Can you explain what the root cause of the issue was please? Either here, or in the bug report ticket

packages/at_client/lib/src/client/at_client_impl.dart
```diff
@@ -214,44 +214,31 @@ class AtClientImpl implements AtClient {
...
-      bool namespaceAware = true,
-      bool isLocal = false}) async {
+  Future<bool> _delete(AtKey atKey) async {
+    // If metadata is null, initialize metadata
+    atKey.metadata ??= Metadata();
     String keyWithNamespace;
-    if (namespaceAware) {
-      keyWithNamespace = _getKeyWithNamespace(key);
+    if (atKey.metadata!.namespaceAware) {
+      keyWithNamespace = AtClientUtil.getKeyWithNameSpace(atKey, _preference!);
```
gkc, line 229: I notice that _getKeyWithNamespace is still called from the `notifyChange` and `putMeta` methods ... do those parts of the code also need to be looked at?

gkc (changes requested):

packages/at_client/lib/src/client/at_client_impl.dart
sitaram-kalluri: `notifyChange` is currently not in use and it has no usage in at_client. The code is moved to `notification_service_impl.notify` method. Perhaps, can we mark `notifyChange` as deprecated? and We have update in `putMeta` method.

sitaram-kalluri:
> @kalluriramkumar Can you explain what the root cause of the issue was please? Either here, or in the bug report ticket

@gkc: At the moment, the delete method fetches the namespace from the `AtClientPreference.namespace` only. we cannot override the namespace of a specific key.

One use case of setting a different namespace for a specific key is, we fetch the public info (firstName, lastName and profile pic) from the wavi and display it in the buzz app. Now, If someone wants to update/delete the profile pic from buzz app, we have to form a key with the namespace `wavi` in the buzz app.

packages/at_client/lib/src/client/at_client_impl.dart
gkc: Yes notifyChange should be marked deprecated

gkc: I don't understand what you mean by "We have update in putMeta method"?

sitaram-kalluri: @gkc: Sorry for the typo. I meant it has to be updated in the putMeta method also. Shall I make the changes in the same PR or can I create a separate ticket and do the changes.

gkc: OK I see. Up to you re separate ticket

gkc (approved):

packages/at_client/lib/src/client/at_client_impl.dart
sitaram-kalluri: @gkc: Since this is tested and apps team need the changes, merging these changes. Will make the changes for putMeta in separate PR.

PR #855, opened 2022-12-19
feat: client commit log compaction

sitaram-kalluri:
**- What I did**
* Introduce client-side commit log compaction

**- How I did it**
* Introduced `AtClientCommitLogCompaction` in at_client package that calls the compaction job in the at_persistence_secondary_server package.
* Introduced a new variable "commitLogCompactionTimeIntervalInMins" in AtClientConfig that sets the time interval of Compaction Job.
* Changed AtClientImpl as a class implementing "AtSignChangeListener". On switch atSign, the compaction cron job of the previous atSign is stopped and the cron job on the new atSign is started.
* In "listenToAtSignChange" method, added the IF condition to check and remove only the previous atSign AtClientImpl instance.
   - Reintroduced the IF condition because the AtClientImpl instance is created before the notifyListeners method.
   - notifyListeners method has switchAtSignEvent as an argument and SwitchAtSignEvent needs CurrentAtClient Instance. So did not move notifyListeners method before the creation of AtClientImpl.
* On switch atSign event, removing the atSign from the static map instance - "atClientInstanceMap" for the following reasons
   * On switch event, to avoid returning the old AtClientImpl instance which is removed from AtClientManager._changeListeners.
      - If the same instance is returned, since it is not in AtClientManager._changeListener upon subsequent switch atSign, "this" AtClientImpl instance will not be notified.
* Added getter on "AtClientCommitLogCompaction" to expose "isCompactionJobRunning" method to assert if the compaction job is running in e2e tests.
* In at_client_service.dart, removed the usage of SyncService from AtClientManger since it is deprecated and replaced it with SyncService from AtClientImpl.
      
**- How to verify it**

* Updated the unit tests in at_client_impl_test.dart to assert the previous atSign compaction job listener is removed on switch atSign events
* Update the unit tests in at_persistence_secondary_server with tests that mimic client-side storage.
  - In the PR: https://github.com/atsign-foundation/at_server/pull/950 , packages/at_persistence_secondary_server/test/at_compaction_test.dart contains the unit tests.
* Added a functional test to verify the commit log compaction on a single atSign. Call the "put" request on the same key to create multiple commit entries and assert that entries in the commit log shrink after the compaction.
* Added end2end test to asset the commit log compaction on a switch atSign scenario and a test to verify that commit log shrinks on compaction 

* feat: Introduce client-side commit log compaction.

Additional Notes:

* Set the dependency overrides to at_persistence_secondary_server to expose AtCompactionService to skip the cron job and run compaction job for testing purpose. Following is the PR : https://github.com/atsign-foundation/at_server/pull/1156
* @gkc: Since "AtCompactionService.executeCompaction" has an input argument of "AtLogType", exported the "AtCompactionService" class in at_persistence_secondary_server. The tests would directly invoke "AtCompactionService.executeCompaction" by passing the client-side commit log to compact.
* Skipped the changes in "AtCompactionJob.scheduleCompactionJob" to skip the cron job since the above change solves the issue.
* Introduced "isScheduled" method in "AtCompactionJob" class which indicates if the compaction job is running or not.


sitaram-kalluri:
Following are the test results of the review comments from @VJag:

a. Test commit-log compaction with switch atSign -  Tested with two atSigns switching between each other. When an atSign is switched, the new atSign's commit log instance is passed to the compaction job. The cron  job of previous atSign is paused and the cron job of a new atSign start/resume.

b. Test with huge size commit log and verify the put operation: Tested with at_client samples and below is the log snippet of the test result.
```
2023-01-03 11:04:03.043778 | Before put operation
FINER|2023-01-03 11:04:03.071493|AtCompactionStats|AtCommitLog: {"atCompactionType":"AtCommitLog","lastCompactionRun":"2023-01-03 05:34:03.071371Z","compactionDurationInMills":"69","preCompactionEntriesCount":"4602","postCompactionEntriesCount":"256","deletedKeysCount":"4346"} 
FINER|2023-01-03 11:04:03.072654|LocalSecondary|calling sync immediate from local secondary 
2023-01-03 11:04:03.083301 | After put operation
put result: true
```

Tested with 20,000+ keys and below is log snippet
```
FINER|2023-01-03 15:12:05.237814|AtCompactionStats|AtCommitLog: {"atCompactionType":"AtCommitLog","lastCompactionRun":"2023-01-03 09:42:05.236724Z","compactionDurationInMills":"229","preCompactionEntriesCount":"20250","postCompactionEntriesCount":"50","deletedKeysCount":"20200"}
```

gkc (changes requested):
* Need automated tests to verify behaviour when switching atSigns
* Can you explain the design rationale for creating a singleton which can act for multiple atSigns versus having an instance per atSign and controlling its lifecycle more explicitly when AtClientManager.setCurrentAtSign is called?

sitaram-kalluri:
> * Need automated tests to verify behaviour when switching atSigns
> * Can you explain the design rationale for creating a singleton which can act for multiple atSigns versus having an instance per atSign and controlling its lifecycle more explicitly when AtClientManager.setCurrentAtSign is called?

@gkc: The automated tests on switch atSign are in my local branch. I am waiting for https://github.com/atsign-foundation/at_client_sdk/pull/867 to be merged so that I can use the same approach in removing the previous atSign AtClientCommitLogCompaction instances. 

Initial idea is to reuse the AtClientCommitLogCompaction to invoke the scheduled job of compaction. But then I removed the singleton nature of AtClientCommitLogCompaction and the changes are in my local branch.

gkc (commented):
Generally LGTM but before we merge this I'd really like to see an overview of the algorithm, and why the commit log compaction is safe to run on client side even though client-side commit log usage is very different from server-side usage

sitaram-kalluri:
> Generally LGTM but before we merge this I'd really like to see an overview of the algorithm, and why the commit log compaction is safe to run on client side even though client-side commit log usage is very different from server-side usage

The difference between the client and cloud secondary commit log is, the client commit log will have entries with null commit IDs which are the uncommitted entries yet to sync to the cloud secondary.

```dart
commitLogMap.removeWhere((key, value) => value.commitId == null);
```
Above is a code snippet from CommitLogKeystore.getDuplicateEntries, which is invoked by the CompactionJob to get the duplicate entries. we ignore the entries with null commit IDs and so we remove them from commitLogMap to preserve the entries. At the moment, we did not target removing the multiple entries of the same key in the uncommitted entries list.

Then, the commit IDs are sorted in descending order and get the distinct commit entry of each key with the highest commit ID and remove the remaining entries.
```dart
var sortedKeys = commitLogMap.keys.toList(growable: false)
      ..sort((k1, k2) => _sortByCommitId(commitLogMap[k2], commitLogMap[k1]));
    var tempSet = <String>{};
    var expiredKeys = <int>[];
    for (var entry in sortedKeys) {
      _processEntry(entry, tempSet, expiredKeys, commitLogMap);
    }
    return expiredKeys;
```

packages/at_client/CHANGELOG.md
```diff
@@ -1,10 +1,12 @@
 ## 3.0.51
+- feat: Introduce commit log compaction to keep size of commit log thin
```
gkc, line 2: 3.0.51 has been published, this change will be in 3.0.52

packages/at_client/lib/src/preference/at_client_config.dart
```diff
@@ -11,4 +11,9 @@ class AtClientConfig {
   /// Represents the at_client version.
   /// Must always be the same as the actual version in pubspec.yaml
   final String atClientVersion = '3.0.51';
+
```
gkc: will need to change to 3.0.52 (here and in pubspec.yaml)

gkc (changes requested):
Couple of changes re version and CHANGELOG (next version will be 3.0.52)

You mentioned in a comment on this PR
> a. Test commit-log compaction with switch atSign - Tested with two atSigns switching between each other. When an atSign is switched, the new atSign's commit log instance is passed to the compaction job. The cron job of previous atSign is paused and the cron job of a new atSign start/resume.
> 
> b. Test with huge size commit log and verify the put operation: Tested with at_client samples and below is the log snippet of the test result.

Please add test (a) to the end-to-end test pack, and add a variation of test (b) to the functional test pack (the variation being, the number of commit log entries such that generation of them and running of the compaction job should all complete within 30 seconds at most.

gkc:
@sitaram-kalluri @purnimavenkatasubbu is this PR ready for re-review?

sitaram-kalluri:
> @sitaram-kalluri @purnimavenkatasubbu is this PR ready for re-review?

@gkc: Updated the functional tests to execute put operations and sync when compaction is running in asynchronously. Please review the changes.

gkc (approved):

PR #858, opened 2022-12-22
feat: Add new method to check sync

### LINKED ISSUE #832: isInSync should return an enum

Currently, the isInSync() returns either true or false.

It should return whether serverAhead or clientAhead or InSync,

_No response_

### Additional context

_No response_

sitaram-kalluri:
**- What I did**
* Replace `isInSync` with `checkIfClientAndServerInSync` in `SyncService` and deprecate `isInSync`. 
* The `checkIfClientAndServerInSync` returns `Future<SyncResult>` which has `syncStatus`, `localCommitId` and `serverCommitId`
* Add the following states to syncStatus
  * serverAhead
  * clientAhead
  * inSync
* Add `localCommitId` and `serverCommitId` to `SyncResult` class; where localCommitId and serverCommitId are set and returned along with SyncStatus

**- How I did it**
* Added a new method in `SyncService` class

**- How to verify it**
* Added the following unit tests:
  * The test to verify the method returns server-Ahead when the server commit-id is greater than the client commit-id
  * The test to verify the method returns client-Ahead when the client commit-id is greater than the server commit-id
  * The test to verify the method returns server-Ahead when the server commit-id is greater than the client commit-id and the client has uncommitted entries.
  * The test to verify the method returns inSync when the client and server are in sync and the client does not have uncommitted entries.

* Deprecate isInSync and add `checkClientAndServerInSync` to check if server and client are in sync.

### DIFF
--- a/packages/at_client/CHANGELOG.md
+++ b/packages/at_client/CHANGELOG.md
@@ -1,3 +1,5 @@
+## 3.0.50
+- fix: Add `checkIfClientAndServerInSync` in `SyncService` and deprecate `isInSync`
 ## 3.0.49
 - fix: Enable AtKey.namespace overrides the namespace in AtClientPreference in AtClient delete method
 - fix: Fixed a bug where initial notifications fails to decrypt - invalid pad block issue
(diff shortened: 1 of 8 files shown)


sitaram-kalluri:
There is a need for an uncommitted list cached in the at_client package to address this issue without impacting the performance. Hence closing this PR for now

PR #864, opened 2023-01-04
fix: Switch atsign bug

murali-shris:
**- What I did**
- fixed issue related to clearing of sync progress listener while switching atsign
**- How I did it**
- introduced a public method removeAllProgressListeners() in SyncService abstract class for clearing progress listeners
- added implementation for removeAllProgressListeners() in SyncServiceImpl
- changed syncProgressListener from Set to List as there is no equals implementation for SyncProgressListener
- removed static nature of syncProgressListener 
**- How to verify it**
- run the unit test switch_atsign_test.dart


packages/at_client/lib/src/service/sync_service_impl.dart
```diff
@@ -867,7 +868,9 @@ class SyncServiceImpl implements SyncService, AtSignChangeListener {
       _cron.close();
       _logger.finer(
           'removing from _syncServiceMap: ${_atClient.getCurrentAtSign()}');
-      _syncServiceMap.remove(_atClient.getCurrentAtSign());
+      final previousSyncService =
+          _syncServiceMap.remove(_atClient.getCurrentAtSign());
+      previousSyncService?.removeAllProgressListeners();
```
gkc: What should be the behaviour when we call switchAtSign and a sync is currently in progress?

murali-shris: * current behaviour:
   - pending sync requests are cleared
   - ongoing sync from server will be completed. But on client side, the responses will be ignored is my guess since sync service will not be active for the previous at client. I have requested sachin/nitesh to validate this.
* Ideal behaviour:
   solution 1:
    - inform the client/app through a warning/callback stating sync is in progress while switch atsign is attempted.
   solution 2:
   - gracefully stop the ongoing sync on client side and then perform switch atsign

gkc: Agreed. I'm leaning towards solution 2. I think we should add `stop()` methods to AtClient, to NotificationService and to SyncService, and have AtClientManager make explicit calls to those methods when switching atSigns. We can add a ServiceState enum e.g. "new, starting, started, stopping, stopped" SyncServiceImpl.stop() could stop its cron etc, set its overall state to "stopping", and the sync loop if it's in progress can check this state and gracefully exit its sync loop

gkc: "gracefully" means - state of key store and etc is good on the client side such that the next time a sync is run for that atSign from the client then it runs successfully

gkc (approved):

PR #876, opened 2023-01-10
fix: Fix pub score

### LINKED ISSUE #875: Fix pub dev score in at_client
### Describe the bug

pub.dev score now is 120/140. Fix the issues and get the score back to 140

### Steps to reproduce

Run "dart analyze "

### Expected behavior

There should be no errors.

### Screenshots

_No response_

_No response_

### Additional context

_No response_

purnimavenkatasubbu:
**- What I did**
fixed pub.dev score issue

**- How I did it**
Removed unnecessary imports, fixed dart format issues, and Added comments.

**- How to verify it**
All the tests should pass


packages/at_client/lib/src/service/encryption_service.dart
```diff
@@ -316,6 +318,7 @@ class EncryptionService {
       throw KeyNotFoundException('private encryption key not found');
     }
     var sharedKey =
+        // ignore: deprecated_member_use_from_same_package
```
murali-shris, line 321: instead of annotating in each place can you check whether there is an option for add lint rules in analysis_options.yaml

purnimavenkatasubbu: verified with  deprecated_member_use_from_same_package: ignore
It still throws issues.

murali-shris (approved):

PR #913, opened 2023-02-08
fix: Sync to local fails to delete a cached key

### LINKED ISSUE #908: cached key not getting deleted
### Describe the bug

When an atKey is deleted from the sender's side, the cached copy is not getting deleted from receiver's end.

AtKey :
```
 var key = AtKey()
      ..key = "test_delete"
      ..sharedBy = AtClientManager.getInstance().atClient.getCurrentAtSign()
      ..sharedWith = "@snowboard89"
      ..metadata = Metadata()
      ..metadata!.ccd = true
      ..metadata!.ttr = -1;
      
      await AtClientManager.getInstance().atClient.put(key, "value");
      
      AtClientManager.getInstance().atClient.delete(key);
```

On receiver's end:
`cached:@snowboard89:test_delete.atbuzz@local2010`

### Steps to reproduce

1. Share an AtKey like mentioned above.
2. share with an atSign and then delete.
3. lookup for the key on receiver's end

### Expected behavior

The cached key should get deleted.

### Screenshots

_No response_

### Smartphones

- iPhone 12 simulator

atBuzz

### Additional context

_No response_

sitaram-kalluri:
**- What I did**
[Problem]
* The cached key in the local secondary is not deleted when a sync response contains a delete operation.

[Root Cause]
* In sync_service_impl.dart, the string representation of the AtKey instance (received from the server via the sync response) is set to DeleteVerbBuilder.AtKey. When a "toString" is called on the "DeleteVerbBuilder", the "cached:<sharedWith>" is duplicated causing the incorrect key.

**- How I did it**
* Before calling "toString" on "DeleteVerbBuilder", convert the string representation of "AtKey" to "AtKey" object and set it to atKeyObj

**- How to verify it**
* Tested manually and cached keys get deleted in local secondary.
* Updated the unit tests and end2end tests to asset the changes.

* Sync to local fails to delete a cached key

### DIFF
--- a/packages/at_client/CHANGELOG.md
+++ b/packages/at_client/CHANGELOG.md
@@ -1,5 +1,6 @@
-## 3.0.54
+## 3.0.55
 - fix: Sync to local fails to delete a cached key
+## 3.0.54
 - fix: ensure forText notifications are decrypted successfully when using at_commons 3.0.35 or greater
 ## 3.0.53
 - feat: Introduce commit log compaction to keep size of commit log thin
--- a/packages/at_client/CHANGELOG.md
+++ b/packages/at_client/CHANGELOG.md
@@ -1,3 +1,5 @@
+## 3.0.54
+- fix: Sync to local fails to delete a cached key
 ## 3.0.53
 - feat: Introduce commit log compaction to keep size of commit log thin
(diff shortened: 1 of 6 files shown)


gkc (commented):
LGTM - nicely done!

gkc (approved):

PR #920, opened 2023-02-15
fix: onboard when the device is offline

sitaram-kalluri:
**- What I did**
[Problem]
 * Onboarding an atSign paired to a device fails when the network is down, This prevents fetching the data from the local storage when the device is offline.

[Root Cause]
  * During onboarding, the "OnBoardingStatus" is determined for which the encryption public key is fetched from the cloud secondary server. When the network is not reachable, fetching of encryption public key returns an exception causing the onboard process to fail.

**- How I did it**
[Solution]
 * Added a try-catch block around the code that fetches the encryption public key from the server. Upon exception, the error is logged and "null" is returned. 
 * Refactored the code in "_getServerEncryptionPublicKey" method as below:
      * Currently, the lookup verb command is used to fetch the encryption public key. On an authenticated connection, the lookup verb fails to fetch the data and returns "data:null". In this case, llookup verb is used to fetch the encryption public key. In a worst-case scenario, two calls to remote secondary are sent.
      * Modified the code such that If the connection is authenticated, then set the command to llookup verb, else set to command to lookup verb and run executeCommand to fetch the data.
 * Modified the condition to check the OnboardingStatus in the "getKeyRestorePolicy" method:
      * If the encryption public key is NULL on the cloud and local secondary (OR) encryption public key is NULL in cloud secondary and NOT-NULL is local secondary, then Onboarding Status is set to "Activate".
      * Also, if the encryption public key is NULL in cloud secondary and NOT-NULL is local secondary, then Onboarding Status is set to "SyncToServer".
      * Since the condition to check "SyncToSever" is the same as the condition after OR in the condition to check "Activate" the "SyncToServer" will not be returned.
      * Also the condition encryptionPublicKey NOT NULL in local secondary does not seem correct for "ACTIVATE", because if we generate encryption key pair only after activating the atSign. Hence removed the later part of the condition to check the "OnboardingStatus.ACTIVATE"
       
**- How to verify it**
  * Ran wavi app and atmosphere app with the changes and able to onboard the app.
  * Added a new unit test to onboard an app when remote secondary returns an exception.

* Onboard a paired atSign when the device is offline

### DIFF
--- a/packages/at_client_mobile/lib/src/at_client_service.dart
+++ b/packages/at_client_mobile/lib/src/at_client_service.dart
@@ -368,9 +368,7 @@ class AtClientService {
     var localEncryptionPublicKey =
         await keyChainManager.getEncryptionPublicKey(atSign);
     if (_isNullOrEmpty(localEncryptionPublicKey) &&
-            _isNullOrEmpty(serverEncryptionPublicKey) ||
-        (_isNullOrEmpty(serverEncryptionPublicKey) &&
-            !(_isNullOrEmpty(localEncryptionPublicKey)))) {
+        _isNullOrEmpty(serverEncryptionPublicKey)) {
       return OnboardingStatus.ACTIVATE;
     } else if (!_isNullOrEmpty(serverEncryptionPublicKey) &&
         _isNullOrEmpty(localEncryptionPublicKey)) {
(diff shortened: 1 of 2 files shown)


gkc:
Please test at_talk

sitaram-kalluri:
> Please test at_talk

@gkc: Sure.

murali-shris (approved):

gkc (commented):
@sitaram-kalluri Is this PR still reviewable, or are we going for an alternative solution?

sitaram-kalluri:
> @sitaram-kalluri Is this PR still reviewable, or are we going for an alternative solution?

@gkc: We are going with an alternate solution. So closing this PR.

PR #925, opened 2023-02-17
fix: at_client.putInternal() throws null error

srieteja:
**- What I did**
- AtKey.namespace has a null check in putInternal() that is throwing a null check error when a key is created without a namespace.
- Added a null check to the if-clause that checks for any upper case characters in AtKey's key and namespace.

**- How to verify it**
- when a key is created without namespace an exception should not be thrown.

fix: at_client.putInternal() throws null error

### DIFF
--- a/packages/at_client/CHANGELOG.md
+++ b/packages/at_client/CHANGELOG.md
@@ -1,3 +1,5 @@
+## 3.0.56
+-fix: AtClient.put() throws null-check error when key's namespace is null
 ## 3.0.55
 - fix: Amend Monitor's socket message handler so that it separates multiple 'simultaneous' responses correctly.
 - fix: Sync to local fails to delete a cached key
--- a/packages/at_client/lib/src/client/at_client_impl.dart
+++ b/packages/at_client/lib/src/client/at_client_impl.dart
@@ -474,8 +474,7 @@ class AtClientImpl implements AtClient, AtSignChangeListener {
       atKey.namespace ??= preference?.namespace;
     }
 
-    if (upperCaseRegex.hasMatch(atKey.key!) ||
-        upperCaseRegex.hasMatch(atKey.namespace!)) {
+    if (upperCaseRegex.hasMatch(atKey.key!)) {
       _logger.info(
           'AtKey: ${atKey.key}.${atKey.namespace} contains upper case characters,'
           'converting the key to lower case');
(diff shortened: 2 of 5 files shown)


gkc:
@srieteja I've pushed a change which I think resolves this issue cleanly. Please take a look

srieteja:
@gkc calling toString() on AtKey will convert the key and it's parts to lowercase. For the sake of this if-clause we have to preserve the case of AtKey.

srieteja:
One thing that's bugging me is that there is a verifyPutRequest() before this if-clause, which should throw an exception if the namespace is null. With that check present the control flow should not trigger this if-clause for an AtKey with 'null' namespace.

gkc:
> One thing that's bugging me is that there is a verifyPutRequest() before this if-clause, which should throw an exception if the namespace is null. With that check present the control flow should not trigger this if-clause for an AtKey with 'null' namespace.

It is perfectly OK for namespace to be null

validatePutRequest says that it isn't OK is for namespace to be null both in the key AND in the AtClientPreference
```dart
    if ((atKey.namespace == null || atKey.namespace!.isEmpty) &&
        (atClientPreference.namespace == null ||
            atClientPreference.namespace!.isEmpty)) {
      throw AtKeyException('namespace is mandatory');
    }
```

gkc:
> @gkc calling toString() on AtKey will convert the key and it's parts to lowercase. For the sake of this if-clause we have to preserve the case of AtKey.

Gotcha. Have pushed a change which reverts to the original but includes a null check

gkc (changes requested):
Please
1. Update pubspec, changelog and AtClientConfig
2. Add at least two unit tests (one with a null namespace, the other with a non-null namespace which contains some upper-case chars) which prove the change works.

gkc (changes requested):
Unit tests?

srieteja:
> Unit tests?

Working on them. There's a minor inconvenience, I'm unable to pass a mock object of the PutRequestTransformer into putInternal() trying to figure out.

gkc:
@srieteja Simplest way to enable these unit tests to be written: extract the relevant code into its own place and just test that. I've pushed a change with that refactoring.

gkc:
@srieteja Now you can write unit tests which create AtKeys with various combinations of key and namespace, then call the `ensureLowerCase(AtKey atKey)` method on the AtClientImpl, and then verify the state of the AtKey afterwards

gkc (approved):
LGTM

PR #936, opened 2023-02-28
fix: Invalid atKey in notification_response_transformer.dart

### LINKED ISSUE #772: Calling toString on the notification atKey has atsign twice prefixed and appended
### Describe the bug

When atKey.toString is called on the key received in the notification callback, the resultant has the duplicate atsign's. 
For example, the key looks like `@baboon9blue:@baboon9blue:shared_key@81bored7@81bored7` instead of `@baboon9blue:shared_key@81bored7`.

### Steps to reproduce

1. Send a notification of the type messageType.key from the server 
2. On the at_client SDK, when the notification is received, call toString on the atKey and the resultant has atsign twice prefixed and appended.

Root Cause:

In the update_verb_handler.dart, we prefix and append the sharedWith and sharedBy to the key before sending the notification. So, the key in the notification looks like "@‎bob:phone@alice"; when logging the key, we have an atKey.toString which will again add sharedWith and sharedBy to the key. Hence the atsign's are shown twice

### Expected behavior

When toString is called on the atKey, the resultant should have atSign's prefixed and appended only once.

sitaram-kalluri:
**- What I did**
- The PR fixes the git issue: https://github.com/atsign-foundation/at_client_sdk/issues/772

**- How I did it**
- Trim the sharedWith and sharedBy from the atKey.key received in the notification call back

**- How to verify it**
- Manually tested the changes and looks good.
- The existing unit and e2e tests should pass.

- Incorrect atKey.toString() in notification_response_transformation.dart

### DIFF
--- a/packages/at_client/CHANGELOG.md
+++ b/packages/at_client/CHANGELOG.md
@@ -1,5 +1,7 @@
+## 3.0.57
+- fix: Incorrect atKey.toString() in notification_response_transformation.dart
 ## 3.0.56
--fix: AtClient.put() throws null-check error when key's namespace is null
+- fix: AtClient.put() throws null-check error when key's namespace is null
 ## 3.0.55
 - fix: Amend Monitor's socket message handler so that it separates multiple 'simultaneous' responses correctly.
 - fix: Sync to local fails to delete a cached key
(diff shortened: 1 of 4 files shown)


gkc (approved):
LGTM

PR #940, opened 2023-03-02
test: Added e2e test for syncing cached key deletion

### LINKED ISSUE #932: Cached keys are not synced to cloud secondary when entry.operation equal CommitOp.DELETE
### Describe the bug

When I delete a cached key from my app, then uninstall and reinstall the app on iOS simulator the cached key reappears.

### Steps to reproduce

1. First I  created a key using `await atClientManager.atClient.notificationService.notify(NotificationParams.forUpdate(),)`
2. Then I tried deleting the key on the recipient end using `await atClientManager.atClient.delete()`
3. And then `flutter: FINER|2023-02-24 13:24:40.325022|SyncService (@batman87)|cached:@batman87:9089b66e-87f9-4802-b93f-d1b4c05374a3.at_dude@blizzard81 is skipped. cached keys will not be synced to cloud secondary` appear in my debug console but the key is deleted in my app.
4. Then I delete the app from the iOS simulator
5. When I reinstall the app on the IOS simulator the cached keys reappears.

### Expected behavior

I expected the deleted cached keys to remain deleted even if I uninstall and reinstall the app after deleting the cached key.

### Screenshots

_No response_

### Smartphones

- Device: Iphone 12
- OS: [e.g. iOS15.5]
- Browser [nil]
- Version [15.5]

AtDude

### Additional context

_No response_

purnimavenkatasubbu:
**- What I did**
Added an e2e test 
**- How I did it**
Created a cached key and deleted that cached key at the receiver's local secondary and verified the key in the remote secondary after sync
**- How to verify it**

### DIFF
--- a/tests/at_end2end_test/test/deletion_key_test.dart
+++ b/tests/at_end2end_test/test/deletion_key_test.dart
@@ -63,6 +63,9 @@ void main() {
           ..sharedWith(sharedWithAtSign)
           ..cache(-1, true))
         .build();
+    currentAtClientManager = await AtClientManager.getInstance()
+        .setCurrentAtSign(currentAtSign, namespace,
+            TestPreferences.getInstance().getPreference(currentAtSign));
     await currentAtClientManager.atClient.put(atKey, 'dummy_cached_value');
     await E2ESyncService.getInstance()
         .syncData(currentAtClientManager.atClient.syncService);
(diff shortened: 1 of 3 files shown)


tests/at_end2end_test/test/deletion_key_test.dart
```diff
@@ -106,4 +109,75 @@ void main() {
...
+        false);
+    // Asserts cached key is deleted from the server in the sharedWith atSign
+    var scanResultAfterDelete = await sharedWithAtClientManager.atClient
+        .getRemoteSecondary()!
+        .executeCommand('scan\n', auth: true);
+    expect(
+        scanResultAfterDelete!.contains(
+            'cached:$sharedWithAtSign:testcachedkey.wavi$currentAtSign'),
+        false);
+  }, timeout: Timeout(Duration(minutes: 1)));
```
gkc, line 183: Looks good. Couple of things which could be improved:
* Suggest you use 'currentAtSign' to 'sharedByAtsign' for clarity
  `var sharedByAtsign = currentAtSign` if necessary
* The name 'sharedWithAtClientManager' suggests that it will always return sharedWith atClient.
  Instead of 
    ```
      sharedWithAtClientManager = await AtClientManager.getInstance()
          .setCurrentAtSign(sharedWithAtSign, namespace,
              
      TestPreferences.getInstance().getPreference(sharedWithAtSign));  
    ```
  I'd suggest
    ```
    AtClient sharedWithAtClient = await AtClientManager.getInstance()
        .setCurrentAtSign(sharedWithAtSign, namespace,
        TestPreferences.getInstance().getPreference(sharedWithAtSign))
        .atClient;
  ```
  and subsequently use `sharedWithAtClient` rather than `sharedWithAtClientManager.atClient`
* There is repetition of literal values (e.g. 'testCachedKey',  'cached:$sharedWithAtSign:testcachedkey.wavi$currentAtSign}' which should be extracted into variables

gkc (changes requested):

tests/at_end2end_test/test/deletion_key_test.dart
purnimavenkatasubbu: Updated the code as per the suggestions

gkc (approved):

PR #956, opened 2023-03-14
fix: Add value validations

### LINKED ISSUE #923: client-side validation of value length
### This ticket documents several related issues

***

### 1 - Missing client-side validation on length of value for notifications
The validation in AtClientValidation.validatePutRequest includes a check on the length of the value
```dart
    if (value.length > atClientPreference.maxDataSize) {
      throw BufferOverFlowException(
          'The length of value exceeds the buffer size. Maximum buffer size '
          'is ${atClientPreference.maxDataSize} bytes. Found ${value.toString().length} bytes');
    }
```

However, the validation in AtClientValidation.validateNotificationRequest includes no such check. As a result, it is possible to attempt to send notifications with payloads larger than the server maximum (approximately 10,230,000 bytes), which will ultimately lead to a waste of a lot of time and network traffic, because the server will reject it

The obvious fix would be to add the same size validation to validateNotificationRequest. However, there is also ...

### 2 - Payload size validation should be done **_after_** transformation for transmission or storage
The validation referred to above in validatePutRequest is being called before calling the PutRequestTransformer. As a result, the length being checked is the length _before_ encryption and encoding. We should be validating the length __after__ the encryption and encoding.

### 3 - The documentation for AtClientPreference.maxDataSize is incorrect and misleading. Currently it says:
```dart
  /// Maximum data size a secondary can store. Temporary solution. Have to fetch this from
  /// server using stats verb.
  int maxDataSize = 10*1024*1024;
```
There are several problems here.
1. Given the current implementations, the only useful semantic of maxDataSize is "the maximum size of a request which the server can handle.
2. The actual maximum **_request_** size is set in the InboundMessageListener class `final _buffer at_commons.ByteBuffer(capacity: 10240000);`
3. There is no way to fetch that value from the server using the stats verb
4. The full request includes the atProtocol verb, the metadata and the actual value (which may be encrypted and encoded). Verb and metadata together will never exceed 10,000 bytes in length, so a sensible maxDataSize to set in AtClientPreference would be 10230000

sitaram-kalluri:
**- What I did**
* Fixes the git issue: https://github.com/atsign-foundation/at_client_sdk/issues/923

**- How I did it**
* Added IF conditions to check if the value size exceeds the buffer limit for put and notify methods.

**- How to verify it**
* Added unit test to verify put methods throws an exception when value exceeds the maximum buffer limit.
* Added unit test to verify notify method throws an exception when the value exceeds the maximum buffer limit.

TODO: Refactor the notify method in the "notification_service_impl.dart" to pass the mock object of the "NotificationRequestTransformer" instance and assert the buffer overflow exception is caught and propagated back to the client. Prefer making these changes in a separate PR.

* Added condition to prevent value exceeding the buffer limit.


packages/at_client/lib/src/preference/at_client_preference.dart
```diff
@@ -43,7 +43,7 @@ class AtClientPreference {
 
   /// Maximum data size a secondary can store. Temporary solution. Have to fetch this from
```
murali-shris: we have to change this documentation. See # 3 in the ticket.

packages/at_client/lib/src/client/at_client_impl.dart
```diff
@@ -517,8 +520,15 @@ class AtClientImpl implements AtClient, AtSignChangeListener {
...
     // Transform put request
     // Optionally passing encryption private key to sign the public data.
-    UpdateVerbBuilder verbBuilder = await PutRequestTransformer(this)
-        .transform(tuple, encryptionPrivateKey: encryptionPrivateKey);
+    UpdateVerbBuilder verbBuilder = await putRequestTransformer.transform(tuple,
+        encryptionPrivateKey: encryptionPrivateKey);
+    // Validate the size of the value after encryption/encoding
+    // Since AtClientPreference is mandatory argument in create method, _preference
+    // will not be null.
+    if (verbBuilder.value.length > _preference!.maxDataSize) {
```
murali-shris, line 528: similar check can be done in remoteSecondary.executeCommand. if final command to secondary greater than what secondary can accept

murali-shris (dismissed):

packages/at_client/lib/src/client/at_client_impl.dart
sitaram-kalluri: Added check in executeCommand in RemoteSecondary.

packages/at_client/lib/src/preference/at_client_preference.dart
sitaram-kalluri: Updated the dart docs

packages/at_client/lib/src/client/at_client_impl.dart
```diff
@@ -517,8 +520,15 @@ class AtClientImpl implements AtClient, AtSignChangeListener {
...
+      throw BufferOverFlowException(
+          'The length of value exceeds the buffer size. Maximum buffer size is ${_preference!.maxDataSize} bytes. Found ${value.toString().length} bytes');
```
gkc: Say "exceeds the maximum allowed length." instead of "exceeds the buffer size."

packages/at_client/lib/src/client/remote_secondary.dart
```diff
@@ -85,6 +85,10 @@ class RemoteSecondary implements Secondary {
   }
 
   Future<String?> executeCommand(String atCommand, {bool auth = false}) async {
+    if (atCommand.length > _preference.maxDataSize) {
+      throw BufferOverFlowException(
+          'The length of value exceeds the buffer size. Maximum buffer size is ${_preference.maxDataSize} bytes. Found ${atCommand.length} bytes');
```
gkc: Say "exceeds the maximum allowed length." instead of "exceeds the buffer size."

packages/at_client/lib/src/service/notification_service_impl.dart
```diff
@@ -392,6 +393,34 @@ class NotificationServiceImpl
...
+  void notificationValueValidation(
+      NotificationParams notificationParams, NotifyVerbBuilder builder) {
+    switch (notificationParams.messageType) {
+      case MessageTypeEnum.key:
+        // Since value is not mandatory in AtNotification, perform validation only if
+        // value is not null.
+        if (builder.value != null &&
+            builder.value.length > _atClient.getPreferences()!.maxDataSize) {
+          throw BufferOverFlowException(
+              'The length of value exceeds the buffer size. Maximum buffer size is ${_atClient.getPreferences()!.maxDataSize} bytes. Found ${builder.value.toString().length} bytes');
```
gkc: Say "exceeds the maximum allowed length." instead of "exceeds the buffer size."

packages/at_client/lib/src/service/notification_service_impl.dart
```diff
@@ -392,6 +393,34 @@ class NotificationServiceImpl
...
+        }
+        break;
+
+      case MessageTypeEnum.text:
+        // When messageType is text, the "text" to notify is added to the key. Hence validating
+        // the key length
+        if (builder.atKey!.length > _atClient.getPreferences()!.maxDataSize) {
+          throw BufferOverFlowException(
+              'The length of value exceeds the buffer size. Maximum buffer size is ${_atClient.getPreferences()!.maxDataSize} bytes. Found ${builder.value.toString().length} bytes');
```
gkc: Say "exceeds the maximum allowed length." instead of "exceeds the buffer size."

gkc (changes requested):
Looking good! I've made a few small requests to improve the readability of the thrown exception's message

packages/at_client/lib/src/service/notification_service_impl.dart
sitaram-kalluri: Updated the exception message.

packages/at_client/lib/src/service/notification_service_impl.dart
sitaram-kalluri: Updated exception message.

packages/at_client/lib/src/client/remote_secondary.dart
sitaram-kalluri: Updated exception message

packages/at_client/lib/src/client/at_client_impl.dart
sitaram-kalluri: Updated exception message.

gkc (approved):

PR #968, opened 2023-03-20
feat: Initial support of additional encryption metadata enabling encryption future-proofing

gkc:
**- What I did**
feat: Initial support of additional encryption metadata enabling encryption future-proofing

**- How I did it**
- feat: Added awareness of new encryption-related metadata into various classes
- feat: Added ivBase64 optional parameters to the relevant methods so we can use the iv/nonce metadata
- feat: Added support for new _decryption_ behaviour
- feat: Added `AtClientPreference.atProtocolEmitted` which controls _encryption_ behaviour
- test: Added full-client-stack _unit_ tests for put and get, for self data and shared data, with a mock RemoteSecondary, to verify legacy and latest encryption behaviour
- _**fix:**_ Fixed a bug in AbstractAtKeyEncryption.getSharedKey which was always trying to load private key from the key store even if we are using an AtChops
- test: Added new _end-to-end_ encryption / decryption tests
- feat: Added PutRequestOptions to AtClient's put(), putText() and putBinary() to allow to specify whether to include the encrypted shared key in shared records' metadata
- test: Added more end-to-end tests which exercise the PutRequestOptions functionality

**- How to verify it**
Tests pass


packages/at_client/lib/src/client/at_client_impl.dart
```diff
@@ -903,107 +857,6 @@ class AtClientImpl implements AtClient, AtSignChangeListener {
     return _preference;
   }
 
-  @override
-  Future<String?> notifyChange(NotificationParams notificationParams) async {
```
gkc, line 907: As this method was deprecated a long time ago, marked it deprecated here also, moved it to the end of this file, and made it a simple wrapper around NotificationService.notify()

packages/at_client/lib/src/client/request_options.dart
```diff
@@ -5,3 +5,7 @@ class RequestOptions {}
 class GetRequestOptions extends RequestOptions {
   bool bypassCache = false;
 }
+
+class PutRequestOptions extends RequestOptions {
+  bool storeSharedKeyEncryptedWithData = true;
```
murali-shris: storeSharedKeyEncryptedMetadata will be a better name?

packages/at_client/lib/src/converters/encryption/aes_converter.dart
```diff
@@ -99,3 +108,11 @@ class AESDecryptionSink extends ChunkedConversionSink<List<int>> {
     _outSink.close();
   }
 }
+
+IV getIV(String? ivBase64) {
+  if (ivBase64 == null) {
+    return IV.fromLength(16);
```
murali-shris, line 114: IV.fromLength(16) always returns all zeros. This is a security vulnerability since IV becomes a constant list. There was an issue raised by a git user for this. But I am unable to find the issue. We should use IV.fromSecureRandom(16)

packages/at_client/lib/src/util/encryption_util.dart
```diff
@@ -9,25 +9,36 @@ import 'package:at_utils/at_logger.dart';
 
 class EncryptionUtil {
   static final _logger = AtSignLogger('EncryptionUtil');
+
+  static IV getIV(String? ivBase64) {
+    if (ivBase64 == null) {
+      return IV.fromLength(16);
```
murali-shris, line 15: replace with IV.fromSecureRandom(16)

packages/at_client/pubspec.yaml
```diff
@@ -28,15 +28,16 @@ dependencies:
   http: ^0.13.5
   internet_connection_checker: ^1.0.0+1
   async: ^2.9.0
-  at_persistence_spec: ^2.0.10
-  at_persistence_secondary_server: ^3.0.51
+  at_persistence_spec: 2.0.12
```
murali-shris: this should be ^2.0.12.
Dart won't allow fixed version for published packages

murali-shris (changes requested):

packages/at_client/lib/src/client/request_options.dart
gkc: Changed.

packages/at_client/lib/src/converters/encryption/aes_converter.dart
gkc: Indeed. The entire point of this PR is to resolve that issue for us, and enable us to start using IVs

packages/at_client/lib/src/util/encryption_util.dart
gkc: No. The entire point of this PR is to enable us to start using IVs. Current versions of the SDK do not use IVs. This version of the SDK (1) will be able to decrypt when a non-zeros IV is supplied in metadata (2) will be able to set and use a non-zeros IV when encrypting.

packages/at_client/pubspec.yaml
gkc: Changed

murali-shris (approved):

PR #971, opened 2023-03-21
fix: Set isCached to true in metadata for cached keys

### LINKED ISSUE #939: prepareMetadata sets isCached to false
### Describe the bug

prepareMetadata() sets isCached to false by default even for a cached key

### Steps to reproduce

1. First from atsign-1 notify a key with ttr to atsign-2
2. After sync, Do a get() of the cached key in the atsign-2
3. Check the metadata of the key. 

### Expected behavior

isCached should be set to true for a cached key

### Screenshots

_No response_

### Smartphones

_No response_

_No response_

### Additional context

_No response_

sitaram-kalluri:
**- What I did**
* Fix "isCached" in the metadata is not set to "true" when fetching the cached keys in AtClient.get method.

**- How I did it**
* If the key starts with "cached:" set metadata.isCached to true
* Updated the existing end-to-end test - "bypasscache_test.dart"
   * The test failed because, in the get method, the metadata.isCached of "getKey" variable is being set to true. So, subsequently, when the lookup verb is executed,  "cached:" is prefixed in the string representation of the key causing an incorrect key - cached:null:verificationnumber@alice. Fixed the issue by initializing the getKey variable every time before performing the look verb.

**- How to verify it**
* Updated the existing e2e test to assert metadata.isCached is set to true

* Set metadata.isCached to true for cached keys


gkc (approved):
LGTM

We should have a follow-up session to eliminate AtClientUtil.prepareMetadata as this functionality should properly reside in commons (and probably already does)

PR #983, opened 2023-03-30
fix: Invalid key in sync causes infinite loop

### LINKED ISSUE #955: Sync failing to complete due to incomplete handling of exceptions that the client may encounter
See https://github.com/atsign-foundation/at_client_sdk/issues/951 for context

The client SyncService should handle any exception it encounters while writing data locally, and ensure that the sync never gets stuck in a loop

sitaram-kalluri:
**- What I did**
* Fix sync running into an infinite loop when an invalid key is sent from the server in the sync response.

**- How I did it**
* Introduce a new variable "lastReceivedServerCommitId" which keeps track of the last commit entry sent by the server. When fetching the next sent of entries to sync, the client sends the "lastReceivedServerCommitId" instead of "lastSyncedCommitId" into the local keystore.

**- How to verify it**
* Updated the unit tests to accommodate the changes.

* Fix: Sync running into infinite loop when an invalid key is present in the entries to sync into client

### DIFF
--- a/packages/at_client/lib/src/service/sync_service_impl.dart
+++ b/packages/at_client/lib/src/service/sync_service_impl.dart
@@ -775,17 +775,14 @@ class SyncServiceImpl implements SyncService, AtSignChangeListener {
   Future<int> _getLastReceivedServerCommitId() async {
     // If "lastReceivedServerCommitId" key exists, fetch the data and return the
     // last received server commit id.
-    var isLastReceivedServerKeyExists = _atClient
-        .getLocalSecondary()!
-        .keyStore!
-        .isKeyExists(_lastReceivedServerCommitIdAtKey.toString());
-    if (isLastReceivedServerKeyExists) {
+    try {
       var response = await _atClient.get(_lastReceivedServerCommitIdAtKey);
       return int.parse(response.value);
+    } on KeyNotFoundException {
+      // If the key does not exist, fall back to previous logic, which is
+      // return last synced commit id.
+      return _getLocalCommitId();
     }
-    // If the key does not exist, fall back to previous logic, which is
-    // return last synced commit id.
-    return _getLocalCommitId();
   }
 
   /// Returns the local commit id. If null, returns -1.
--- a/packages/at_client/lib/src/service/sync_service_impl.dart
+++ b/packages/at_client/lib/src/service/sync_service_impl.dart
@@ -778,7 +778,7 @@ class SyncServiceImpl implements SyncService, AtSignChangeListener {
     try {
       var response = await _atClient.get(_lastReceivedServerCommitIdAtKey);
       return int.parse(response.value);
-    } on KeyNotFoundException {
+    } on AtKeyNotFoundException {
       // If the key does not exist, fall back to previous logic, which is
       // return last synced commit id.
       return _getLocalCommitId();
--- a/packages/at_client/lib/src/service/sync_service_impl.dart
+++ b/packages/at_client/lib/src/service/sync_service_impl.dart
@@ -59,6 +59,9 @@ class SyncServiceImpl implements SyncService, AtSignChangeListener {
   /// Returns the currentAtSign associated with the SyncService
   String get currentAtSign => _atClient.getCurrentAtSign()!;
 
+  /// A local AtKey to persist the last received server commitId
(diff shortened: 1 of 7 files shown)


packages/at_client/lib/src/service/sync_service_impl.dart
```diff
@@ -748,6 +772,22 @@ class SyncServiceImpl implements SyncService, AtSignChangeListener {
...
   }
 
+  Future<int> _getLastReceivedServerCommitId() async {
+    // If "lastReceivedServerCommitId" key exists, fetch the data and return the
+    // last received server commit id.
+    var isLastReceivedServerKeyExists = _atClient
+        .getLocalSecondary()!
+        .keyStore!
+        .isKeyExists(_lastReceivedServerCommitIdAtKey.toString());
+    if (isLastReceivedServerKeyExists) {
```
murali-shris: can we simplify the method like this
try  {
var response = await _atClient.get(_lastReceivedServerCommitIdAtKey);
      return int.parse(response.value);
} on KeyNotFoundException {
 return _getLocalCommitId();
}

packages/at_client/lib/src/service/sync_service_impl.dart
```diff
@@ -406,93 +409,111 @@ class SyncServiceImpl implements SyncService, AtSignChangeListener {
...
-        break;
-      }
-      // Iterates over each commit
-      for (dynamic serverCommitEntry in syncResponseJson) {
-        _sendTelemetry('_syncFromServer.forEachEntry.start', {
-          "atKey": serverCommitEntry['atKey'],
-          "operation": serverCommitEntry['operation'],
-          "commitId": serverCommitEntry['commitId'],
+    try {
+      while (serverCommitId > lastReceivedServerCommitId) {
```
murali-shris, line 422: what will happen here when initially there is connection to perform sync and in between sync connection drops.

murali-shris (approved):

murali-shris (changes requested):
re-request

packages/at_client/lib/src/service/sync_service_impl.dart
sitaram-kalluri: Sure Murali

packages/at_client/lib/src/service/sync_service_impl.dart
sitaram-kalluri: In "_syncFromServer" method, the lastReceivedServerCommitId is updated at the end of the batch response. The logic to update the lastReceivedServerCommitId is wrapped inside a "finally" block to ensure the lastReceivedServerCommitId is updated even if an exception occurs.

So, When network drops or if sync stops in between, the lastReceivedServerCommitId is presisted and in the next run, the lastReceivedServerCommitId is fetched and request the server for the commit entries from that point.

packages/at_client/lib/src/service/sync_service_impl.dart
```diff
@@ -59,6 +59,9 @@ class SyncServiceImpl implements SyncService, AtSignChangeListener {
   /// Returns the currentAtSign associated with the SyncService
   String get currentAtSign => _atClient.getCurrentAtSign()!;
 
+  /// A local AtKey to persist the last received server commitId
+  late AtKey _lastReceivedServerCommitIdAtKey;
```
gkc: Please mark this variable as final as well as late

gkc (changes requested):
Mainly LGTM. Have left a request for a small change and have two questions
1. Have you tested using an app?
2. Are there unit tests which simulate the server sending 'invalid' keys in the sync response?

sitaram-kalluri:
> Mainly LGTM. Have left a request for a small change and have two questions
> 
> 1. Have you tested using an app?
> 2. Are there unit tests which simulate the server sending 'invalid' keys in the sync response?

On 1. I am tested with the buzz app, Yet to test with wavi app.
On 2. Added a unit test to assert sync do no enter into infinite loop when there is a bad key in sync response.

packages/at_client/lib/src/service/sync_service_impl.dart
sitaram-kalluri: Sure Gary.

gkc (approved):

murali-shris (approved):

PR #989, opened 2023-04-07
fix: Enhance the metadata to retain existing metadata or unset the metadata

sitaram-kalluri:
**-What I did**
* Enhance the metadata as follows:
  * When an attribute is set to a different value than the existing one, overwrite with the new value
  * When an attribute is not set, then fetch the existing metadata and add it to the current metadata
  * When an attribute is explicitly set to null, reset the metadata.

**-How I did**
* In LocalSecondary, introduce a method "_unsetOrRetainMetadata", that has the implementation for the above behaviour.
*  Refactoring: Move the code that sets the metadata in the "executeVerb" method to a new method "_getAtMetadataFromVerbBuilder".

**-How to verify it**
* Added the following unit tests that assert the behaviour
  * A test to verify existing metadata is retained when new metadata is set
  * A test to verify new value is set to the old value
  * A test to verify setting null reset the metadata

**-Description for CHANGELOG**
* fix: Enhance the metadata to retain existing metadata or unset the metadata


gkc (changes requested):
LGTM. However as this is in a core part of the SDK, I believe we need to test all of our core apps against this branch - at_talk, sshnoports, atmospherePro, atBuzz, ...

sitaram-kalluri:
> 

Sure, Gary.

purnimavenkatasubbu:
Verified the branch with atBuzz, atWavi, atMospherePro, and at_talk.Everything looks good. Attaching the logs.
[atmosphereProMetadatalogs.txt](https://github.com/atsign-foundation/at_client_sdk/files/11220346/atmosphereProMetadatalogs.txt)
[atTalkMetadataLogs.txt](https://github.com/atsign-foundation/at_client_sdk/files/11220348/atTalkMetadataLogs.txt)
[buzzMetadataLogs.txt](https://github.com/atsign-foundation/at_client_sdk/files/11220349/buzzMetadataLogs.txt)
[waviMetadataBranchOldAtsignLogs.txt](https://github.com/atsign-foundation/at_client_sdk/files/11220352/waviMetadataBranchOldAtsignLogs.txt)
[waviMetadatNewAtsignLogs.txt](https://github.com/atsign-foundation/at_client_sdk/files/11220354/waviMetadatNewAtsignLogs.txt)

gkc:
After discussion with @sitaram-kalluri we are going to take a different approach to ensure that (1) changes from atServer sync responses are always fully applied to local data store (2) make it possible in the client to accurately handle situations where a previously-non-null metadata field is deliberately set to null by application code

PR #991, opened 2023-04-10
fix: Update lastReceivedServerCommitId when uncommitted entries are synced

### LINKED ISSUE #990: The "lastReceivedServerCommitId" AtKey does not update when uncommitted entries are synced to server
### Describe the bug

Referring to the issue https://github.com/atsign-foundation/at_client_sdk/issues/955, "lastReceivedServerCommitId" key is introduced and holds the last entry the client received from the server. The key keeps updating when a new entry is synced from the server to the client to keep track of the latest server commit-id. 

The key is NOT updated when an uncommitted entry is synced to the server and therefore, the lastReceivedServerCommitId falls behind the actually synced localCommitId. 

For example, let's say, the serverCommitId is at 10 and localCommitId is at 8 and there is one uncommitted entry in the client.
After the server changes are synced to the client, the localCommitId will move to 10 and so is the "lastReceivedServerCommitId". Then the uncommitted entry is synced to the server and so, the serverCommitId moves to 11 and the commit-entry in local storage will be updated with the commit-id sent by the server (here 11), making the localCommitId 11. But since the lastReceivedServerCommitId is updated when syncing the uncommitted entries, the "lastReceivedServerCommitId" remains at 10. 

Therefore, in the subsequent sync, the entries from commit-Id 10 are fetched, instead of fetching from commit-id 11. If the last entry and the uncommitted entry happen to be the same key, then the value of the uncommitted entry will be overwritten by the value from the server.

### Steps to reproduce

Refer to "test/bypasscache_test.dart: bypass cache test" in at_end-to-end tests.

### Expected behaviour

* The test - "test/bypasscache_test.dart: bypass cache test" should pass.
* The lastReceivedServerCommitId must be updated when an uncommitted entry is synced to the server.

sitaram-kalluri:
**- What I did**
* Fix the issue: https://github.com/atsign-foundation/at_client_sdk/issues/990

**- How I did it**
* Update the "lastReceivedServerCommitId" when uncommitted entries are synced to the server and the respective entries in the client are updated. This fixes the issue-990 mentioned above.

**- How to verify it**
* Added unit test to verify the "lastReceivedServerCommitId" is updated when an uncommitted entry is synced to the server

### DIFF
--- a/packages/at_client/lib/src/service/sync_service_impl.dart
+++ b/packages/at_client/lib/src/service/sync_service_impl.dart
@@ -392,8 +392,13 @@ class SyncServiceImpl implements SyncService, AtSignChangeListener {
             await syncUtil.updateCommitEntry(
                 commitEntry, commitId, _atClient.getCurrentAtSign()!);
             // update the last received server commit id
-            await _atClient.put(
-                _lastReceivedServerCommitIdAtKey, commitId.toString());
+            // When an invalid key is received, the commit-id is -1, and in this
+            // case the "lastReceivedServerCommitId" must NOT be updated, because
+            // we do not want lose the track of latest server commit id.
+            if (commitId != -1) {
+              await _atClient.put(
+                  _lastReceivedServerCommitIdAtKey, commitId.toString());
+            }
             keyInfoList.add(KeyInfo(commitEntry.atKey,
                 SyncDirection.localToRemote, commitEntry.operation));
           } on Exception catch (e) {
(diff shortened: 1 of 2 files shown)


packages/at_client/lib/src/service/sync_service_impl.dart
```diff
@@ -391,6 +391,9 @@ class SyncServiceImpl implements SyncService, AtSignChangeListener {
             _logger.finer('***batchId:$batchId key: ${commitEntry.atKey}');
             await syncUtil.updateCommitEntry(
                 commitEntry, commitId, _atClient.getCurrentAtSign()!);
+            // update the last received server commit id
+            await _atClient.put(
```
murali-shris: sometimes commitId can be -1 when there is no response from server or server returns data:-1. 
I guess we  should not update the lastReceivedServerCommitId in this case

murali-shris (changes requested):

packages/at_client/lib/src/service/sync_service_impl.dart
sitaram-kalluri: Thanks, Murali. Added IF condition and pushed the changes.

murali-shris (approved):

PR #1004, opened 2023-04-26
fix: Failing sync functional test

sitaram-kalluri:
**- What I did**
* Fix the following functional test that fails intermittently - "Verifying sync progress - server ahead"

Error from the GitHub actions run: 

```
Expected: non-empty
    Actual: <null>
  NoSuchMethodError: The getter 'isNotEmpty' was called on null.
  Receiver: null
  Tried calling: isNotEmpty at #0      Object.noSuchMethod (dart:core-patch/object_patch.dart:38:5)
  #1      _NotEmpty.matches (package:matcher/src/core_matchers.dart:30:67)
  #2      _expect (package:matcher/src/expect/expect.dart:138:30)
  #3      expect (package:matcher/src/expect/expect.dart:56:3)
  #4      main.<anonymous closure>.<anonymous closure> (file:///home/runner/work/at_client_sdk/at_client_sdk/tests/at_functional_test/test/atclient_sync_callback_test.dart:190:9)
```
Failure occurred on the following Github actions run - https://github.com/atsign-foundation/at_client_sdk/actions/runs/4799158155

Root Cause:

The sync request is being forced to run twice, just as an additional measure, if the keys do not sync on the first run. 
At the moment, the assertions happen only on the second syncProgress callback. But if the sync happens to sync all the keys in the first run, then on the second run, there will be nothing to sync and the sync progress call back will have null like below

```
SyncProgress: SyncProgress{atSign: @alice🛠, syncStatus: SyncStatus.success, isInitialSync: false, startedAt: 2023-04-26 12:51:40.001219Z, completedAt: 2023-04-26 12:51:40.049674Z, message: null, keyInfoList:null, localCommitIdBeforeSync:null, localCommitId:null, serverCommitId:null}
```

**- How I did it**
* Added a check that asserts if localCommitIdBeforeSync does not equal localCommitId which means keys have synced. So assert on the sync-progress attributes.
If both are equal, it means there is nothing to sync and so no assertions, except for SyncStatus which should be "success"

**- How to verify it**
* The failing functional test should pass.

### DIFF
--- a/tests/at_functional_test/test/atclient_sync_callback_test.dart
+++ b/tests/at_functional_test/test/atclient_sync_callback_test.dart
@@ -180,13 +180,15 @@ void main() {
     await FunctionalTestSyncService.getInstance()
         .syncData(atClientManager.atClient.syncService);
 
-    int progressCallbacksReceived = 0;
-    int requiredProgressCallbacks = 2;
     progressListener.streamController.stream
         .listen(expectAsync1((SyncProgress syncProgress) {
-      progressCallbacksReceived++;
-      if (progressCallbacksReceived == requiredProgressCallbacks) {
-        expect(syncProgress.syncStatus, SyncStatus.success);
+      print('SyncProgress: $syncProgress');
+      expect(syncProgress.syncStatus, SyncStatus.success);
+      // If localCommitIdBeforeSync and localCommitId (local commitId after sync)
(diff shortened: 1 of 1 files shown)


gkc (approved):
LGTM - thanks @sitaram-kalluri !

PR #1007, opened 2023-04-27
feat: Add AtClientParticulars to enhance logging

### LINKED ISSUE #63: Introduce concept of a client ID

One current complication we face when diagnosing issues is that we don't have any way of distinguishing between the many different client programs which connect to the atServer.

While diagnosing https://github.com/atsign-foundation/at_server/issues/1287 we concluded that it would have been very helpful for each client to have an identifier (a guid) which is created once, held locally for that client, and sent as part of the client config which is sent via the `from` verb.

This gives us some information as to which app any given connection originated from e.g. the AtmospherePro app, an sshnp client, an at_talk client, the Buzz app, etc etc

_No response_

### Additional context

_No response_

### LINKED ISSUE #64: Additional information in client config

One current complication we face when diagnosing issues is that we don't have any way of distinguishing between the many different client programs which connect to the atServer.

In addition to the changes proposed in #63 (add a client ID) we should also allow the sending of additional information in the client config such as the name of the app, the version of the app, the platform upon which the app is running, etc

_No response_

### Additional context

See also #63 

This has been created as a separate issue for clarity but we could choose to implement them at the same time

sitaram-kalluri:
**- What I did**
* Introduce clientId, appName, appVersion and platform to distinguish requests from several clients
* Added clientId in the sync_service_impl and remote_secondary logs to compare with the server logs.

**- How I did it**
* Added a new class "AtClientParticulars" that contains the above-mentioned fields. Added "@‎experimental" tag to the class.
* The "AtClientParticulars" is exposed in the  "AtClientPreferences" for the apps to populate the appName, appVersion and platform.

### DIFF
--- a/packages/at_client/lib/src/preference/at_client_preference.dart
+++ b/packages/at_client/lib/src/preference/at_client_preference.dart
@@ -1,4 +1,5 @@
 import 'package:at_client/at_client.dart';
+import 'package:at_client/src/preference/at_client_particulars.dart';
 import 'package:meta/meta.dart';
 import 'package:version/version.dart';
 
@@ -100,6 +101,9 @@ class AtClientPreference {
   /// This instance variable is experimental, for now
   @experimental
   Version atProtocolEmitted = Version(1, 5, 0);
+
+  @experimental
+  AtClientParticulars atClientParticulars = AtClientParticulars.getInstance();
 }
 
 @Deprecated("Use SyncService")
--- a/packages/at_client/lib/src/preference/at_client_preference.dart
+++ b/packages/at_client/lib/src/preference/at_client_preference.dart
@@ -103,7 +103,7 @@ class AtClientPreference {
   Version atProtocolEmitted = Version(1, 5, 0);
 
   @experimental
-  AtClientParticulars atClientParticulars = AtClientParticulars.getInstance();
+  AtClientParticulars atClientParticulars = AtClientParticulars();
 }
 
 @Deprecated("Use SyncService")
--- a/packages/at_client/lib/src/service/sync_service_impl.dart
+++ b/packages/at_client/lib/src/service/sync_service_impl.dart
@@ -11,6 +11,7 @@ import 'package:at_client/src/response/json_utils.dart';
 import 'package:at_client/src/service/notification_service_impl.dart';
 import 'package:at_client/src/service/sync/sync_request.dart';
 import 'package:at_client/src/service/sync_service.dart';
+import 'package:at_client/src/util/logger_util.dart';
 import 'package:at_client/src/util/network_util.dart';
 import 'package:at_client/src/util/sync_util.dart';
 import 'package:at_commons/at_builders.dart';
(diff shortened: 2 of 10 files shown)


gkc (changes requested):
Have merged your at_commons PR

Please can you publish new version of at_commons please, then update pubspec in this PR and re-request review

packages/at_client/lib/src/preference/at_client_preference.dart
```diff
@@ -100,6 +101,9 @@ class AtClientPreference {
   /// This instance variable is experimental, for now
   @experimental
   Version atProtocolEmitted = Version(1, 5, 0);
+
+  @experimental
+  AtClientParticulars atClientParticulars = AtClientParticulars.getInstance();
```
gkc: Why a singleton?

packages/at_client/lib/src/service/sync_service_impl.dart
```diff
@@ -505,7 +507,8 @@ class SyncServiceImpl implements SyncService, AtSignChangeListener {
       ..regex = _atClient.getPreferences()!.syncRegex
       ..limit = _atClient.getPreferences()!.syncPageLimit
       ..isPaginated = true;
-    _logger.finer('** syncBuilder ${syncBuilder.buildCommand()}');
+    _logger.finer(
+        '${_atClient.getPreferences()!.atClientParticulars.clientId} | syncBuilder ${syncBuilder.buildCommand()}');
```
gkc: Why not add a client-only extension to AtSignLogger as you have done in at_server?

gkc (changes requested):

packages/at_client/lib/src/preference/at_client_preference.dart
sitaram-kalluri: Since the appName, appVersion and platform are unique for the at_client instance, In future, if we want to access the data inside AtClientParticulars anywhere in at_client, it would be easy to call AtClientParticulars.getInstance(), then relying on "AtClientManager.atClient.getPreferences..".

packages/at_client/lib/src/service/sync_service_impl.dart
sitaram-kalluri: Sure Gary, will update this it is in at_server and push the changes.

packages/at_client/lib/src/preference/at_client_preference.dart
gkc: What about when switching atSigns? Using a singleton forces each client to have the same clientId - this is not what we want.

Singletons should never be used unless there is a strong architectural reason to have them. There's no such reason here.

sitaram-kalluri: Got it Gary, will remove the singleton nature here.

gkc (approved):

gkc (approved):
LGTM thank you @sitaram-kalluri

PR #1017, opened 2023-05-04
test: Fix failing functional tests due to sync race conditions

### LINKED ISSUE #840: test: functional test atclient_sharedkey_test.dart is occasionally failing
### Describe the bug

functional test atclient_sharedkey_test.dart is occasionally failing

### Steps to reproduce

Has happened at least twice in the past three days. It eventually fails after a three minute timeout. 

### Expected behavior

test should pass. Likely a race condition related to checking if sync is complete.

purnimavenkatasubbu:
**- What I did**
I added a check to see if syncStatus is a success && localCommitId equals serverCommitID. If both are equal, Exit the code as sync completed.
Exit if sync process exceeds 30 seconds
Call sync again if localCommitId is not equal to server commit id.
**- How I did it**
Added a condition to see if localCommiId equals serverCommitI

**- How to verify it**
all the functional tests should pass

### DIFF
--- a/tests/at_functional_test/lib/src/sync_service.dart
+++ b/tests/at_functional_test/lib/src/sync_service.dart
@@ -2,7 +2,6 @@ import 'dart:async';
 
 import 'package:at_client/at_client.dart';
 import 'package:at_client/src/service/sync_service.dart';
-import 'package:at_client/src/service/sync_service_impl.dart';
 import 'package:at_utils/at_logger.dart';
 
 /// The class represents the sync services for the functional tests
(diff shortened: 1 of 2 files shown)


tests/at_functional_test/test/atclient_sync_callback_test.dart
```diff
@@ -166,57 +166,26 @@ void main() {
...
-            expect(keyInfo.syncDirection, SyncDirection.remoteToLocal);
-          }
-        });
-      }
-      // At the end of the test (either in first run or by end of second run),
-      // localCommitId should be equal to serverCommitId
-      if (isLocalCommitIdEqualServerCommitId == false) {
-        if (syncProgress.localCommitId == syncProgress.serverCommitId) {
-          isLocalCommitIdEqualServerCommitId = true;
+      expect(syncProgress.keyInfoList, isNotEmpty);
```
sitaram-kalluri: Intermittently, sync progress returns localCommitId and serverCommitId as null in which case the keyInfoList will be empty. Then the "expect" fails. 
Please surround "expect" inside an IF condition as follows:
```dart
if (syncProgress.localCommitIdBeforeSync != syncProgress.localCommitId) {
    expect(syncProgress.keyInfoList, isNotEmpty);
    expect(syncProgress.localCommitId,
          greaterThan(syncProgress.localCommitIdBeforeSync!));
}

**Review by sitaram-kalluri — CHANGES_REQUESTED** [review]

File: tests/at_functional_test/lib/src/sync_service.dart [hunk 2]
```diff
@@ -18,20 +19,39 @@ class FunctionalTestSyncService {
   }
 
   Future<void> syncData(SyncService syncService) async {
-    var isSyncInProgress = true;
-    var functionalTestSyncProgressListener = FunctionalTestSyncProgressListener();
+    // Setting sync request threshold to 1 to expedite sync process
```
sitaram-kalluri, line 23: It would be nice to have "SyncOptions" as an optional parameter to the "syncData" method similar to the way we have in E2E tests. So that we can have tests wait on a specific key to sync before asserting.

sitaram-kalluri (changes requested):

tests/at_functional_test/test/atclient_sync_callback_test.dart
purnimavenkatasubbu: Updated

sitaram-kalluri (approved):

PR #1024, opened 2023-05-10
fix: Sync missing entries when running with multiple clients and add …

### LINKED ISSUE #1016: syncProgress.conflictInfo isn't returning the local and remote value
### Describe the bug

When there is a conflicting key in the local and remote secondaries, syncProgress.conflictInfo returns the localValue and remoteValue as null instead of the correct values

### Steps to reproduce

Run the sync_conflict_test.dart
https://github.com/atsign-foundation/at_client_sdk/commit/d2071f54c48d4720876ce3f8bc474a72ee46af7b

### Expected behavior

remoteValue and localValue should be null

### Screenshots

Screenshot from 2023-05-03 15-14-21

### Smartphones

_No response_

_No response_

### Additional context

_No response_

sitaram-kalluri:
**- What I did**
* **Issue-1:** When multiple clients pump data to the server and attempt to sync the changes, conflicts arise resulting in missing and inconsistencies between clients. Due to conflicts, Client-1 may be missing the commit-Id synced from Client-2 and vice-versa. This occurs because the commit-Id in client-1 is greater than the commit-Id synced from client-2.
  * Please refer to the following [google docs](https://docs.google.com/spreadsheets/d/1JLwe0OupRJSkPC52r9Bo3kshcWF5FRHvLdgHuqj1Jyg/edit?usp=share_link) for more details on the issue.

* **Issue-2:** The ConflictInfo does not set the remote and local value. Following is the git issue: https://github.com/atsign-foundation/at_client_sdk/issues/1016

**- How I did it**
* **For issue 1:** On the client side, currently, we are updating the "lastReceivedServerCommitId" when pushing and pulling the server commit entries. To fix the issue, update the "lastReceivedServerCommitId" only when the client pulls the changes. Because of this fix, "lastReceivedServerCommitId" is not updated on pushing the changes and hence we get synced entries back to the client. If the same exists in the uncommitted entries list the data will be overwritten. To address this issue, on pulling the changes, if an entry with the same key exists in uncommitted entries, skip committing that particular entry to local.
* With new changes, lastReceivedServerCommitId will not be updated when pulling the changes. Hence removed unit tests that assert the same.

* **For Issue 2:** Referring to the "Fix for issue-1", when pulling the changes from the server to the local, if the serverCommitEntry exists in the list of uncommitted entries, then the particular serverCommitEntry is skipped (the entry will not sync to the local secondary), and the lastReceivedServerCommitId is updated. WIth these changes, the control does not go into the section that populates ConflictInfo, because if the serverCommitEntry exists in the uncommitedEntries list, we skip processing the entry. So, refactor the code to populate the conflictInfo and skip syncing the serverCommitEntry to local.

**- How to verify it**
* Updated the unit test to assert the conflict info changes.
* Extended the existing unit test to verify conflictInfo is being set for a shared key 
* Add a new unit test to verify ConflictInfo set errorOrExceptionMessage when exception occurs in populating the ConflictInfo.

* Fix: Missing sync entries when multiple clients pump data concurrently.
* Fix: ConflictInfo does not set the remote and local value.


packages/at_client/lib/src/service/sync_service_impl.dart
```diff
@@ -441,12 +434,34 @@ class SyncServiceImpl implements SyncService, AtSignChangeListener {
         List<dynamic> listOfCommitEntriesToSync =
```
murali-shris: you can rename listOfCommitEntriesToSync to listOfCommitEntriesFromServer

murali-shris (changes requested):

packages/at_client/lib/src/service/sync_service_impl.dart
sitaram-kalluri: Sure Murali, Will make this change.

gkc:
This PR is still marked as Draft, is it ready for review?

sitaram-kalluri:
> This PR is still marked as Draft, is it ready for review?

@gkc: Yes, the PR is ready for review. 

The sync isolate test has pending work. At the moment, the test has only 5 keys to inject. The keys are limited to 5 (for now) to ensure the test runs in GitHub action successfully and then incrementally, scale up the keys to 50.

gkc (changes requested):
In the PR description you mention
> Since the changes in at_server are not yet merged, added a skip tag to avoid test failure.

Instead of a `skip` tag, please instead check the atServer version and require version >= 3.0.32

sitaram-kalluri:
> In the PR description you mention
> 
> > Since the changes in at_server are not yet merged, added a skip tag to avoid test failure.
> 
> Instead of a `skip` tag, please instead check the atServer version and require version >= 3.0.32

Sure, Gary.

sitaram-kalluri:
@gkc: The end-2-end tests are failing due to time-out exception. When i run the test in local, i do not see the timeout exception. My guess is because there are not lot of commit-id's to sync, test is taking more than usual time to sync. I will debug and find way to fix the timeout exception. Meanwhile, can we remove the "tests/at_end2end_test/test/sync_multiple_client_test.dart" from the PR and push the other changes? Once the test is stable, will add it to the e2e test suite.

gkc:
SGTM

packages/at_client/lib/src/service/sync_service_impl.dart
```diff
@@ -478,9 +501,9 @@ class SyncServiceImpl implements SyncService, AtSignChangeListener {
           serverCommitEntry['atKey'],
           SyncDirection.remoteToLocal,
           convertCommitOpSymbolToEnum(serverCommitEntry['operation']));
-      ConflictInfo? conflictInfo =
-          await _checkConflict(serverCommitEntry, uncommittedEntries);
-      keyInfo.conflictInfo = conflictInfo;
+      // ConflictInfo? conflictInfo =
+      //     await _checkConflict(serverCommitEntry, uncommittedEntries);
+      //keyInfo.conflictInfo = conflictInfo;
```
gkc: Why is the above code commented out?

packages/at_client/lib/src/service/sync_service_impl.dart
```diff
@@ -580,17 +587,15 @@ class SyncServiceImpl implements SyncService, AtSignChangeListener {
       if (serverMetaData != null && serverMetaData[IS_ENCRYPTED] == "true") {
         final decryptionManager = AtKeyDecryptionManager(_atClient)
             .get(atKey, _atClient.getCurrentAtSign()!);
-
         // ignore: prefer_typing_uninitialized_variables
         var serverDecryptedValue;
-
         if (serverEncryptedValue != null && serverEncryptedValue.isNotEmpty) {
           serverDecryptedValue =
               await decryptionManager.decrypt(atKey, serverEncryptedValue);
```
gkc: What should happen if decryption happens to fail?

packages/at_client/test/sync_new_test.dart
```diff
@@ -1697,145 +1698,6 @@ void main() {
...
-
-      SyncServiceImpl syncService = await SyncServiceImpl.create(mockAtClient,
-          atClientManager: mockAtClientManager,
-          notificationService: mockNotificationService,
-          remoteSecondary: mockRemoteSecondary) as SyncServiceImpl;
-
-      await syncService.syncInternal(-1, SyncRequest()..result = SyncResult());
-      expect(await syncService.getLastReceivedServerCommitId(), 1);
-    });
-
```
gkc, line 1838: Why were the above tests removed?

gkc:
> > In the PR description you mention
> > > Since the changes in at_server are not yet merged, added a skip tag to avoid test failure.
> > 
> > 
> > Instead of a `skip` tag, please instead check the atServer version and require version >= 3.0.32
> 
> Sure, Gary.

I don't see this in the PR ... ?

gkc (changes requested):

packages/at_client/test/sync_new_test.dart
sitaram-kalluri: Earlier, the "lastReceivedServerCommitId" is updated when sync entries are pushed to server and pulled from server. With the new approach we do not update the lastReceivedServerCommitId when we sync the uncommitted entries to server. Hence removed the code that updates the "lastReceivedServerCommitId" when we push the changes to server. Hence removed the related unit tests also.

sitaram-kalluri:
> > > In the PR description you mention
> > > > Since the changes in at_server are not yet merged, added a skip tag to avoid test failure.
> > > 
> > > 
> > > Instead of a `skip` tag, please instead check the atServer version and require version >= 3.0.32
> > 
> > 
> > Sure, Gary.
> 
> I don't see this in the PR ... ?

Referring to the above discussion, The test is removed from the PR since the test often fails due to TimeOut Exception. Will raise a separate PR once the test is stable.

gkc (dismissed):

sitaram-kalluri:
@gkc: Have updated the PR with the following changes:
* Extend the existing unit test to assert on setting the values into ConflictInfo for a shared key.
* Added a new unit test to assert when exception occurs when decrypting the value in "setConflictInfo" method.

gkc (approved):

murali-shris (approved):

PR #1033, opened 2023-05-15
fix: option to delete local secondary when remote has been reset

### LINKED ISSUE #900: Trying to connect with the wrong atKeys results in a lot of errors

When an atSign has been "reset" and new keys are cut by a client, old clients break as they are using the wrong keys. This is particulary acute with atclients that send notifications first (at_talk as an example).

atclient should delete entire local datastore if it detects that the public key on the server is not the same as the public key in its local datastore

_No response_

### Additional context

_No response_

srieteja:
Closes https://github.com/atsign-foundation/at_client_sdk/issues/900
**- What I did**
- Introduce a method that checks if the remote secondary has been reset; returns true if the remote secondary has been reset.
- Introduce a method to delete local storage.
- Introduce a method to close instances of commitLog and hive in the Storage manager

**- How I did it**
- Fetch AT_ENCRYPTION_PUB_KEY from remote secondary and local secondary and if they are not the same, this is considered a reset.
- Additional method has been introduced [deleteLocalSecondaryStorageWithConsent()] which takes a consent flag and deletes the local files belonging to the current atsign.

**- How to verify it**
- unit tests present in at_client_impl_test.dart

fix: option to delete local secondary when remote has been reset


packages/at_client/lib/src/client/at_client_impl.dart
```diff
@@ -881,6 +881,113 @@ class AtClientImpl implements AtClient, AtSignChangeListener {
...
+    if (userConsentToDeleteLocalSecondaryStorage) {
+      _logger
+          .info('Deleting CommitLog local storage at: $commitLogDirectoryPath');
+      _deleteLocalCommitLog(commitLogDirectoryPath);
+      _logger.info('Deleting hive local storage at: $hiveStorageDirectoryPath');
+      _deleteLocalHiveStorage(hiveStorageDirectoryPath);
+    }
+  }
+
+  void _deleteLocalCommitLog(String commitLogDirectoryPath) {
```
murali-shris: have a single method instead of two methods for deleting commit log and main storage. 
e.g _deleteStorage(String storageDirectory, String storageType)//storageType can be commitlog or main

packages/at_client/lib/src/client/at_client_impl.dart
```diff
@@ -881,6 +881,113 @@ class AtClientImpl implements AtClient, AtSignChangeListener {
     }
   }
 
+  @override
+  Future<void> isSecondaryReset() async {
```
murali-shris: this method should return Future<bool>

srieteja: resolved as part of [6b0efb8](https://github.com/atsign-foundation/at_client_sdk/pull/1033/commits/6b0efb847d7cd4d3e5e080a026622bb83b9402ad)

packages/at_client/lib/src/client/at_client_impl.dart
srieteja: change made as part of [6b0efb8](https://github.com/atsign-foundation/at_client_sdk/pull/1033/commits/6b0efb847d7cd4d3e5e080a026622bb83b9402ad)

packages/at_client/lib/src/client/at_client_impl.dart
```diff
@@ -881,6 +884,112 @@ class AtClientImpl implements AtClient, AtSignChangeListener {
...
+          ?.getEncryptionPublicKey(getCurrentAtSign()!);
+    } on Exception catch (e) {
+      _logger.severe(
+          'Exception caused fetch EncryptionPublicKey from LocalSecondary.'
+          ' Unable to complete reset check | Cause $e');
+      return false;
+    }
+    // Fetch EncryptionPublicKey from RemoteSecondary
+    try {
+      PLookupVerbBuilder plookup = PLookupVerbBuilder()
```
murali-shris, line 989: what is the implication if we directly do an unauth lookup instead of auth plookup?

packages/at_client/pubspec.yaml
```diff
@@ -40,6 +40,14 @@ dependencies:
   meta: ^1.8.0
   version: ^3.0.2
 
+dependency_overrides:
```
murali-shris: can you raise a PR for at_commons change?

srieteja: The at_commons change is irrelevant now. It was to include the AtResetException, which we decided to not use.

packages/at_client/lib/src/client/at_client_impl.dart
srieteja: If we directly do an unauth lookup everything should be fine. The plookup is a backup for when the connection is authenticated(public key lookup on authenticated connections is not feasible from what I have observed)

PR #1045, opened 2023-05-25
fix: Incorrect commitId gets updated against commit entry when a batc…

sitaram-kalluri:
**- What I did**
* In the uncommitted entries list, when preparing the Sync batch request,  if there exist two entries of the same key where the initial commit entry is "update" and later is "delete", then the commit entry with "update" operation is skipped adding to the Batch Request. In this case, the size of the batch request is one less than the size of uncommitted entries. 

* On receiving the batch response, the uncommitted list is iterated and the commit id from the batch response is updated to its respective commit entry. Since an entry is skipped in the batch request (but the entry exists in the uncommitted entries list), the commit id gets updated to the skipped entry (here, the entry with CommitOp. update). The commit-id of the entry synced to the server will remain null (here, the entry with CommitOp.Delete will not be updated).

**- How I did it**

* In the "getBatchRequests" method, introduce a temporary list - "removeUncommittedEntriesList". When preparing the Sync batch request, if a key is skipped, then add it to "removeUncommittedEntriesList". In the end, iterate "removeUncommittedEntriesList" and remove the entries from the uncommitted entries list. Also, to avoid stale commit-entry in the commit log keystore, remove the entry from the keystore.

**- How to verify it**
* Added a unit test that asserts the following:
  * In an uncommitted entries list, if the same key is updated and deleted, the commit id should be updated against deleted entry.

* Added a functional test that asserts the same as above.

* Incorrect commit-Id gets updated against commit entry when a batch skips an entry

### DIFF
--- a/packages/at_client/CHANGELOG.md
+++ b/packages/at_client/CHANGELOG.md
@@ -5,6 +5,7 @@
   - Add new optional named parameter `deleteRequestOptions` to AtClient.delete
   - Add `useRemoteAtServer` to DeleteRequestOptions. When set, the delete
     request will be sent directly to the remote atServer
+- fix: Incorrect commitId gets updated against commit entry when a sync-batch skips an entry
 ## 3.0.59
 - fix: Sync running into infinite loop when an invalid key is present in the entries to sync into client
 - fix: Redundant logs generated for an internal key[lastReceivedNotification] while sending notifications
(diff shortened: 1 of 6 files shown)


gkc:
LGTM

gkc (approved):

PR #1054, opened 2023-06-05
feat: at_client offline access

### LINKED ISSUE #915: Being able to get an AtClient without having connectivity to your atServer (i.e offline access)

The problem could be that you have an application that has data which has been synched to your device but now the device is offline. With the current SDK although the data is available locally you cannot get to it as you are offline.

 This kills a number of use case for instance atBuzz only works if you have an Internet connection.

If the atSign has already synched at least once then return an atClient that can read/write locally and when the atServer is not reachable. 

none

### Additional context

This is important for both flutter and dart apps

sitaram-kalluri:
**- What I did**
* Enable at_client to usage when the device is offline.
* AtClient can be used in offline mode only if the atSign is paired with the device.

**- How I did it**
* In at_client_impl, Initialise the storage and verify if the encryption public key is available in the storage. This is to ensure that the atSign is paired with the device.
* If the above returns true, allow the client to use at_client in offline mode.

* Enable at_client usage when device is in offline mode.

### DIFF
--- a/packages/at_client/lib/src/client/at_client_impl.dart
+++ b/packages/at_client/lib/src/client/at_client_impl.dart
@@ -882,6 +882,34 @@ class AtClientImpl implements AtClient, AtSignChangeListener {
     }
   }
 
+  /// Initializes the Hive Keystore and fetches encryption public key to ensure
+  /// the atSign is previously onboarded.
+  /// If the key is available, returns false
+  static Future<bool> verifyStorageForOfflineAccess(
+      String atSign, AtClientPreference atClientPreference) async {
+    SecondaryPersistenceStore secondaryPersistenceStore =
+        SecondaryPersistenceStoreFactory.getInstance()
+            .getSecondaryPersistenceStore(atSign)!;
+
+    // Hive box might close when app is killed or device is restarted
+    // Hence initializing the hive box.
+    await secondaryPersistenceStore
+        .getHivePersistenceManager()!
+        .init(atClientPreference.hiveStoragePath!);
+
+    var hiveKeyStore = secondaryPersistenceStore.getSecondaryKeyStore()!;
+    AtData? atData;
+    try {
+      atData = await hiveKeyStore.get('$AT_ENCRYPTION_PUBLIC_KEY$atSign');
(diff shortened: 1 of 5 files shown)


packages/at_client/lib/src/client/at_client_impl.dart
```diff
@@ -882,6 +882,34 @@ class AtClientImpl implements AtClient, AtSignChangeListener {
     }
   }
 
+  /// Initializes the Hive Keystore and fetches encryption public key to ensure
+  /// the atSign is previously onboarded.
+  /// If the key is available, returns false
+  static Future<bool> verifyStorageForOfflineAccess(
```
gkc: Maybe rename to 'canUseOffline'

gkc: I'd prefer
1. Instance method rather than static method
2. Method added to the at_client spec
3. Clarify expected usage - for example first I create an atClient, and then I check if it can be used offline

gkc (changes requested):

packages/at_client/lib/src/client/at_client_impl.dart
sitaram-kalluri: @gkc :
 1. Renamed the method and removed "static" keyword.
 2. The method is added to the at_client_spec. 
 3. The current behaviour is, the app checks if the device is offline. If yes invokes ''isAtClientReadyForOfflineAccess" which initializes the "AtClient" and verifies if the encryption public key is available in the KeyStore.

gkc: What is the developer experience we want here? Can you link the ticket here please and then if not already done, in the ticket lay out the problem we want to solve and then think about what is the lowest effort way for a developer to avoid / handle that problem.

gkc:
Have added a comment to the [ticket](https://github.com/atsign-foundation/at_client_sdk/issues/915)

gkc (changes requested):
See comment on ticket #915

sitaram-kalluri:
The changes are not valid. We have a different approach in addressing the issue. Hence closing this PR.

Issue #1071, opened 2023-06-20
Invalid or corrupted pad block when using AtClient.get


Xlin123:
### Describe the bug

Using AtClient.get with shared symmetric keys will throw an invalid pad block exception.

This [issue](https://github.com/atsign-foundation/at_client_sdk/issues/1057) fixed the sync_service, changes for this will most likely need to be done in decryption_service.

### Steps to reproduce

1. Using Dart REPL, make sure to add dependency override for at_client using git
2.  use /get on a shared symmetric key
3. error

### Expected behavior

@gkc mentioned that he wanted at least a better exception but ideally the better solution would be the decrypted value for “our” copy, or an exception for “their” copy.

1.  alice retrieving their copy of the symmetric key that alice created for bob` /get shared_key.bob@alice` ==> should be decryptable by alice
2. alice retrieving bob's copy of the symmetric key that alice created for bob `/get @bob:shared_key@alice `==> not decryptable by alice, as alice encrypted this with bob's public key, and alice does not have bob's private key
3.  alice retrieving their copy of the symmetric key that bob created for alice` /get @alice:shared_key@bob `==> should be decryptable by alice
4. alice retrieving bob's copy of the symmetric key that bob created for alice ==> this is not actually possible, it is only ever visible to bob in bob's atServer

### Screenshots

![image](https://github.com/atsign-foundation/at_client_sdk/assets/42496953/48e54d7b-696c-417d-90a2-42ae1ab70de3)

Xlin123:
bug is fixed as of now, just need to write tests.

PR #1072, opened 2023-06-20
fix: added a check to handle self symmetric shared

Xlin123:
**- What I did**
I fixed bug with AtClient.get and shared symmetric keys #1071 
**- How I did it**
Added symmetric shared key specific exceptions and also made sure they used proper encryption type on the keys.
**- How to verify it**
A made a test in decryption_service_test, I'm not sure if I did it right or if it even ran in the suite. 
Fixed bug with AtClient.get and shared symmetric keys


Xlin123:
conclusion: to fix the issue I need to use at_chops to do RSA encryption because it's trying to use the private symmetric key to decrypt an AES key. If there's any place where at_chops is used so I can see an example it would be very helpful.

gkc:
See here to understand what happens when, as `@bob`, you decrypt a message from `@alice` https://github.com/atsign-foundation/at_protocol/blob/trunk/usage-examples/how-to-exchange-encrypted-data.md#4-fetch-the-symmetric-key-which-alice-shared

The code which handles this is in shared_key_decryption.dart, and part of that code handles the fetching of the symmetric key, decoding from base64, and decrypting using at_chops and the 'receiving' atSign's  private key

Xlin123:
Ahhh, okay got it. I did some work on self_key_decryption first since keys like "shared_key.bob@alice" I think fall under self keys. Let me know if that's incorrect and I can just revert changes. I'll do shared_key_decryption when I get home.

packages/at_client/lib/src/decryption_service/self_key_decryption.dart
```diff
@@ -19,7 +20,16 @@ class SelfKeyDecryption implements AtKeyDecryption {
           intent: Intent.decryptData,
           exceptionScenario: ExceptionScenario.decryptionFailed);
     }
-
+    if (atKey.key == "shared_key") {
```
gkc, line 23: The copy alice will create for themselves of a shared key they cut for bob will be `shared_key.bob@alice`. Will this test work? What happens when you use this branch with your REPL?

packages/at_client/lib/src/decryption_service/self_key_decryption.dart
```diff
@@ -19,7 +20,16 @@ class SelfKeyDecryption implements AtKeyDecryption {
...
+      if (atKey.sharedWith != _atClient.getCurrentAtSign()) {
```
gkc: sharedWith is not set - when we create `shared_key.bob@alice`  - again, what happens when you run this branch with your REPL? Do you get the desired outcomes?

gkc (changes requested):

Xlin123:
Two decryptable cases
image

However, ` /get @bob:shared_key@alice ` isn't throwing it's exception properly. I should finish it tonight.

Xlin123:
image
@gkc   I think I fixed everything.

gkc:
Please add unit tests which verify each case

PR #1073, opened 2023-06-21
fix: Add a new functional test and refactor the existing functional tests

### LINKED ISSUE #1011: Sync test with multiple clients pumping data.
Test:
-------
* Write a system test case that injects the data from multiple clients concurrently. 
Assertions:
* After pumping the data allow the clients and the server to be in sync. 
* Then for each key injecting, ensure the commit-Id between the server and both clients is the same.

- [x] Write a test with concurrently pumps data into the server
- [x] Enhance stats verb to get the latest commit-Id of each key from the server

sitaram-kalluri:
**- What I did**

- Add a new functional test - ''A test to verify the commit log entries when keys are synced from multiple clients".  The tests starts the two clients in the isolates to and runs put and delete operations concurrently. At the end, fetches the commit log of both the client and server and compares the commit-id's and commit-op's of the keys.

- Refactored the existing functional tests as follows:
   - The code that initialises the AtClientManagers and sets the encryptions keys into the local key-store is executed in each test which is not necessary. Moved the piece of code into setUpAll method that runs before the starts of all the tests in the given test file.
   - In the tearDown method, the key-stores files are removed at the end of each test which causes sync to run at the start of subsequent test which is time consuming. Hence removed the tearDown method from all the tests.
   - At the moment, the setEncryptionKeys method inserts the encryption keys into the local keystore at the start of each test even if the keys are already present if the keystore. Enhanced code such that, verifies if the encryption keys are present in the local keystore and inserts only keys are not present, else returns. Also moved "setEncryptionsKeys" method to "/lib/src" and renamed as "AtEncryptionKeysLoader"
   - Replaced the "sync" calls and "Future.delayed" that waits until the sync is completed in tests with "FunctionalTestSyncService.syncData" that ensures the client and server are in sync
   - Removed the "getPreferences" in that are redundant in each test. Every test fetches the preferences from the "TestUtils.getPreferenes"
   - Updated the "test" package version to "1.24.3"
   - The tests in at_client_sync_conflict_test have assertions outside the test method. Hence the test will be marked "passed" even when an exception or failure occurs. Also, observed the following exception in the test log
   `SEVERE|2023-06-21 13:02:25.277937|SyncService (@alice🛠)|unable to inform sync progress to listener Instance of 'MySyncProgressListener'. Reason: Expected: <true>
  Actual: <false>`
   - Refactored the tests in at_client_sync_conflict_test.dart to have assertions inside the "test" method.

* Replaced the localCommitId with lastReceivedServerCommitId in the statsNotificationListener.
  - The "IsInSync" method is modified to use the "lastReceivedServerCommitId" instead of the "localCommitId" for synchronization between the client and server.
  - When changes are pulled from the server, the "localCommitId" is incremented, while the "lastReceivedServerCommitId" remains unchanged. For example, let's assume the "serverCommitId" is at 10, after pulling the changes from server, the "localCommitId" arrives to 10, but the "lastReceivedServerCommitId" will be at 6. In this situation, the "IsInSync" method would return false. However, in the statsNotificationListener, since the localCommitId and ServerCommitId are equal, the sync request is not added to the syncRequestQueue, leading to sync not being triggered. To fix this, replaced the "localCommitId" with "lastReceivedServerCommitId".


purnimavenkatasubbu (dismissed):

packages/at_client/lib/src/service/sync_service_impl.dart
```diff
@@ -154,7 +154,7 @@ class SyncServiceImpl implements SyncService, AtSignChangeListener {
       final serverCommitId = notification.value;
       int localCommitId = -1;
       try {
-        localCommitId = await _getLocalCommitId();
+        localCommitId = await getLastReceivedServerCommitId();
```
gkc: Please add an explanation of this change to the PR description. Is there a test (of any kind - unit test, functional test or end-to-end test) which checks that this code is working correctly?

gkc (changes requested):

packages/at_client/lib/src/service/sync_service_impl.dart
gkc: Thanks for updating the description. Is there a test which will fail if this change is reverted?

gkc: > Thanks for updating the description. Is there a test which will fail if this change is reverted?

@sitaram-kalluri ^^

sitaram-kalluri: @gkc : Yes added a unit test too in the followng commit. https://github.com/atsign-foundation/at_client_sdk/pull/1073/commits/24100f61e445db64f7e361a664e3ee91e00c4dff#diff-cc97e657d252f9b053e893ca0682e931ff543e6aa9969febb9a7e6e75321955f

gkc (approved):
LGTM

gkc:
@VJag please can you review this also?

packages/at_client/lib/src/service/sync_service_impl.dart
```diff
@@ -169,7 +170,8 @@ class SyncServiceImpl implements SyncService, AtSignChangeListener {
           ..syncStatus = SyncStatus.failure;
         _informSyncProgress(syncProgress);
       }
-      if (serverCommitId != null && int.parse(serverCommitId) > localCommitId) {
+      if (serverCommitId != null &&
```
VJag, line 173: Can we wrap this into a function like "isServerHasNewChanges(<args>)? You may skip it if you are happy with what we have.

tests/at_functional_test/test/atclient_put_test.dart
```diff
@@ -2,171 +2,123 @@ import 'dart:io';
...
     // phone.me@alice🛠
     var phoneKey = AtKey()
       ..key = 'phone'
       ..sharedWith = '@bob🛠';
     var value = '+1 100 200 300';
-    var putResult = await atClient.put(phoneKey, value);
+    var putResult = await atClientManager.atClient.put(phoneKey, value);
     expect(putResult, true);
-    var getResult = await atClient.get(phoneKey);
+    var getResult = await atClientManager.atClient.get(phoneKey);
```
VJag, line 32: For tests like this, the sync is happening as a side effect. It would be great if we can prevent it so that tests will be faster. Options could be:

1. Write key to key store but not commit log
2. Test using unsynced keys
3. Futuristic: shouldSync can be an attribute on probably metadata where setting it to false will prevent a sync and it can be defaulted to true

VJag (approved):
LGTM

PR #1103, opened 2023-08-03
fix: collection_test.dart - Add syncOption to wait until the cached k…

sitaram-kalluri:
**- What I did**
- Fix the failing test - "Model operations - save() with reshare() as true test"
- In the earlier fix, added syncOptions on the sender client which ensure the key is synced from the sender's local secondary to the sender's remote secondary. 
- The issue now occurs due to the cached key not getting synced from the receiver's remote secondary to the receiver's local secondary. Added syncOptions on the receiver's side as well to wait until the cached key sync from the receiver's remote secondary to the local secondary.
- Further, temporarily down the log level to "finest" for this test and added a print statement that prints the response on the receiver's side, to get more information - just incase if the test fails again.

**- How to verify it**
- The failing test should pass.

### DIFF
--- a/tests/at_end2end_test/test/collection_test.dart
+++ b/tests/at_end2end_test/test/collection_test.dart
@@ -9,7 +9,6 @@ import 'package:at_end2end_test/src/sync_initializer.dart';
 import 'package:at_end2end_test/src/test_initializers.dart';
 import 'package:at_end2end_test/src/test_preferences.dart';
 import 'package:test/test.dart';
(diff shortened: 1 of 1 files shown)


gkc (changes requested):
LGTM, please remove the setting of log level to 'finest'

sitaram-kalluri:
> LGTM, please remove the setting of log level to 'finest'

Removed the log level

gkc (approved):

PR #1135, opened 2023-10-11
fix: make namespace for local keys NOT mandatory

### LINKED ISSUE #1075: at_client expects internal keys to have namespace (when namespace is not provided through preferences)
### Describe the bug

some atsigns after being activated with activate_cli/onboarding_cli end up in a situation where sync keeps failing due to an exception 
````
SEVERE|2023-06-26 13:34:51.888723|SyncService (@34thundering)|Exception in sync cb049548-7780-4cf0-95c1-7bde8faa1d93. Reason: Failed to syncData caused by namespace is mandatory 
````

When a namespace is not set through at_client preferences, at_client while creating internal keys (e.g. lastReceivedCommitId, lastReceivedNotification) expects a namespace explicitly defined for these keys. And since these keys do not have a specific namespace, the client throws an exception which terminates the client.

### Steps to reproduce

1. First I 'activate a new atsign with activate_cli'
2. Then I 'use that atsign with onboaring_cli'
3. And then 'insert a key into the local secondary'
4. This results in a Sync Failure and the syncs keep failing forever

### Expected behavior

Sync should succeed and not throw any exceptions

### Possible solution

1. exclude internal keys from the namespace scrutiny (possibly in the validatePutRequest() in at_client_impl)
2. use a constant/predefined namespace for all internal keys

AtOnbaordingCli

### PR DESCRIPTION
**Pull request by srieteja**
Closes https://github.com/atsign-foundation/at_client_sdk/issues/1075

**- What I did**
- make namespace not mandatory for local keys

**- How I did it**
- exclude local keys from namespace verification which takes place in AtClientValidation.validatePutRequest()
 
**- How to verify it**
- unit tests added for public, private and shared keys asserting that namespace is required for these keys
- unit test added for local keys asserting that namespace is not mandatory

fix: make namespace for local keys NOT mandatory

### CODE REVIEW

**Comment by gkc** [comment]
@srieteja why is this necessary?

**Comment by srieteja** [comment]
> @srieteja why is this necessary?

@gkc There was an issue a couple of months ago where the lastReceivedServerCommitId key was not being created if namespace is not set in at_client_preferences(no namespace is set while creating the key). We discussed this on the arch call and concluded that we should not enforce namespace validation on local keys to resolve this issue. I think I remember you were a part of the discussion.

more details at: https://github.com/atsign-foundation/at_client_sdk/issues/1075

**Comment by gkc** [comment]
Got it - thanks.

File: packages/at_client/lib/src/util/at_client_validation.dart [hunk 1]
```diff
@@ -97,14 +97,15 @@ class AtClientValidation {
     }
     // If key starts with 'cached:' (or) if the metadata of key has isCached set to true; it represent
     // a cached key. Prevent client updating the cached key. Hence throw exception.
-    if (atKey.key!.startsWith('$CACHED:') ||
+    if (atKey.key!.startsWith('${AtConstants.cached}:') ||
         (atKey.metadata != null && atKey.metadata!.isCached)) {
       throw AtKeyException('Cannot create/update a cached key');
     }
     // If namespace is not set on key and in preferences, throw exception
     if ((atKey.namespace == null || atKey.namespace!.isEmpty) &&
```
murali-shris, line 105: change this condition to
if (((atKey.namespace == null || atKey.namespace!.isEmpty &&) !atKey.isLocal) &&
        (atClientPreference.namespace == null ||
            atClientPreference.namespace!.isEmpty)
        )

murali-shris:
make the version change in pubspec and AtClientConfig

murali-shris (changes requested):

packages/at_client/lib/src/util/at_client_validation.dart
srieteja: change made as part of [27600db](https://github.com/atsign-foundation/at_client_sdk/pull/1135/commits/27600dbb07a1f422a7e254200b1007e97f54567d)

srieteja:
> make the version change in pubspec and AtClientConfig

The version has been updated to 3.0.66 in pubspec and at_client_config

murali-shris (approved):

sitaram-kalluri (approved):

PR #1142, opened 2023-10-16
feat: remove useAtChops references in code

murali-shris:
**- What I did**
- changes to enforce atClient to use atChops for authentication 
- remove the references to AtClientPreference.privateKey in tests
- remove the use of EncryptionUtil in few classes which was used as fallback if useAtChops=false
**- How I did it**
- added a method _createAtChops in AtClientImpl. If atChops is not injected from at_client_mobile/at_onboarding_cli, then keys will be read from local secondary and atChops instance will be created.
- removed use of EncryptionUtil in shared_key_decryption.dart, abstract_atkey_encryption.dart
- removed code which reads from private key in monitor.dart
- deprecated useAtChops experimental flag in AtClientPreference. Removed all references in at_client
- removed call to SignInPublicData.signData in put_request_transformer.dart
- refactored unit tests to remove references to useAtChops and mocked atChops where required
- removed references to AtClientPreference.privateKey in tests
- added a method in functional tests  TestUtils.initAtClient(atSign, namespace); This will create at_chops instance and set it in the call to AtClientManager.setCurrentAtSign and also removes duplicate code in setUpAll methods in functional tests
**- How to verify it**
- unit and functional tests should pass
- sanity in apps/cli with this branch of at_client. 
@purnimavenkatasubbu  Please take this up.


gkc (approved):
LGTM - thank you @murali-shris

PR #1153, opened 2023-10-29
feat: make AtRpc use ephemeral notifications

gkc:
**- What I did**
feat: make at rpc use ephemeral notifications

**- How I did it**
See commits

**- How to verify it**
Tests pass


cconstab (approved):
Checks passed and walked through with gkc

PR #1161, opened 2023-11-09
fix: compare the complete atKey to check for a reserved key

purnimavenkatasubbu:
This PR contains changes required by at_client to match the new reserved key regex in https://github.com/atsign-foundation/at_libraries/pull/445

**- What I did**
- Compare the entire key to perform a reserved key check in NotificationResponseTransformer :
  > Previously only the key part of the atkey was compared for the reserved key check. With the updates to the reserved key regex this check will fail. To ensure this does not happen the entire atkey(including sharedWith and sharedBy) is compared with the reserved key regex now.

**- How I did it**
- Pass atKey.toString() to the AtKey.getType() call in NotificationResponseTranformer(previously atKey.key was used)

**- How to verify it**
- All existing tests should pass

compare the complete atKey to check for a reserved key


packages/at_client/pubspec.yaml
```diff
@@ -40,10 +40,17 @@ dependencies:
   meta: ^1.8.0
   version: ^3.0.2
 
+dependency_overrides:
```
sitaram-kalluri: If the changes in at_commons are published, please remove the dependency_overrides and updated the hosted version

sitaram-kalluri (changes requested):

packages/at_client/pubspec.yaml
srieteja: sure @sitaram-kalluri. Will do that as soon as at_commons is published

srieteja:
closing this PR as another PR contains these changes https://github.com/atsign-foundation/at_client_sdk/pull/1179

PR #1169, opened 2023-11-20
chore: Remove python as it's no longer used for dependency override

cpswan:
Fixes https://github.com/atsign-foundation/at_client_sdk/security/dependabot/4

**- What I did**

Removed all Python components

**- How I did it**

```
rm tools/pyproject.toml tools/update_requirements.sh tools/requirements.txt .github/workflows/update_python_requirements.yml
```

**- How to verify it**

Dependabot alert will clear

chore: Remove python as it's no longer used for dependency override


gkc (approved):
Yay!

PR #1187, opened 2023-12-20
feat: schedule job that removes expired keys

### LINKED ISSUE #1660: Improve commit log handling for expired (`expiresAt` has passed) records

Summary
- We are managing deletions of expired records on the client side via server creating delete entries in the commit log, and the sync process. This is unnecessary as clients also have the ability to detect expired records and delete them.
- As a result, the commit log grows every time a record expires. So if you create a million records over a period of a week, all of which expire within an hour, you will have a commit log with a million entries in it, which will stay there forever

Detail
- `@alice` sends notification to `@bob` with ttr:-1 (cache indefinitely on bob's atServer) and ttl:3600000 (1 hour)
  - or does a put with ttr:-1 and ttl:36000000
- when bob's (and alice's, in the case of a `put` rather than a `notify`) atServers finds the expired records during a sweep, currently a record is added to the commit log
- as a result, the commit log grows every time a record with a new key is shared / notified
- this isn't necessary, since client-side storage can detect locally stored (e.g. by sync) expired records and delete them

- client-side, when a record is fetched, if it has expired then it should be deleted (and not added into local uncommitted entries)
- client-side, there should be regular background sweeps for expired records - or, at a minimum, at startup time (i.e. creation of LocalSecondary object)
- server-side, when a record is fetched, if it has expired then it should be deleted, and the existing commit entry should be deleted, and no new commit entry should be added

_No response_

### Additional context

_No response_

srieteja:
closes https://github.com/atsign-foundation/at_server/issues/1660
**- What I did**
- with changes in https://github.com/atsign-foundation/at_server/pull/1713, commit entries of delete entries for expired keys will NOT be added to the commitLog and hence will not be synced. Hence, this requires AtClient to have its own job to remove expired keys.
- To solve the abovementioned situation, this PR introduces a scheduled job to remove expired keys.

**- How I did it**
- introduced new param in AtClientPrefs 'expiryCheckTimeIntervalMins' which defines the interval at which the delete expired keys job should run. Defaults to 10 min
- schedule the deleteExpiredKeys job while creating the persistence instances [in StorageManager]

**- How to verify it**
- manual testing has been done for now to verify the updated behaviour
- unit tests present to validate the changes

feat: schedule job that removes expired keys


gkc (changes requested):
LGTM but will not approve for now, given that you are still
> evaluating the possibility of writing unit tests

srieteja:
This PR needs https://github.com/atsign-foundation/at_server/pull/1713 merged and at_persistence_secondary published. Will wait unit that is done.

srieteja:
Added unit tests. Consumed latest version of at_persistence_secondary_server v3.0.61. Ready for review

packages/at_client/lib/src/preference/at_client_preference.dart
```diff
@@ -66,6 +66,9 @@ class AtClientPreference {
 
   Duration monitorHeartbeatInterval = Duration(seconds: 10);
 
+  /// Time interval for the scheduled task that removes expired keys from local keyStore
+  int expiryCheckTimeIntervalMins = 10;
```
gkc: Can you make this a Duration please, allows for more flexibility

gkc (changes requested):

packages/at_client/lib/src/preference/at_client_preference.dart
srieteja: chnaged as part of https://github.com/atsign-foundation/at_client_sdk/pull/1187/commits/bf650bef4d6bd58f36e67e0d55ca5bf26fe95e5b

gkc (commented):
LGTM

srieteja:
The unit tests have been updated. Ready to be reviewed/approved

murali-shris (approved):

gkc (approved):
lgtm

PR #1223, opened 2024-01-25
feat: Introduce methods to set SPP and fetch OTP from the secondary server

### LINKED ISSUE #1219: Introduce methods to set SPP and fetch OTP in at_client SDK
Introduce two new methods to set Semi Permanent Pass-code (SPP) and fetch the OTP  from the server in at_client SDK

sitaram-kalluri:
**- What I did**
- Introduce "setSPP" and "getOTP" methods in at_client_spec to set the semi permanent pass-code and fetch the OTP from the secondary server.
- Introduce "enrollmentId" as optional parameter to AtClient.create method and AtClient._init method to set the enrollment Id which will be propagated to the AtLookup instance used for APKAM authentication.
- Update the at_demo_data from v1.0.1 to v1.0.3 which has RSA Key-pair added APKAM.
- In the "at_keys_initializer.dart" file, eliminate the dependency on "at_demo_credentials.dart," as it duplicates the content found in the "at_demo_data.dart" file.
- Update the test condition in the enrollment_test.dart - "validate client functionality to fetch pending enrollments on legacy pkam authenticated client"
  The initial condition was 
```
    expect(enrollmentRequests.length, 4);
   // 4 entries - 2 entries from this test
   // + 2 entries from the other test in this file.
```
With this condition, there might be intermitent failure depending on the order of the test run. Hence the condition as follows - `expect(enrollmentRequests.length > 2, true);`

**- How I did it**
- The setSPP method accepts a String which represents an SPP. Performs validations and if successful, sends the SPP to secondary server via the executeCommand, else returns an InvalidPINException.
- The getOTP method fetches the OTP from the secondary server and returns to the client.

**- How to verify it**
- Added unit tests to assert on the negative scenarios that returns InvalidPINException for an Invalid SPP supplied.
- Added functional test to create a new SPP and submit and enrollment request with the SPP.

- Introduce methods to set SPP and fetch OTP from the secondary server

### DIFF
--- a/packages/at_client/lib/src/client/at_client_impl.dart
+++ b/packages/at_client/lib/src/client/at_client_impl.dart
@@ -145,9 +145,10 @@ class AtClientImpl implements AtClient, AtSignChangeListener {
           localSecondaryKeyStore: localSecondaryKeyStore,
           atChops: atChops,
           atClientCommitLogCompaction: atClientCommitLogCompaction,
-          atClientConfig: atClientConfig);
+          atClientConfig: atClientConfig,
+          enrollmentId: enrollmentId);
 
-      await atClientImpl._init(enrollmentId: enrollmentId);
+      await atClientImpl._init();
     }
 
     await atClientImpl!._startCompactionJob();
@@ -164,7 +165,8 @@ class AtClientImpl implements AtClient, AtSignChangeListener {
       SecondaryKeyStore? localSecondaryKeyStore,
       AtChops? atChops,
       AtClientCommitLogCompaction? atClientCommitLogCompaction,
-      AtClientConfig? atClientConfig}) {
+      AtClientConfig? atClientConfig,
+      this.enrollmentId}) {
     _atSign = AtUtils.fixAtSign(theAtSign);
     _logger = AtSignLogger('AtClientImpl ($_atSign)');
     _preference = preference;
@@ -182,7 +184,7 @@ class AtClientImpl implements AtClient, AtSignChangeListener {
     _atClientCommitLogCompaction = atClientCommitLogCompaction;
   }
 
-  Future<void> _init({String? enrollmentId}) async {
+  Future<void> _init() async {
     if (_preference!.isLocalStoreRequired) {
       if (_localSecondaryKeyStore == null) {
         var storageManager = StorageManager(preference);
@@ -193,8 +195,6 @@ class AtClientImpl implements AtClient, AtSignChangeListener {
       _atChops ??= await _createAtChops(_atSign);
     }
 
-    // Set enrollment id
-    this.enrollmentId ??= enrollmentId;
     // Now using ??= because we may be injecting a RemoteSecondary
     _remoteSecondary ??= RemoteSecondary(_atSign, _preference!,
         atChops: atChops,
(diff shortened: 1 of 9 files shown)


gkc (commented):
LGTM - I won't approve, I'll let @murali-shris approve once he has reviewed

packages/at_client/lib/src/client/at_client_spec.dart
```diff
@@ -534,6 +537,28 @@ abstract class AtClient {
       Function streamCompletionCallBack,
       Function streamReceiveCallBack);
 
+  /// Sets a Semi Permanent Passcode(SPP) in the secondary server key-store.
+  /// A Semi Permanent Passcode (SPP) is 6 character alpha-numeric for submitting
+  /// an enrollment request. Only the connections which have access to manage
+  /// namespace are allowed to set SPP
+  ///
+  /// Throws [InvalidPinException] when an invalid SPP is provided.
+  ///
```
murali-shris, line 549: document the exception when an unauthorized connection calls setSPP

tests/at_functional_test/pubspec.yaml
```diff
@@ -8,11 +8,16 @@ environment:
 
 dependencies:
   uuid: 3.0.7
+  at_demo_data:
```
murali-shris: why can't we use published at_demo_data?

murali-shris (changes requested):

tests/at_functional_test/pubspec.yaml
sitaram-kalluri: Sure Murali, will publish the at_demo_data and update the version here.

packages/at_client/lib/src/client/at_client_impl.dart
```diff
@@ -192,6 +193,8 @@ class AtClientImpl implements AtClient, AtSignChangeListener {
       _atChops ??= await _createAtChops(_atSign);
     }
 
+    // Set enrollment id
+    this.enrollmentId ??= enrollmentId;
```
murali-shris: have this line in AtClientImpl._ instead of _init(). you can remove the param to _init

packages/at_client/lib/src/client/at_client_impl.dart
```diff
@@ -907,6 +910,48 @@ class AtClientImpl implements AtClient, AtSignChangeListener {
...
+    bool hasMatch = RegExp(r'[\W-]+').hasMatch(spp);
+    if (hasMatch) {
+      throw InvalidPinException.message("$spp is not a valid SPP");
+    }
+    String? otpVerbResponse;
+    try {
+      otpVerbResponse =
+          await _remoteSecondary?.executeCommand('otp:put:$spp\n', auth: true);
+    } on AtLookUpException catch (e) {
+      throw InvalidPinException(e.errorCode, e.errorMessage);
```
murali-shris: we have to check the error code and then throw InvalidPinException. It can be any AtLookUpException here

murali-shris (changes requested):

packages/at_client/lib/src/client/at_client_impl.dart
sitaram-kalluri: This is addressed in [38d3293](https://github.com/atsign-foundation/at_client_sdk/pull/1223/commits/38d3293920af315d493a991a59b61eb49e201704)

packages/at_client/lib/src/client/at_client_impl.dart
sitaram-kalluri: Added validations for the SPP before the invocation of remoteSecondary.executeCommand. So for any invalid pin, the exception will be returned from at_client itself (Calls will not be send to secondary server). Hence converting any exception from secondary server as `AtClientException`

murali-shris (approved):

PR #1225, opened 2024-01-30
fix: Add EnrollmentStatusCallback function for receiving status callback

sitaram-kalluri:
**- What I did**
- Introduce enrollmentCallback to return the enrollment status to client when an enrollment is approved or denied.


gkc (changes requested):
Use a Future rather than a callback please.

PS This looks like an enhancement (feat:) not a fix

sitaram-kalluri:
> Use a Future rather than a callback please.
> 
> PS This looks like an enhancement (feat:) not a fix

Sure Gary. Will replace the callback with Future.

PR #1262, opened 2024-03-18
feat: consume changes reg filtering of enrollment requests through enroll:list

srieteja:
**- What I did**
- introduce new param 'enrollmentStatusFilter' in EnrollListRequestParam class
- use the above param in AtClientImpl.fetchEnrollmentRequests() to build the command that lists only enrollment requests that match the filter.

**- How I did it**
- EnrollListRequestParam.enrollmentStatusFilter will be used by users to pass the desired enrollment statuses they wish to list.
- AtClientImpl.fetchEnrollmentRequests() will use this filter to construct a relevant command which will be passed to the server.

**- How to verify it**
- unit test added, that verifies that appropriate command is generated using the EnrollVerbBuilder and the parsing of response from server.
- Yet to evaluate if a functional test is needed to validate the change

feat: consume changes reg filtering of enrollment requests through enroll:list

### DIFF
--- a/packages/at_client/lib/src/client/at_client_impl.dart
+++ b/packages/at_client/lib/src/client/at_client_impl.dart
@@ -665,15 +665,15 @@ class AtClientImpl implements AtClient, AtSignChangeListener {
 
   @override
   Future<List<EnrollmentRequest>> fetchEnrollmentRequests(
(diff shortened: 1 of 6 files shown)


srieteja:
changes moved to https://github.com/atsign-foundation/at_client_sdk/pull/1263

PR #1263, opened 2024-03-19
feat: enhance EnrollmentService.fetchEnrollmentRequests() to filter enrollment requests

srieteja:
**- What I did**
- introduce new param 'enrollmentStatusFilter' in EnrollListRequestParam class
- use the above param in EnrollmentService.fetchEnrollmentRequests() to build the command that lists only enrollment requests that match the filter.

**- How I did it**
- EnrollListRequestParam.enrollmentStatusFilter will be used by users to pass the desired enrollment statuses they wish to list.
- EnrollmentService.fetchEnrollmentRequests() will use this filter to construct a relevant command which will be passed to the server.

**- How to verify it**
- unit test added, that verifies that appropriate command is generated using the EnrollVerbBuilder and the parsing of response from the server.
- Yet to evaluate if a functional test is needed to validate the change

feat: enhance EnrollmentService.fetchEnrollmentRequests() to filter enrollment requests


srieteja:
closing this PR due to incorrect origin

PR #1295, opened 2024-04-12
fix: Store auth keys into keychain upon enrollment approval

sitaram-kalluri:
**- What I did**
- Store the Enrollment auth keys into keychain manager upon enrollment approval.
 - Earlier, upon approving the enrollment, the auth keys are encrypted and stored into a atKeys text file. Change the behaviour to store the keys into key-chain manager upon enrollment approval.
 - Removed the optional parameter "filePath" in the enroll method because we do not generate atKeys file and filePath is not used.
 - Removed the code related to generating of atKeys file.

**- How to verify it**
- Unit tests should pass.

### DIFF
--- a/packages/at_client_mobile/lib/src/keychain_manager.dart
+++ b/packages/at_client_mobile/lib/src/keychain_manager.dart
@@ -5,9 +5,9 @@ import 'dart:typed_data';
 
 import 'package:at_client_mobile/src/atsign_key.dart';
 import 'package:at_utils/at_logger.dart';
+import 'package:biometric_storage/biometric_storage.dart';
 import 'package:crypton/crypton.dart';
 import 'package:encrypt/encrypt.dart' as encrypt;
-import 'package:biometric_storage/biometric_storage.dart';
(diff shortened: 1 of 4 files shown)


packages/at_client_mobile/lib/src/keychain_manager.dart
```diff
@@ -24,6 +26,7 @@ class KeyChainManager {
   static final KeyChainManager _singleton = KeyChainManager._internal();
 
   static final _logger = AtSignLogger('KeyChainUtil');
+  late PackageInfo packageInfo;
```
murali-shris: should this var be public?

murali-shris (dismissed):

packages/at_client_mobile/lib/src/keychain_manager.dart
sitaram-kalluri: It do have usage outside this class. Marked as private variable.

murali-shris (approved):

PR #1321, opened 2024-05-20
feat: Add enrollment setup and run tests with APKAM authentication

### LINKED ISSUE #1309: Need some e2e tests for APKAM-enrolled clients
### Describe the bug

We have no e2e tests right now which prove that APKAM-enrolled atsign_1 can successfully communicate with APKAM-enrolled atsign_2

### Expected behavior

Something like this in the e2e tests:
- create and then approve an APKAM enrollment for atSign1
- create and then approve an APKAM enrollment for atSign2
- Test put, get, delete (and their auto-notifications) atSign1 to atSign2 using APKAM to APKAM, APKAM to pkam, pkam  APKAM (in this shorthand terminology, pkam to pkam is what we currently test)
- Ideally refactor existing e2e tests to extract out the common stuff so we can then run them with whichever combination we want (current: pkam to pkam)

sitaram-kalluri:
**- What I did**
- Modify the end-to-end tests to run with APKAM authentication.
- Add "revoke" method to enrollment_service.dart to support "enroll:revoke" operation.
- In AtClientImpl.dart, if the AtKey.namespace is null, we set the namespace from AtClientPreferences to the key. In collection framework, the namespace of the key is added to the "AtKey.key" and "AtKey.namespace" is unset. As a result the namespace from the "AtClientPreferences" is added causing duplicate/multiple namespaces being added to the key. Modified the code in "default_key_maker.dart" to fix the issue. Removed the namespace from the AtKey.key and added to AtKey.namespace.

**- How I did it**
- Add "enrollment_setup.dart" script that submits an enrollment request to the server and approves the enrollment request. 
Upon approval of the request, the authentication keys are written to the atKeys file which are used by the tests for authentication. 
- The enrollment_setup.dart file is invoked by the github actions to generate at the atkeys file prior to running of tests.
- In Config14.yaml, introduce "authType" variables which is set to "apkam"  to run authenticate at the atsigns via the APKAM authentication. The tests run between the trunk version of sever - @ce2e1 and canary version of server -@ce2e4.  In Config12.yaml, the "authType" is set to "pkam" to retain the existing way of running the tests.
- In the "setUpAll" of each test, fetch the "authType" from the config.yaml to perform the authentication.
- Since "notify:text" is being deprecated, removed the tests related to it.
- In config14.yaml, the firstAtSign and secondAtSign are configured as "@ce2e1" and "@ce2e2" respectively, causing the tests to run against the trunk branch on both servers. However, the intended behavior of config14.yaml is to test against both the trunk branch and the canary version on the server. Therefore, updated the secondAtSign to "@ce2e4" and the fourthAtSign to "@ce2e2" to align with this intention.
- Add "enrollment_teardown.dart" file which will revoke/deny the approved/pending enrollments respectively after the running the tests.

**- How to verify it**
- All tests should pass.

- Run E2E tests with APKAM authentication.


tests/at_end2end_test/test/enrollment_setup.dart
```diff
@@ -0,0 +1,239 @@
...
+      AtLookUp atLookUp = AtLookupImpl(
+          currentAtSign,
+          AtClientManager.getInstance().atClient.getPreferences()!.rootDomain,
+          AtClientManager.getInstance().atClient.getPreferences()!.rootPort);
+
+      EnrollmentRequest enrollmentRequest = EnrollmentRequest(
+          appName: 'wavi-$random',
+          deviceName: 'iphone',
+          otp: 'ABC123',
+          namespaces: {'wavi': 'rw', 'buzz': 'rw'});
```
gkc: We should clean up enrollments from previous test runs. In apkam_setup.sh the NoPorts e2e tests we do the following before submitting the new enrollment request for this test run:
```
  logInfo "Denying any pending enrollment requests for $atSign with apkamAppName $apkamApp and apkamDeviceName $apkamDev"
  $authBinary deny -r "$atDirectoryHost" -a "$atSign" --arx "$apkamApp" --drx "$apkamDev" || return $?

  logInfo "Revoking any approved enrollment requests for $atSign with apkamAppName $apkamApp and apkamDeviceName $apkamDev"
  $authBinary revoke -r "$atDirectoryHost" -a "$atSign" --arx "$apkamApp" --drx "$apkamDev" || return $?
```
where `$apkamApp` has been set to "e2e_all"

For these tests I suggest
- use an apkam appName of 'e2e_test'
- before submitting the new enrollment request:
  - fetch all pending enrollment requests where the appName is 'e2e_test', and deny all of them
  - fetch all approved enrollment requests where the appName is 'e2e_test', and revoke all of them

gkc (changes requested):
Excellent work @sitaram-kalluri. Just one change I see as necessary right now, see review comment

tests/at_end2end_test/lib/src/test_initializers.dart
```diff
@@ -16,17 +19,32 @@ class TestSuiteInitializer {
     return _singleton;
   }
 
-  Future<void> testInitializer(String atSign, String namespace) async {
+  Future<void> testInitializer(String atSign, String namespace,
+      {String authType = 'pkam'}) async {
```
gkc: For clarity in the test cases which use this function, would it be better to make `authType` a required parameter so that it's always explicit in the test code which authType is being used?

sitaram-kalluri:
Currently, the E2E tests are running successfully. The unit tests are failing with dart beta version due to following dart issue: https://github.com/dart-lang/sdk/issues/55911#issuecomment-2153606884. Pending work is to update the dependencies and check if the issue is resolved.

gkc (approved):
Great work. Thank you @sitaram-kalluri

PR #1348, opened 2024-07-01
fix: Enable at_client to initialize in offline when network is down

### LINKED ISSUE #915: Being able to get an AtClient without having connectivity to your atServer (i.e offline access)

sitaram-kalluri:
**- What I did**
- This PR introduces offline access capabilities to the AtClient SDK, allowing the client to initialize for offline access when network connection is not available during the authentication of an atSign. 

**- How I did it**

- The "authenticate" method has been updated to support offline access for the AtClient. Within the method, the call to "authenticate" from the AtAuth package is enclosed in a try block. If a network failure occurs during this call, the exception is caught. The code then checks if the user is already onboarded. If the user is onboarded, the AtClient is subsequently initialized for offline access.
- Added AtLookup as optional parameter to AtAuthServiceImpl to enable injecting mock instance for writing unit tests. If the optional parameter "AtLookup" is not injected, then the it is initialized in "_initAtClient" method which is the existing behaviour.

**- How to verify it**
- Added unit tests for the below scenarios
  - Test to verify AtClient initializes successfully in offline mode upon network failure when keychain manager contains keys.
  - Test to verify AtClient initialization fails when network is offline and keychain manager does not have keys.
  - Test to verify authenticate is successful and initialization is successful 

- Enable at_client to initialize in offline when network is down


gkc:
Thanks @sitaram-kalluri ... I will set aside some time tomorrow to test this out in various scenarios with noports and at_talk. @cconstab you may be interested also

gkc:
> Thanks @sitaram-kalluri ... I will set aside some time tomorrow to test this out in various scenarios with noports and at_talk. @cconstab you may be interested also

Ah this only applies to at_client_mobile. What manual testing has been done with apps, @sitaram-kalluri ?

gkc:
> > Thanks @sitaram-kalluri ... I will set aside some time tomorrow to test this out in various scenarios with noports and at_talk. @cconstab you may be interested also
> 
> Ah this only applies to at_client_mobile. What manual testing has been done with apps, @sitaram-kalluri ?

Manual testing documented [here](https://github.com/atsign-foundation/at_client_sdk/issues/915#issuecomment-2213273789)

gkc:
Following discussion with @sitaram-kalluri, we will look at introducing a feature flag so that application code needs to opt in to this behaviour rather than having it happen by default, as it can lead to changes in app behaviour if this feature is enabled by default

gkc:
> Following discussion with @sitaram-kalluri, we will look at introducing a feature flag so that application code needs to opt in to this behaviour rather than having it happen by default, as it can lead to changes in app behaviour if this feature is enabled by default

After further discussion, we have decided not to introduce a feature flag.

gkc (approved):

PR #1359, opened 2024-07-22
feat: Add feature flag to allow atClient for offline access

sitaram-kalluri:
**- What I did**
- Reference to the PR: https://github.com/atsign-foundation/at_client_sdk/pull/1348, added feature flag to allow at_client initialize in offline access.

**- How I did it**
- Added a preference in AtClientPreferences class. The default value is set to false to retain the existing behavior. If the flag is set to true, at_client is initialized even when network connectivity is not available.


gkc:
@sitaram-kalluri As discussed, let's close this PR as unnecessary.

PR #1374, opened 2024-08-05
fix: Notify response transformer fix

### LINKED ISSUE #1332: Make NotificationResponseTransformer respect atNotification.isEncrypted == false
### Describe the bug

If a client receives a notification where isEncrypted is explicitly false, it nonetheless continues to treat it as though isEncrypted were true

### Steps to reproduce

Came across this bug when I was running an sshnpd with legacy atKeys at the same time as I was doing enroll requests; the enroll request notifications explicitly have isEncrypted == false (which is correct) but NotificationResponseTransformer nonetheless tries to decrypt the data and of course fails

### Expected behavior

explicit isEncrypted == false should be respected

### Additional context

See also https://github.com/atsign-foundation/at_server/issues/1944

murali-shris:
fixes: https://github.com/atsign-foundation/at_client_sdk/issues/1332
**- What I did**
- added check in NotificationResponseTransformer to decrypt the notification value only if isEncrypted is NOT false.
**- How I did it**
- added a check before decrypting notification value. decrypt notification value only if isEncrypted is null or true.
- added/modified unit tests in notification_service_test.dart
**- How to verify it**
- tests in notification_service_test.dart should pass

### DIFF
--- a/packages/at_client/lib/src/transformer/response_transformer/notification_response_transformer.dart
+++ b/packages/at_client/lib/src/transformer/response_transformer/notification_response_transformer.dart
@@ -22,7 +22,6 @@ class NotificationResponseTransformer
       Tuple<AtNotification, NotificationConfig> tuple) async {
     // prepare the atKey from the atNotification object.
     AtNotification atNotification = tuple.one;
-    NotificationConfig config = tuple.two;
     String sharedBy = atNotification.from;
     String sharedWith = atNotification.to;
     var key = atNotification.key;
@@ -77,12 +76,6 @@ class NotificationResponseTransformer
       return atNotification;
     }
     return atNotification;
-  }
-
-  // Decrypt the notification key only if isEncrypted flag is set to true or
-  // null (for backward compatibility - default behavior is decrypt the notification key)
-  bool _shouldDecryptKey(bool? isEncrypted) {
-    return isEncrypted == null || isEncrypted == true;
   }
 
   Future<String> _getDecryptedValue(AtKey atKey, String? encryptedValue) async {
--- a/packages/at_client/lib/src/transformer/response_transformer/notification_response_transformer.dart
+++ b/packages/at_client/lib/src/transformer/response_transformer/notification_response_transformer.dart
@@ -77,8 +77,8 @@ class NotificationResponseTransformer
     return atNotification;
   }
 
-  // Decrypt the notification key only if isEncrypted false is set to true or
-  // not set(for backward compatibility where default behavior is encrypt the notification key)
+  // Decrypt the notification key only if isEncrypted flag is set to true or
+  // null (for backward compatibility - default behavior is decrypt the notification key)
   bool _shouldDecryptKey(bool? isEncrypted) {
     return isEncrypted == null || isEncrypted == true;
   }
--- a/packages/at_client/lib/src/transformer/response_transformer/notification_response_transformer.dart
+++ b/packages/at_client/lib/src/transformer/response_transformer/notification_response_transformer.dart
@@ -22,6 +22,7 @@ class NotificationResponseTransformer
       Tuple<AtNotification, NotificationConfig> tuple) async {
     // prepare the atKey from the atNotification object.
     AtNotification atNotification = tuple.one;
+    NotificationConfig notificationConfig = tuple.two;
     String sharedBy = atNotification.from;
     String sharedWith = atNotification.to;
     var key = atNotification.key;
@@ -68,8 +69,9 @@ class NotificationResponseTransformer
         // and is not a user created key.
         // Hence do not decrypt if key's are reserved keys
         AtKey.getKeyType(atKey.toString()) != KeyType.reservedKey) {
-      // decrypt the notification value only if isEncrypted is not set to false
-      if (atNotification.isEncrypted != false) {
+      // decrypt the notification value only if isEncrypted is not set to false and shouldDecrypt is set to true
+      if (atNotification.isEncrypted != false &&
+          notificationConfig.shouldDecrypt) {
         atNotification.value =
             await _getDecryptedValue(atKey, atNotification.value!);
       }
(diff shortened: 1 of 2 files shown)


packages/at_client/lib/src/transformer/response_transformer/notification_response_transformer.dart
```diff
@@ -76,6 +77,12 @@ class NotificationResponseTransformer
     return atNotification;
   }
 
+  // Decrypt the notification key only if isEncrypted flag is set to true or
+  // null (for backward compatibility - default behavior is decrypt the notification key)
+  bool _shouldDecryptKey(bool? isEncrypted) {
```
VJag: I think the doc should be like "Returns true when a notification is encrypted and needs a decryption etc.." also prefer this method taking a notification as an argument

VJag: Then we can write tests around that flag being null, true and false

murali-shris: > I think the doc should be like "Returns true when a notification is encrypted and needs a decryption etc.." also prefer this method taking a notification as an argument

there are two fields on which we do encryption/decryption. 1) notification key 2) notification value. hence i didn't add a method to decrypt by passing in notification object

VJag: Not to decrypt, it is the bool method.

gkc (changes requested):
The changes so far don't address the primary problem as described in the associated issue, which is that the code will attempt to decrypt a standard `update` notification even if isEncrypted is false

```
if ((atNotification.value.isNotNull) &&
        (config.shouldDecrypt && atNotification.id != '-1') &&
        // The shared_key (which is a reserved key) has different decryption process
        // and is not a user created key.
        // Hence do not decrypt if key's are reserved keys
        AtKey.getKeyType(atKey.toString()) != KeyType.reservedKey) {
      // decrypt the value
      atNotification.value =
          await _getDecryptedValue(atKey, atNotification.value!);
      return atNotification;
    }
```

Please add unit tests for standard update notifications to verify (1) that decryption will be attempted if isEncrypted is null (2) that decryption will be attempted if isEncrypted is true (3) that decryption will not be attempted if isEncrypted is false and make implementation changes required to make test (3) pass

murali-shris:
> The changes so far don't address the primary problem as described in the associated issue, which is that the code will attempt to decrypt a standard `update` notification even if isEncrypted is false
> 
> ```
> if ((atNotification.value.isNotNull) &&
>         (config.shouldDecrypt && atNotification.id != '-1') &&
>         // The shared_key (which is a reserved key) has different decryption process
>         // and is not a user created key.
>         // Hence do not decrypt if key's are reserved keys
>         AtKey.getKeyType(atKey.toString()) != KeyType.reservedKey) {
>       // decrypt the value
>       atNotification.value =
>           await _getDecryptedValue(atKey, atNotification.value!);
>       return atNotification;
>     }
> ```
> 
> Please add unit tests for standard update notifications to verify (1) that decryption will be attempted if isEncrypted is null (2) that decryption will be attempted if isEncrypted is true (3) that decryption will not be attempted if isEncrypted is false and make implementation changes required to make test (3) pass

@gkc in current trunk code, there are two flags used in notification response transformer, AtNotification --> isEncrypted and NotificationConfig --> shouldDecrypt.
isEncrypted is used for notification text where the value is sent as part of notification key. 
shouldDecrypt is used for standard update notification. 
Should I revert changes I did for isEncrypted and make the changes for  shouldDecrypt flag or use isEncrypted for both scenarios?

gkc:
@murali-shris
here's the relevant block of code that needs to be changed:
```
else if ((atNotification.value.isNotNull) &&
        (config.shouldDecrypt && atNotification.id != '-1') &&
        // The shared_key (which is a reserved key) has different decryption process
        // and is not a user created key.
        // Hence do not decrypt if key's are reserved keys
        AtKey.getKeyType(atKey.toString()) != KeyType.reservedKey) {
      // decrypt the value
      atNotification.value =
          await _getDecryptedValue(atKey, atNotification.value!);
      return atNotification;
    }
```

- isEncrypted is part of the notification message's metadata
- currently, isEncrypted is **not checked** - i.e. the notification's value is always decrypted 
- the change required is to check if isEncrypted == false, and do not decrypt in that case, regardless of what shouldDecrypt says, since there is no point in even trying to decrypt a message which is not encrypted

gkc:
i.e. change this code from the above block from
```
      // decrypt the value
      atNotification.value =
          await _getDecryptedValue(atKey, atNotification.value!);
```

to

```
      if (atNotification.isEncrypted != false) {
        // decrypt the value
        atNotification.value =
        await _getDecryptedValue(atKey, atNotification.value!);
      }
```

packages/at_client/lib/src/transformer/response_transformer/notification_response_transformer.dart
```diff
@@ -62,15 +61,18 @@ class NotificationResponseTransformer
...
+        atNotification.id != '-1' &&
         // The shared_key (which is a reserved key) has different decryption process
         // and is not a user created key.
         // Hence do not decrypt if key's are reserved keys
         AtKey.getKeyType(atKey.toString()) != KeyType.reservedKey) {
-      // decrypt the value
-      atNotification.value =
-          await _getDecryptedValue(atKey, atNotification.value!);
+      // decrypt the notification value only if isEncrypted is not set to false
+      if (atNotification.isEncrypted != false) {
```
gkc: If it's not false then we need to decrypt - but only if `config.shouldDecrypt` is true i.e. this line should read, I think, `if (atNotification.isEncrypted != false && config.shouldDecrypt)`

packages/at_client/test/notification_service_test.dart
```diff
@@ -276,9 +276,35 @@ void main() {
...
+          await notificationResponseTransformer.transform(Tuple()
+            ..one = atNotification
+            ..two = (NotificationConfig()
+              ..regex = '.*'
+              ..shouldDecrypt = true));
+      expect(transformedNotification.value, 'decryptedValue');
+    });
+
+    test(
+        'A test to verify notification value is decrypted when isEncrypted is set to true',
```
gkc, line 305: We need another test, to verify that the notification value is not decrypted when (a) isEncrypted is true, and (b) shouldDecrypt is false

gkc (changes requested):

gkc (approved):

PR #1398, opened 2024-09-16
feat: at_client put method - introduce shouldEncrypt flag

### LINKED ISSUE #1401: Introduce shouldEncrypt in PutRequestOptions for clients to not encrypt self key/shared key data

* Currently isEncrypted in metadata is set to false even though data is encrypted.
* Client don't have control if they choose not to encrypt data for self key/shared key

* Introduce a flag shouldEncrypt in PutRequestOptions which will be true by default
* New client can still see old data which can be encrypted with isEncrypted=false. Handle this scenario while decrypting

_No response_

### Additional context

_No response_

murali-shris:
Fixes https://github.com/atsign-foundation/at_client_sdk/issues/1401
Depends on https://github.com/atsign-foundation/at_server/pull/2099 for intermittent functional test failure in sync_multiple_client_test

**- What I did**
- Ability for clients to set shouldEncrypt=false when they don't want to encrypt data for self and shared key

**- How I did it**
- added a flag shouldEncrypt with default value=true in PutRequestOptions. This value will be ignored for public keys(metadata.isEncrypted is always false). For self and sharedkey this value will be honoured and data will be encrypted(shouldEncrypt=true) or not encrypted(shouldEncrypt=false)
- Refactored put_request_transformer for readability
- Changes in get_response_transformer to handle two scenarios  1) isEncrypted for new data(post these changes where isEncrypted=true for self/shared keys) 2) isEncrypted=false for old self/shared key.
- modified sync_multiple_client_test to replicate the intermittent issue consistently
- added unit and functional test

**- How to verify it**
- tests should pass
Dependency tests
https://github.com/atsign-foundation/at_libraries/pull/673
https://github.com/atsign-foundation/noports/pull/1387
TODO:
Replace dependency overrides with published versions
at_commons PR to be reviewed and merged 
https://github.com/atsign-foundation/at_libraries/pull/669


murali-shris:
fix for functional test issue in server
https://github.com/atsign-foundation/at_server/pull/2099/files

packages/at_client/lib/src/transformer/response_transformer/get_response_transformer.dart
```diff
@@ -35,39 +36,66 @@ class GetResponseTransformer
...
+    } else {
+      // for old data, try decrypting the value. if decryption fails, set the original value.
       try {
-        atValue.value =
-            await decryptionService.decrypt(tuple.one, atValue.value) as String;
-      } on AtException catch (e) {
-        e.stack(AtChainedException(Intent.fetchData,
-            ExceptionScenario.decryptionFailed, 'Failed to decrypt the data'));
-        rethrow;
+        atValue.value = await _decrypt(atValue, decryptionService, tuple.one);
```
gkc, line 54: Are you distinguishing between different types of exceptions that could be thrown here - for example unable to fetch or decrypt the symmetric key is different from having everything that's required but having the decryption fail (because it's not encrypted)

gkc (changes requested):

packages/at_client/lib/src/transformer/response_transformer/get_response_transformer.dart
murali-shris: yes Gary. Trying to decrypt data that is not encrypted results in FormatException. Other decryption related exceptions are thrown as AtException(existing behavior) from _decrypt

packages/at_client/lib/src/transformer/response_transformer/get_response_transformer.dart
```diff
@@ -35,39 +36,66 @@ class GetResponseTransformer
...
+    }
+    decryptionManager ??= AtKeyDecryptionManager(_atClient);
+    var decryptionService =
+        decryptionManager!.get(tuple.one, _atClient.getCurrentAtSign()!);
     // Decrypt the data, for other keys
-    if (!(decodedResponse['key'].startsWith('public:')) &&
-        !(decodedResponse['key'].startsWith('cached:public:'))) {
-      var decryptionService = AtKeyDecryptionManager(_atClient)
-          .get(tuple.one, _atClient.getCurrentAtSign()!);
+    // For new encrypted data after AtClient v3.2.1, isEncrypted will be true by default
```
gkc: What does "by default" mean? Is it more accurate to say that "isEncrypted will be set to true when data is encrypted"?

gkc (changes requested):
https://github.com/atsign-foundation/at_client_sdk/pull/1398#discussion_r1776736042

packages/at_client/lib/src/transformer/response_transformer/get_response_transformer.dart
murali-shris: isEncrypted will be set to true for self/shared keys, if caller doesn't set anything in PutRequestionOptions. I will change the wording.

gkc (approved):

PR #1421, opened 2024-10-22
feat: add the AtClientBindings mixin to this package

gkc:
**- What I did**
feat: add the AtClientBindings mixin to this package. This mixin was initially added in the noports_core package, but has broader applicability

**- How I did it**
- added lib/src/mixins/at_client_bindings.dart
- added lib/at_client_mixins.dart
- updated pubspec, changelog and at_client_config for version 3.3.0

**- How to verify it**
- manually verified via noports packages

**-Additional context**
Note that at_client_bindings.dart is not exported via the main export file (lib/at_client.dart) but in a new export file (lib/at_client_mixins.dart) in order that we not cause any annoying import changes to be needed in packages which already import at_client package and already have an AtClientBindings defined locally

### DIFF
--- a/packages/at_client/CHANGELOG.md
+++ b/packages/at_client/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 3.3.0
+- feat: add the AtClientBindings mixin which was initially added to the 
+  noports_core package but has broader applicability.
+
 ## 3.2.2
 - build[deps]: Upgraded dependencies for the following packages:
   - at_commons to v5.0.0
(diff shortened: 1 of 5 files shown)


xavierchanth (approved):

xavierchanth:
Do we need some dartdoc explaining usage and when you should or shouldn't use this?

gkc:
> Do we need some dartdoc explaining usage and when you should or shouldn't use this?

I will make another PR after at_policy package has been published

PR #1440, opened 2024-11-25
feat: Uptake public key hash changes

sitaram-kalluri:
**- What I did**
- Uptake "PublicKeyHash" changes into at_client package.
   -  In "SharedKeyDecryption.dart", when decrypting the shared key, compare the hash value of the encryption public key with the hash value in atKey.metadata.pubKeyHash. If the value mismatch, throw exception.
   - Add publicKeyHash value in sync_service_impl.dart to sync the value to the remote secondary and fetch the value from the remote secondary.
   - Add  the publicKeyHash value in at_notification.metadata.

**- How to verify it**
- Updated the existing unit test to verify publicKeyHash is populated when encrypting the value.
- Updated the existing unit test to verify the exception is thrown if publicKeyHash mismatch.
- Updated the existing unit test to verify decryption is successful if publicKeyHash value matches. 

- feat: Uptake public key hash changes

Note: In the pubspec.yaml, the at_persistence_secondary_server is set to "2121-uptake-public-key-hash-at-persistence_secondary-server" branch to consume the public key hash changes. Once the [PR](https://github.com/atsign-foundation/at_server/pull/2159) in at_server is merged, will replace with the hosted version.


packages/at_client/lib/src/response/at_notification.dart
```diff
@@ -38,8 +38,8 @@ class AtNotification {
       metadata.sharedKeyEnc = json['metadata'][AtConstants.sharedKeyEncrypted];
       // AtContants.sharedWithPublicKeyHash will be sent by the server starting v3.0.52
       // Notifications received from Secondary server before 3.0.52 does not contain
-      // AtConstants.sharedWithPublicKeyHash. Therefore, check for null.
-      if (json['metadata'][AtConstants.sharedWithPublicKeyHash] != null) {
+      // AtConstants.sharedWithPublicKeyHash. Therefore, check for null String.
+      if (json['metadata'][AtConstants.sharedWithPublicKeyHash] != "null") {
```
gkc: This seems bad

sitaram-kalluri: Sure Gary, will revisit the code and modify the IF condition.

sitaram-kalluri: In the `MonitorVerbHandler.dart` file of `at_secondary_server`, the `pubKeyHash` object is converted into a JSON-encoded string and sent to the client. When the pubKeyHash value is null, the jsonEncode method returns a string with the value "null", which causes metadata.pubKeyHash to be set to `"null"` instead of `null`.


Code snippet from MonitorVerbHandler.dart:

`"pubKeyHash": jsonEncode(atNotification.atMetadata?.pubKeyHash?.toJson())`


To ensure that metadata.publicKeyHash contains a null object on the client side, modify the code in `MonitorVerbHandler.dart` as follows:

```
notification.metadata?.putIfAbsent(
          "pubKeyHash",
          () => (atNotification.atMetadata?.pubKeyHash != null)
              ? jsonEncode(atNotification.atMetadata?.pubKeyHash?.toJson())
              : null);
```

@gkc : Let me know if this approach looks good or suggest an alternate approach please.

gkc: Yep looks fine. Please add unit test(s) to cover

sitaram-kalluri: Sure, Thanks Gary.

gkc (approved):

Issue #404, opened 2022-02-15
AtClientManager.synchService.isInSync crashes after a few calls .


cconstab:
**Describe the bug**
Crashed app when checking that status of Sync between local and remote secondaries

**To Reproduce**
When using this code 

```
  while (true) {
    print('sync in progress');
    var a = await atClientManager.syncService.isInSync();
    print('Are we waiting ?' + a.toString());
    sleep(const Duration(seconds: 5));
  }
  print('OK Ready');
  ```
  
  I always get a Unhandled exception after a few cycles..
```
 Error : FormatException: Unexpected character (at character 1)
data:[{"atKey":"_latestnotificationid.ai6bh@ai6bh","value":"{\"id\":\"50dc4...
^


Unhandled exception:
NoSuchMethodError: The method '[]' was called on null.
Receiver: null
Tried calling: [](0)
#0      Object.noSuchMethod (dart:core-patch/object_patch.dart:68:5)
#1      SyncUtil.getLatestServerCommitId (package:at_client/src/util/sync_util.dart:87:18)
<asynchronous suspension>
#2      SyncServiceImpl._getServerCommitId (package:at_client/src/service/sync_service_impl.dart:410:27)
<asynchronous suspension>
#3      SyncServiceImpl.isInSync (package:at_client/src/service/sync_service_impl.dart:392:26)
<asynchronous suspension>
#4      main (file:///C:/Users/colin/GitHub/@company/mwc_demo/dart/iot_sender/bin/iot_sender.dart:52:13)
<asynchronous suspension>
PS C:\Users\colin\GitHub\@company\mwc_demo\dart\iot_sender> 

```

Looks like a low level protocol problem 

**Expected behavior**

**Screenshots**

**Were you using an @‎application when the bug was found?**
- MWC demo sender code

**Additional context**
Isthere or should there be a better way to check ?

gkc:
@murali-shris please can you take a look since you're working on the sync progress listener at the moment

murali-shris:
> @murali-shris please can you take a look since you're working on the sync progress listener at the moment

Sure Gary.

murali-shris:
root cause:
The call to isInSync from the main method, interferes with a sync from server. Both share the same remote secondary. Hence we get sync verb response in the place we are expecting latest server commit Id.
possible resolution:
in the call to isInSync create a new remote secondary instead of reusing the remote secondary.

To be discussed in architecture call

murali-shris:
[...]

PR #1448, opened 2024-12-10
fix: isInSync bug

### LINKED ISSUE #1444: SyncServiceImpl.isInSync() does not work with APKAM
### Describe the bug

`SyncServiceImpl.isInSync()` creates a new RemoteSecondary each time it is called. The reasons for doing this are not entirely clear, given that `SyncServiceImpl.create()` already creates a RemoteSecondary , but that is another story.

In SyncServiceImpl.create() the enrollmentId is passed to the RemoteSecondary constructor:
```
    remoteSecondary ??= RemoteSecondary(
        atClient.getCurrentAtSign()!, atClient.getPreferences()!,
        atChops: atClient.atChops, enrollmentId: atClient.enrollmentId);
```

However in `SyncServiceImpl.isInSync()` it is not:
```
      remoteSecondary = RemoteSecondary(
          _atClient.getCurrentAtSign()!, _atClient.getPreferences()!,
          atChops: _atClient.atChops);
```

Authentication will then fail, since the enrollmentId must be supplied in order to authenticate with APKAM keys.

### Steps to reproduce

Create a program like this:
```
import 'dart:io';

import 'package:at_cli_commons/at_cli_commons.dart';

Future<void> main(List<String> args) async {
  try {
    var atClient = (await CLIBase.fromCommandLineArgs(args)).atClient;
    await atClient.syncService.isInSync();
    exit(0);
  } catch (e) {
    if (e is ArgumentError) {
      print(CLIBase.argsParser.usage);
    }
    print(e);
    exit(1);
  }
}
```

its dependencies are
```
dependencies:
  at_cli_commons: ^1.3.0
  at_client: ^3.3.0
```

Running it with a PKAM keys file:
```
gary@gkc2019 example % dart bin/is_in_sync_example.dart -a @garycasey -n test
Connecting ... Connected
Is in sync? false
```

Running it with an APKAM keys file:

```
gary@gkc2019 example % dart bin/is_in_sync_example.dart -a @garycasey -k ~/@garycasey.test.mbp.atKeys -n test
Connecting ... Connected
Exception: Exception: Unable to fetch latest server commit id: Exception: Exception: Failed connecting to @garycasey. error:{"errorCode":"AT0401","errorDescription":"Exception: pkam authentication failed"}
```

### Expected behavior

isInSync should also pass enrollmentId to the RemoteSecondary constructor

### Additional context

As part of fixing this I'd also like to see whether we can eliminate isInSync's creation of its own RemoteSecondary

murali-shris:
fixes https://github.com/atsign-foundation/at_client_sdk/issues/1444
**- What I did**
- Fix for isInSync bug in SyncServiceImpl
**- How I did it**
- remove the logic to create new secondary in isInSync method. Re-use _remoteSecondary from create(..) method
- Logic to create new secondary in isInSync was added for an old bug 
https://github.com/atsign-foundation/at_client_sdk/issues/404. This bug is no longer reproducible. Hence removing this logic.
**- How to verify it**
- Tests done
https://github.com/atsign-foundation/at_client_sdk/issues/1444#issuecomment-2532111449

### DIFF
--- a/packages/at_client/lib/src/service/sync_service_impl.dart
+++ b/packages/at_client/lib/src/service/sync_service_impl.dart
@@ -539,6 +539,7 @@ class SyncServiceImpl implements SyncService, AtSignChangeListener {
   /// commit-id to sync into the local keystore.
   Future<List<dynamic>> _getEntriesToSyncFromServer(
       int lastReceivedServerCommitId) async {
+    // Sync verb syntax has to be changed before removing these deprecations
     var syncBuilder = SyncVerbBuilder()
       ..commitId = lastReceivedServerCommitId
       ..regex = _atClient.getPreferences()!.syncRegex
@@ -767,13 +768,8 @@ class SyncServiceImpl implements SyncService, AtSignChangeListener {
   ///Throws [AtClientException] if cloud secondary is not reachable
   @override
   Future<bool> isInSync() async {
-    late RemoteSecondary remoteSecondary;
     try {
-      remoteSecondary = RemoteSecondary(
-          _atClient.getCurrentAtSign()!, _atClient.getPreferences()!,
-          atChops: _atClient.atChops);
-      var serverCommitId =
-          await _getServerCommitId(remoteSecondary: remoteSecondary);
+      var serverCommitId = await _getServerCommitId();
 
       var lastReceivedServerCommitId = await getLastReceivedServerCommitId();
 
@@ -794,8 +790,6 @@ class SyncServiceImpl implements SyncService, AtSignChangeListener {
       var cause = (e is AtException) ? e.getTraceMessage() : e.toString();
       _logger.severe('exception in isInSync $cause');
       throw AtClientException.message(e.toString());
-    } finally {
-      unawaited(remoteSecondary.atLookUp.close());
     }
   }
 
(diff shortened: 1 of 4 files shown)


gkc:
@murali-shris I presume the reason the bug from way back is no longer reproducible is because of [this](https://github.com/atsign-foundation/at_libraries/blob/trunk/packages/at_lookup/lib/src/at_lookup_impl.dart#L595)

packages/at_client/lib/src/service/sync_service_impl.dart
```diff
@@ -539,6 +539,7 @@ class SyncServiceImpl implements SyncService, AtSignChangeListener {
   /// commit-id to sync into the local keystore.
   Future<List<dynamic>> _getEntriesToSyncFromServer(
       int lastReceivedServerCommitId) async {
+    // Sync verb syntax has to be changed before removing these deprecations
```
gkc, line 542: Can you do a PR in at_lookup to change SyncVerbBuilder so it has default values for the deprecated things; and then do another PR here to remove the deprecated params

gkc (approved):

gkc:
@murali-shris approved but please see my request above https://github.com/atsign-foundation/at_client_sdk/pull/1448#discussion_r1879803959

PR #1499, opened 2025-04-01
fix: add retries to at_client_bindings.notify (3 by default)

### LINKED ISSUE #1828: Notifications to retry if there is a network issue for instance

Occasionally either the desktop app or the command line tools "hang" /fail and have to be restarted. When testing at_talk in poor network conditions, it is resilient because it retries notifications three times and holds back for few seconds for the network to get back online.

To include something like this in the notifications used in the noports dart implementations, from the at_talk code

```dart
  // back off retries (max 3)
  for (int retry = 1; retry < 4; retry++) {
    try {
      NotificationResult result = await notificationService.notify(
          NotificationParams.forUpdate(key,
              value: input, notificationExpiry: Duration(seconds: 360), strategy: StrategyEnum.all),
          waitForFinalDeliveryStatus: false,
          onSentToSecondary: (p0) {},
          checkForFinalDeliveryStatus: false);
      if (result.atClientException != null) {
        logger.warning(result.atClientException);
        print('${chalk.brightRed.bold('\r\x1b[KError Sending: ')} retry number $retry of 3');
        await Future.delayed(Duration(milliseconds: (500 * (retry))));
      } else {
        success = true;
        break;
      }
    } catch (e) {
      logger.warning(e);
    }
  }
  return (success);
}
```

have this in the atclient notification code itself behind the scenes..

### Additional context

_No response_

xavierchanth:
closes https://github.com/atsign-foundation/noports/issues/1828

(I don't know if this ^^ will auto close issues in other repos, but I'm curious to find out)

**- What I did**

- Changed return of `AtClientBindings.notify` type from `Future<void>` to `Future<NotificationResult>`
- Added new optional parameter called `maxTries` which defaults to 3
- wrapped the notify command in a retry loop

**- How I did it**

- Used a `do while` instead of `for` which seems weird, but it ensures a few things:
  - The notify always runs at least once
    - Thus, result is never left unset before we reach the end of the function,
      which can't be guaranteed with a for loop.
  - Any maxTries value less than 1 is effectively silently set to 1, 
    instead of us needing to do validation and throwing on invalid input

**- How to verify it**

fix: add retries to at_client_bindings.notify (3 by default)

### DIFF
--- a/packages/at_client/lib/src/mixins/at_client_bindings.dart
+++ b/packages/at_client/lib/src/mixins/at_client_bindings.dart
@@ -6,25 +6,44 @@ mixin AtClientBindings {
 
   AtSignLogger get logger;
 
-  Future<void> notify(
+  Future<NotificationResult> notify(
     AtKey atKey,
     String value, {
     required bool checkForFinalDeliveryStatus,
     required bool waitForFinalDeliveryStatus,
     required Duration ttln,
+
+    /// maxTries must be a non-zero positive integer
+    int maxTries = 3,
   }) async {
-    await atClient.notificationService.notify(
-      NotificationParams.forUpdate(atKey,
-          value: value, notificationExpiry: ttln),
-      checkForFinalDeliveryStatus: checkForFinalDeliveryStatus,
-      waitForFinalDeliveryStatus: waitForFinalDeliveryStatus,
-      onSuccess: (NotificationResult notification) {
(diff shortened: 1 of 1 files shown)


gkc (approved):

xavierchanth:
> (I don't know if this ^^ will auto close issues in other repos, but I'm curious to find out)

This does in fact work! Yay!

PR #1520, opened 2025-06-08
build: update dependencies

gkc:
**- What I did**
build: update dependencies including major version changes of at_persistence packages
fix: fix syntax issue resulting from major version change in at_persistence packages
fix: fix AtCollection bug triggered by major version change in at_persistence packages
fix: tightened up code for handling AtKeyNotFoundExceptions in AtCollectionQueryOperationsImpl
chore: clean up lint resulting from upgrading lint package dependency to ^6.0.0

**- How I did it**

**- How to verify it**
Tests pass

### DIFF
--- a/packages/at_client/lib/src/client/at_client_impl.dart
+++ b/packages/at_client/lib/src/client/at_client_impl.dart
@@ -483,7 +483,7 @@ class AtClientImpl implements AtClient, AtSignChangeListener {
   }
 
   @visibleForTesting
-  ensureLowerCase(AtKey atKey) {
+  dynamic ensureLowerCase(AtKey atKey) {
     if (upperCaseRegex.hasMatch(atKey.key) ||
         (atKey.namespace != null &&
             upperCaseRegex.hasMatch(atKey.namespace!))) {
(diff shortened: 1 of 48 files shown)


packages/at_client/lib/src/client/at_client_impl.dart
```diff
@@ -483,7 +483,7 @@ class AtClientImpl implements AtClient, AtSignChangeListener {
   }
 
   @visibleForTesting
-  ensureLowerCase(AtKey atKey) {
+  dynamic ensureLowerCase(AtKey atKey) {
```
xavierchanth, line 486: I would prefer some doc comments whenever we use a dynamic type in a function signature.
But it also looks like this can maybe return an `AtResponse`, is PutResponseTransformer of that type?

gkc: @XavierChanth This function's return type should be `void` - however there is a unit test which asserts the return value of this function as `null` which failed when I made the return type void. ensureLowerCase is nothing to do with responses, it just makes sure that the AtKey's string representation contains no uppercase, and it does that by calling AtKey.toString(). This is a all a bit of a rat's nest but I didn't have time to do anything other than make the linter stop complaining.

xavierchanth (approved):

PR #1531, opened 2025-07-03
feat: add `atLookUp` parameter to AtClientManager.setCurrentAtSign etc

gkc:
**- What I did**
- feat: add `atLookUp` parameter to `AtClientManager.setCurrentAtSign`, `AtClientImpl.create`, etc so we can inject an existing AtLookUp instance if we have one rather than having to create a new one and authenticate again

**- How I did it**
See commits

**- How to verify it**
Tests pass

### DIFF
--- a/packages/at_client/lib/src/client/remote_secondary.dart
+++ b/packages/at_client/lib/src/client/remote_secondary.dart
@@ -23,12 +23,15 @@ class RemoteSecondary implements Secondary {
 
   late AtClientPreference _preference;
 
-  late AtLookupImpl atLookUp;
+  late AtLookUp atLookUp;
 
   final AtChops? atChops;
 
   RemoteSecondary(String atSign, AtClientPreference preference,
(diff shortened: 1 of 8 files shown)


packages/at_client/lib/src/client/remote_secondary.dart
```diff
@@ -86,7 +90,7 @@ class RemoteSecondary implements Secondary {
       logger.finer(logger.getLogMessageWithClientParticulars(
           _preference.atClientParticulars,
           'Command sent to server: ${builder.buildCommand()}'));
-      verbResult = await atLookUp.executeVerb(builder);
+      verbResult = (await atLookUp.executeVerb(builder))!;
```
gkc, line 93: AtLookUp and AtLookupImpl signatures for executeVerb differ - return value is nullable in AtLookUp (which it shouldn't be - will need a major version change there to fix that)

xavierchanth (approved):

athandle (approved):

PR #1534, opened 2025-07-08
fix: namespace edge case when sending notifications

gkc:
**- What I did**
- fix: ensure that namespaces in `notify` requests aren't messed up by multipart namespaces in AtClientPreference (e.g. namespace of `foo.bar`)

Assumptions in codebase were that the AtClientPreference.namespace would always be a simple one e.g. `my_app` or `wavi` or whatever. However there is nothing stopping apps setting AtClientPreference.namespace to a multipart namespace e.g. `foo.my_app` or `bar.wavi` or even`foo.bar.baz.wavi`

This bug was exposed by [this fix](https://github.com/atsign-foundation/at_client_sdk/pull/1533)

**- How I did it**
- Fixed faulty logic in NotificationRequestTransformer which was assuming a simple (single-word) AtClientPreference.namespace. See commits.
- Added test cases to cover 

**- How to verify it**
- [x] Tests pass here
- [x] NoPorts tests pass against this branch ([test PR](https://github.com/atsign-foundation/noports/pull/2034))

### DIFF
--- a/packages/at_client/lib/src/transformer/request_transformer/notify_request_transformer.dart
+++ b/packages/at_client/lib/src/transformer/request_transformer/notify_request_transformer.dart
@@ -77,7 +77,8 @@ class NotificationRequestTransformer
           ak.metadata.namespaceAware) {
         ak.namespace ??= atClientPreference.namespace;
         if (atClientPreference.namespace != null &&
-            ak.namespace != atClientPreference.namespace) {
+            !'${ak.key}.${ak.namespace}'
+                .endsWith('.${atClientPreference.namespace!}')) {
           ak.key = '${ak.key}.${ak.namespace}';
           ak.namespace = atClientPreference.namespace;
         }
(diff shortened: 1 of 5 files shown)


packages/at_client/lib/src/transformer/request_transformer/notify_request_transformer.dart
```diff
@@ -77,7 +77,8 @@ class NotificationRequestTransformer
           ak.metadata.namespaceAware) {
         ak.namespace ??= atClientPreference.namespace;
         if (atClientPreference.namespace != null &&
-            ak.namespace != atClientPreference.namespace) {
+            !'${ak.key}.${ak.namespace}'
+                .endsWith('.${atClientPreference.namespace!}')) {
           ak.key = '${ak.key}.${ak.namespace}';
           ak.namespace = atClientPreference.namespace;
         }
```
gkc, line 84: Imagine you have
- an AtKey whose key is `ping.device` and namespace is `sshnp`
- an AtClientPreference.namespace of `device.sshnp`

i.e. ak.namespace is not equal to atClientPreference.namespace, so the predicate is true and the AtKey would have become `ping.device.sshnp.device.sshnp`

but now we check if `${ak.key}.${ak.namespace}` (i.e. `ping.device.sshnp`) **_endsWith_** `.${atClientPreference.namespace}` (i.e. `.device.sshnp`) ... which it does not - so the predicate is false and nothing is appended

cconstab (approved):

xavierchanth (approved):

PR #1541, opened 2025-07-16
feat: deprecate AtClientPreference.atProtocolEmitted

gkc:
**- What I did**
feat: deprecate the (misleadingly named) `AtClientPreference.atProtocolEmitted` and change its default value from 1.5.0 to 2.0.0

**- How I did it**

**- How to verify it**

### DIFF
--- a/packages/at_client/lib/src/service/sync_service_impl.dart
+++ b/packages/at_client/lib/src/service/sync_service_impl.dart
@@ -496,8 +496,8 @@ class SyncServiceImpl implements SyncService, AtSignChangeListener {
             final keyInfo = KeyInfo(
                 serverCommitEntry['atKey'],
                 SyncDirection.remoteToLocal,
-                convertCommitOpSymbolToEnum(serverCommitEntry['operation']));
-            keyInfo.conflictInfo = conflictInfo;
+                convertCommitOpSymbolToEnum(serverCommitEntry['operation']))
+              ..conflictInfo = conflictInfo;
             keyInfoList.add(keyInfo);
             continue;
           }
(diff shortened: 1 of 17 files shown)


packages/at_client/lib/src/service/sync_service_impl.dart
```diff
@@ -623,42 +623,45 @@ class SyncServiceImpl implements SyncService, AtSignChangeListener {
...
+      if (clientAtKey is PublicKey || key.contains('public:')) {
         final serverValue = serverCommitEntry['value'];
         if (localAtValue.value != serverValue) {
           conflictInfo.localValue = localAtValue.value;
           conflictInfo.remoteValue = serverValue;
         }
         return conflictInfo;
       }
+      final serverAtKey = AtKey.fromString(clientAtKey.toString());
+      _setMetadataFromCommitEntry(serverAtKey.metadata, serverCommitEntry);
```
gkc, line 653: The change above fixes a bug where we were attempting, and failing, to decrypt the server copy using the metadata (including encryption metadata) of the client copy

packages/at_client/lib/src/service/sync_service_impl.dart
```diff
@@ -984,76 +987,71 @@ class SyncServiceImpl implements SyncService, AtSignChangeListener {
     return unCommittedEntryBatch;
   }
 
-  void _setMetaData(UpdateVerbBuilder builder, serverCommitEntry) {
+  void _setMetadataFromCommitEntry(Metadata md, Map serverCommitEntry) {
```
gkc, line 990: Renamed for readability, and pass a Metadata argument rather than an UpdateVerbBuilder argument so we can use this function in multiple places.

tests/at_functional_test/test/atclient_sync_conflict_test.dart
```diff
@@ -28,34 +31,25 @@ void main() async {
...
-        ..sharedBy = atSign
-        ..metadata = (Metadata()..isEncrypted = true))
-      ..value = 'sMBnYFctMOg+lqX67ah9UA=='; //encrypted value of 4
-
-    await atClientManager.atClient
-        .getRemoteSecondary()!
-        .executeVerb(updateVerbBuilder);
+    AtKey remoteKey = AtKey.fromString('phone_0.$namespace$atSign');
+    await atClientManager.atClient.put(remoteKey, 'Things have changed',
+        putRequestOptions: PutRequestOptions()..useRemoteAtServer = true);
```
gkc, line 43: Using `PutRequestOptions.useRemoteAtServer` instead of the clunkier use of an UpdateVerbBuilder

xavierchanth (approved):

PR #1622, opened 2025-08-13
feat(at_commons): add AtRootDomain class with parsing of proxy included

xavierchanth:
**- What I did**

- Added AtRootDomain class which stores the root domain and port.
- Based on how AtClientPreference expects proxies
  rootDomain in AtClientPreference should be "proxy:<domain>"

- AtRootDomain.parse factory which allows the following formats:
  - "domain"
  - "domain:port"
  - "proxy:domain"
  - "proxy:domain:port"

Validation included:
- Port is always checked for a valid integer and port number(1-65535).
- Domain: unchecked
- Proxy:
  - Checked that it's "proxy" when we have a 3 part string
  - When it's a 2 part string and not "proxy", it assumes domain:port.
    Thus something like `prozy:proxy0001.atsign.org` will be parsed as domain:port
    and fail because `proxy0001.atsign.org` is not a valid port.

**- How I did it**

**- How to verify it**

See at_root_domain_test.dart tests.

feat(at_commons): add AtRootDomain class with parsing of proxy included

### DIFF
--- a/packages/at_commons/CHANGELOG.md
+++ b/packages/at_commons/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 5.6.0
+
+- feat: add `AtRootDomain` with basic parsing including proxy.
+
 ## 5.5.0
 
 - chore(deps): bump uuid to "^4.0.0"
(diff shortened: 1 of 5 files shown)


JeremyTubongbanua:
I think AtRootServer should be the name of the class

xavierchanth:
> I think AtRootServer should be the name of the class

Calling it a server is misleading. It's just connection info. However `AtRootServerConnectionInfo` is very long.

JeremyTubongbanua (approved):

gkc (approved):

PR #1631, opened 2025-08-20
feat: at_stream package

xavierchanth:
**- What I did**

- I don't intend to publish yet, but would like people to be able to play with the code.

**- How I did it**

**- How to verify it**

- Demos run well
- Tests to follow in a later PR

feat: at_stream package


packages/at_stream/lib/src/at_notification_stream_channel.dart
```diff
@@ -0,0 +1,351 @@
...
+  bool _initStarted = false;
+  final Completer _initCompleter = Completer();
+  Future get initialized => _initCompleter.future;
+  Future get done => Future.wait([_sendController.done, _recvController.done]);
+
+  // Getters
+  @override
+  StreamSink<Value> get sink => _sendController.sink;
+  @override
+  Stream<Value> get stream => _recvController.stream;
```
xavierchanth, line 48: This is the only thing you need to provide to implement StreamChannel if you use the StreamChannelMixin as I have

packages/at_stream/lib/src/at_notification_stream_channel.dart
```diff
@@ -0,0 +1,351 @@
...
+      shouldDecrypt: true,
+    )
+        .transform(NotificationTransformer<Encoded>(this))
+        .transform(recvTransformer)
+        .listen((value) => _recvController.add(value));
+  }
+}
+
+@visibleForTesting
+class NotificationTransformer<Encoded>
```
xavierchanth: This transformer is applied to the receiving stream before the user's transformer is. It doesn't manipulate any Notifications, but it does remove notifications which are control messages. It provides a Dart record of (notification, value).

- notification provided for advanced use cases.
- value provided since the notification.value is a custom json structure that stores the user's value along side other information.

packages/at_stream/lib/src/at_notification_stream_channel.dart
```diff
@@ -0,0 +1,351 @@
...
+  /// yourself.
+  AtNotificationStreamChannel(
+    this.atClient, {
+    required this.otherAtsign,
+    required this.baseNamespace,
+    required this.domainNamespace,
+    required this.recvTransformer,
+    required this.sendTransformer,
+    this.sessionTimeout = const Duration(minutes: 30),
+    bool enableSigning = false, // TODO: add signing to this
```
xavierchanth: signing is not implemented yet.

packages/at_stream/lib/src/at_notification_stream_channel.dart
```diff
@@ -0,0 +1,351 @@
...
+  /// If you use this constructor, you must call [initServer] or [initClient]
+  /// yourself.
+  AtNotificationStreamChannel(
+    this.atClient, {
+    required this.otherAtsign,
+    required this.baseNamespace,
+    required this.domainNamespace,
+    required this.recvTransformer,
+    required this.sendTransformer,
+    this.sessionTimeout = const Duration(minutes: 30),
```
xavierchanth, line 59: What's a reasonable default here? I think this definitely needs to be lower.

xavierchanth:
Converted back to draft to write tests.

packages/at_client/lib/src/mixins/envelope_signing.dart
```diff
@@ -0,0 +1,170 @@
...
+        AtValue,
+        IllegalStateException;
+import 'package:at_client/at_client_mixins.dart' show ApkamSigning;
+import 'package:at_commons/at_commons.dart'
+    show AtSigningVerificationException, AtsignString;
+import 'package:at_utils/at_logger.dart' show AtSignLogger;
+import 'package:meta/meta.dart' show visibleForTesting;
+
+// TODO(xavierchanth): write tests
+mixin EnvelopeSigning on ApkamSigning {
```
gkc, line 29: LGTM my only request would be to mark EnvelopeSigning as `@experimental` for now, as we will probably want to extend it so you can choose to use public:publickey for the signing, or use the ApkamSigning's getPublicSigningKey and include the signingKeyUri in the signed envelope

gkc (changes requested):

packages/at_client/lib/src/mixins/envelope_signing.dart
xavierchanth: Marked as experimental in [e52fbce](https://github.com/atsign-foundation/at_client_sdk/pull/1631/commits/e52fbce3cb26e6a512a8fe2b3ee3b88f90eec3d1)

Xlin123 (dismissed):

gkc (dismissed):

PR #1632, opened 2025-08-21
feat!: Core interfaces and types for the platform

xavierchanth:
**- What I did**

#### at_commons

- chore!(at_commons): converted Atsign and AtsignWithoutAt from typedefs to extension types.
  - both of these types implement `String` but cannot be created without calling `String.toAtsign()` (which does the validation).
- feat(at_commons): New atbytes class which handles keys.
  - Stored as Uint8List since we need decoded keys to encrypt / decrypt. However can be created from and converted to base64 strings.

#### at_client_flutter (new package)

- will replace `at_client_mobile` and `at_onboarding_flutter`

**- How I did it**

**- How to verify it**

feat!: Core interfaces and types for the platform


packages/at_commons/lib/src/key/atbytes.dart
```diff
@@ -0,0 +1,17 @@
...
+  const AtBytes(this.bytes);
+
+  factory AtBytes.fromString(String str) {
+    return AtBytes(base64Decode(str));
+  }
+
+  @override
+  String toString() => base64Encode(bytes);
+
+  bool strEquals(String str) => toString() == str;
```
gkc, line 16: maybe also implement `operator ==(Object)` and `int get hashCode`

gkc (approved):

gkc (changes requested):
small request

PR #1635, opened 2025-08-25
fix: Resolves #1506

### LINKED ISSUE #1506: Very rare race condition with symmetric encryption keys
### Describe the bug

Two `@alice` clients simultaneously want to send encrypted info to `@bob` for the first time
- `@alice 1` checks if symmetric key exists and finds it does not
- `@alice 2` checks if symmetric key exists and finds it does not
- `@alice 1` cuts symmetric key AES1 and stores alice's copy (AES 1 self) to atServer
- `@alice 2` cuts symmetric key AES2 and stores alice's copy (AES2 self) to atServer, overwriting what `alice 1` just placed there
- `@alice 1` checks atServer for bob's copy of the symmetric key, does not find it
- `@alice 1` encrypts AES1 with bob's public key and (AES1 bob), stores locally, and stores to atServer for sharing with bob
- `@alice 2` checks atServer for bob's copy of the symmetric key, finds it (AES1 bob), stores it locally

Storage end state: 
- `@alice 1` has two entries in local storage - 'AES1 self' and 'AES1 bob'
- `@alice 2` has two entries in local storage - 'AES2 self' and 'AES1 bob'
- the atServer has two entries - 'AES2 self' and 'AES1 bob'

Outcomes
- When `@alice 1` sends, it is encrypting with AES1 and includes 'AES1 bob' in the metadata. `@bob` will decrypt 'AES1 bob' and use it to successfully decrypt the payload
- When `@alice 2` sends, it is encrypting with AES2 but including 'AES1 bob' in the metadata. `@bob` will decrypt 'AES1 bob' and _fail_ to decrypt the payload, as it was encrypted with AES2
- If a new client, `@alice 3`, wants to share, then it will pick up the values from atServer ('AES2 self' and 'AES1 bob') and the outcome will be a failure, same as for `@alice 2`

### Steps to reproduce

Writing tests to reproduce, very difficult to reproduce in normal conditions

### Expected behaviour

When an AES key is used to encrypt, `metadata.sharedKeyEnc` should always contain that same AES key, encrypted with bob's public key 

### Additional context
If this situation is currently encountered, the only solution is to delete from the atServer (e.g. manually via atrepl) `shared_key.bob@alice` and `@bob:shared_key@alice` and restart the `@alice` clients (ideally with new local storage, or if they use sync then the deletions will sync to their local storage)

gkc:
**- What I did**
- fix: Resolves #1506
- Also addresses https://github.com/atsign-foundation/operations/issues/287

**- How I did it**
1) For #1506 : Removed a whole bunch of legacy code for management of "their copy of the shared encryption key" since it is no longer required, and creates race conditions. See #1506 for explanation of the race condition.
Quoting comment from the code
```
  /// In order to mitigate race conditions caused by the soon-to-be-legacy
  /// behaviour of having a single symmetric key, we will always
  /// encrypt the actual symmetric key we are using and set it in the
  /// metadata, rather than storing it to data stores etc. It is safe to do
  /// this because for a long time, clients have been decrypting using the
  /// `sharedKeyEnc` in the metadata, which we are always setting.
```

2) For https://github.com/atsign-foundation/operations/issues/287 : we now fully ignore the deprecated `AtClientPreference.atProtocolEmitted`; its value is no longer used as an `if-then-else` guard

**- How to verify it**
- Tests pass
- For extra confidence, this change [was tested](https://github.com/atsign-foundation/noports/pull/2119) via NoPorts end-to-end tests which cover some pretty old versions of the at_client packages via the older NoPorts releases


gkc:
Blocked by https://github.com/atsign-foundation/at_server/pull/2371

packages/at_client/lib/src/client/at_client_impl.dart
```diff
@@ -551,10 +550,7 @@ class AtClientImpl implements AtClient, AtSignChangeListener {
       atKey.namespace ??= preference?.namespace;
     }
 
-    // ignore: deprecated_member_use_from_same_package
-    if (preference!.atProtocolEmitted >= Version(2, 0, 0)) {
-      atKey.metadata.ivNonce ??= EncryptionUtil.generateIV();
-    }
+    atKey.metadata.ivNonce ??= EncryptionUtil.generateIV();
```
gkc, line 553: we always want IVs

packages/at_client/lib/src/encryption_service/abstract_atkey_encryption.dart
gkc, line 1: This file is where the fix for the rare race condition is.

Xlin123, line 15: future thinking (not for this PR): 

thoughts on using sealed class on `AtKeyEncryption` to eventually help support singleton cleanup & potentially different encryption methods?

Xlin123 (approved):
LGTM

packages/at_client/lib/src/encryption_service/abstract_atkey_encryption.dart
gkc: Yes there's a big refactoring in the future here; we need to move to a plugin model where metadata specifies the name of an encryption/decryption scheme being used for a particular data record or notification - e.g. `atsign.legacy` for what we have now, and `atsign.mls.v1` for when we do our first implementation of [MLS](https://en.wikipedia.org/wiki/Messaging_Layer_Security)

PR #1638, opened 2025-08-27
fix: cleaned up all of the lint in the non-flutter packages

gkc:
**- What I did**
Mostly this was lint from the new `strict_top_level_inference` rule. There are some changes which could technically be classified as breaking changes but really aren't because they would previously have thrown runtime errors if null or invalid type arguments were being passed.

**- How I did it**

**- How to verify it**
Tests pass


packages/at_client/lib/src/client/remote_secondary.dart
```diff
@@ -211,13 +199,6 @@ class RemoteSecondary implements Secondary {
     return false;
   }
 
-  Future<void> _restartCallBack(
-      String command, Function notificationCallBack, String privateKey) async {
-    logger.info('auto restarting monitor');
-    // ignore: deprecated_member_use_from_same_package
-    await monitor(command, notificationCallBack, privateKey);
-  }
-
```
gkc, line 220: Removed the above stuff because it is not part of the Secondary, and was being used only by code which has now been removed from AtClientImpl

packages/at_client/lib/src/client/at_client_impl.dart
```diff
@@ -1013,12 +1013,7 @@ class AtClientImpl implements AtClient, AtSignChangeListener {
   @Deprecated('Use AtClient.notificationService')
   Future<void> startMonitor(String privateKey, Function? notificationCallback,
       {String? regex}) async {
-    var monitorVerbBuilder = MonitorVerbBuilder();
-    if (regex != null) {
-      monitorVerbBuilder.regex = regex;
-    }
-    await _remoteSecondary!.monitor(
-        monitorVerbBuilder.buildCommand(), notificationCallback, privateKey);
+    throw UnimplementedError('AtClient.startMonitor has been deprecated');
```
gkc, line 1016: This has been deprecated for a long long time.

packages/at_commons/lib/src/exception/at_client_exceptions.dart
```diff
@@ -4,68 +4,68 @@ import 'package:at_commons/src/exception/at_exceptions.dart';
 
 class AtClientException extends AtException {
   /// The default constructor to preserve the backward compatibility.
-  AtClientException(errorCode, message) : super(message);
+  AtClientException(dynamic errorCode, String message) : super(message);
```
gkc, line 7: Previously the type for the `message` parameter was implicitly dynamic - but the type in the superclass is explicitly String, so while this could technically be called a breaking change, it really isn't because any previous use which didn't pass a String would have resulted in a runtime exception

packages/at_commons/lib/src/exception/at_exceptions.dart
```diff
@@ -41,157 +41,157 @@ class AtException implements Exception {
 
 /// The class [AtConnectException] and its subclasses represent any issues that prevents an connection to the root or the secondary server
 class AtConnectException extends AtException {
-  AtConnectException(message,
+  AtConnectException(String message,
```
gkc, line 44: Same explanation as above

packages/at_commons/lib/src/exception/at_server_exceptions.dart
```diff
@@ -3,39 +3,39 @@
 import 'package:at_commons/src/exception/at_exceptions.dart';
 
 class AtServerException extends AtException {
-  AtServerException(message) : super(message);
+  AtServerException(String message) : super(message);
```
gkc, line 6: Same explanation as above

gkc:
@cpswan Once this is merged I will publish at_commons 5.6.0 and then go back to noports and remove the dependency override. All other changes in this PR can be pushed out in the next normal release of whatever package the change is in.

cpswan (approved):

PR #1646, opened 2025-09-03
chore: refactor at_auth

Xlin123:
**- What I did**
 - Introduced AtKeys & AtKeysIo to house different logics for fileBased, keychain and SIM based keys
 - Introduced AtBytes type into AtKeys to represent each key instead of Strings
 - Removed at_auth_interface.dart as an entrypoint and opted for factory methods for at_enrollment & at_auth
 - Updated dependent code and at_auth tests
 - Functionality for SIM based keys aren't full implemented yet

**- How I did it**
  
**- How to verify it**
- local tests in the at_auth package are passing as well as additional tests added within the PR
- onboarding & authentication examples are functioning as expected

** - Next Steps**
- finish SIM implementation
- @XavierChanth  discussion around atChops implementing AtBytes? wayyyy to many cases where I'm using `AtBytes.toString` & `AtBytes.fromString`  , where in worse case-scenarios I'm using both.

chore: refactor at_auth


packages/at_auth/lib/src/keys/at_keys_io.dart
```diff
@@ -0,0 +1,153 @@
...
+import 'package:at_chops/at_chops.dart';
+import 'package:at_commons/at_commons.dart';
+import 'package:at_utils/at_utils.dart' show AtSignLogger;
+
+abstract class AtKeysIo {
+  FutureOr<AtKeys> read(String atSign);
+  FutureOr write(String atSign, AtKeys atKeys);
+}
+
+class BaseAtKeysIo extends AtKeysIo {
```
xavierchanth, line 15: You could probably make this a mixin https://dart.dev/language/mixins

xavierchanth: e.g. `mixin BaseAtKeysIo on AtKeysIo`.

Then you can drop the unimplemented overrides.

packages/at_auth/lib/src/keys/at_keys_io_impl.dart
```diff
@@ -0,0 +1,80 @@
...
+    }
+
+    throw AtAuthenticationException('atAuthKeys or encryptedKeysMap is required to read from keychain');
+  }
+
+  @override
+  Future write(String atSign, AtKeys atKeys) {
+    // TODO: implement write
+    /// how are sim keys written?
+    return Future.value();
```
xavierchanth, line 71: Now that I think about it, we can't write keys to a sim... that's the whole point of using a sim... it would be a generation flow. Maybe we can do something like:

```
sealed class AtKeysIo {
  read() {}//...
}

abstract class WrittenAtKeysIo extends AtKeysIo {
  write() {} //...
}

abstract class GeneratedAtKeysIo extends AtKeysIo {
  generate() {} // ...
}
```

We'll have to consider this a little more.

But then I think we can do a switch on every AtKeysIo object we see and it has to be either a `WrittenAtKeysIo` or a `GeneratedAtKeysIo` since `AtKeysIo` can't be extended outside of it's file.

xavierchanth: For now, we can leave it as is, this definitely needs more discussion and deeper understanding.

packages/at_auth/lib/src/onboard/at_onboarding_request.dart
```diff
@@ -5,14 +5,13 @@ class AtOnboardingRequest {
...
-  bool enableEnrollment = false;
+
   String rootDomain = 'root.atsign.org';
   int rootPort = 64;
   String? appName;
   String? deviceName;
 
-  /// public key id if [authMode] is [PkamAuthMode.sim]
+  /// public key id for secure element sim based authentication
   String? publicKeyId;
```
xavierchanth, line 15: If we can somehow move this into the SimAtKeysIo class, maybe as a map by atSign, I think that would be for the best.

xavierchanth: I guess the major consideration is: should there be a unique AtKeysIo instance per atSign or per app? Right now we've worked under the assumption is per-app, but as you gain better insight into the code, I think per atSign is worth considering.

xavierchanth (approved):
Overall looks good, to address this comment:

> @XavierChanth discussion around atChops implementing AtBytes? wayyyy to many cases where I'm using AtBytes.toString & AtBytes.fromString , where in worse case-scenarios I'm using both.

I designed AtBytes to optimize for the hotpath, and the hotpath is in utilizing the keys for encryption/decryption, not in reading and writing, hence why the class is an extension type of Uint8List. We don't want to be base64 decoding every time we do a crypto operation, as those are wayyy more frequent than reading and writing, though, that is what you will see more of.

One of the next big steps we will need to do for this is add at_chops into the workspace, and uptake at_chops 3 changes. This is where that decision will matter, we should also make at_chops use AtBytes under the hood.

xavierchanth:
Feel free to merge and continue in a new PR, since this work is ongoing and we aren't merging to trunk it will be easier for me to review

Xlin123:
> I designed AtBytes to optimize for the hotpath, and the hotpath is in utilizing the keys for encryption/decryption, not in reading and writing, hence why the class is an extension type of Uint8List. We don't want to be base64 decoding every time we do a crypto operation, as those are wayyy more frequent than reading and writing, though, that is what you will see more of.
> 
> One of the next big steps we will need to do for this is add at_chops into the workspace, and uptake at_chops 3 changes. This is where that decision will matter, we should also make at_chops use AtBytes under the hood.

understood. I'm in agreement with the above changes and will be making them in the next PR.

packages/at_auth/lib/src/onboard/at_onboarding_request.dart
Xlin123: To this point, is there a reason why we haven't created an AtSign object? I'd imagine this isn't the only functionality that is per atSign.

I understand it's essentially just a string, but if there isn't a design reason should I potentially look into it?

xavierchanth: I would like to make it a String extension type, however it's gonna be a breaking change (to alot of other signatures). There is an extension on String which provides toAtsign() in at_commons.

PR #1686, opened 2025-10-21
chore: copy certs from atservices/at_secondary_proxy to at_onboarding_cli_functional_tests_proxy

srieteja:
**- What I did**
- Copied certs from atservices/at_secondary_proxy to at_onboarding_cli_functional_tests_proxy

**- How to verify it**
- Functional tests in 'at_onboarding_cli_functional_tests_proxy' should be passing now


srieteja:
Note that cacert.pem was also updated, but it looks like the one in at_secondary_proxy and the functional tests is the same one.

cpswan (approved):

cpswan:
@srieteja this should have been done by the refreshcerts workflow, but that doesn't seem to be doing its job properly

srieteja:
@cpswan I'll take a look at the refreshcerts job too, since I'm already here

PR #1687, opened 2025-10-22
fix: Remove manual cert injection

srieteja:
**- What I did**
- Changed the docker image being used in `at_functional_tests` from dev_env -> vip, to ensure uniformity among all the tests
- Removed the certificates from:  
   1. `at_functional_tests/tests/testData`: No longer in use (no references to this cert were found in the tests)
   2. `at_onboarding_cli_tests_proxy/certs`: The `at_proxyserver` docker image already includes the necessary certs. This manual injection is redundant, and in cases where the local certs in 'at_onboarding_cli_tests_proxy' are expired,  they would override the valid certs in the docker image.
- Removed the refreshcerts job entirely, since no certs are being maintained in `at_client_sdk/tests`anymore
- Removed references to these certs (1 reference found and removed)

**- How I did it**
- Removed the refreshcerts job workflow and all manually injected certificates in `at_client_sdk/tests`
- Updated the Docker Compose files to reflect the same (removed mounting of certs to docker images)

**- How to verify it**
- All the tests should pass in GitHub Actions 

fix: remove manual mounting of certs to docker images in at_client_sdk/tests


srieteja:
Discussed this change on the arch call (22 oct) and got no objections, hence moving forward with the change.

xavierchanth (approved):

PR #1697, opened 2025-11-03
fix: fix the "apkam-flutter" branch being used by NoPorts

gkc:
**- What I did**
- fix: the windows-specific 'store / fetch the data in segments' code now works properly when using more than one biometric store name. Bug was exposed when we started storing enrollment info in the keychain as well as just the list of atSigns-atKeys
- fix: also removed all of the code in KeyChainManager which was related to 'useSharedStorage' as that code was very poorly thought out and buggy
- chore: also removed the 'schema migration' code

**- How I did it**
See commits

**- How to verify it**
Verified in NoPorts PR (will add shortly)

### DIFF
--- a/packages/at_client_mobile/lib/src/keychain_manager.dart
+++ b/packages/at_client_mobile/lib/src/keychain_manager.dart
@@ -30,7 +30,7 @@ class KeyChainManager {
 
   KeyChainManager._internal();
 
-  static bool isWindows = true;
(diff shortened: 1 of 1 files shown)


Xlin123 (approved):
LGTM, verified against npdesktop -> `tags/desktop1.6.2+21`

will take note of these changes and adopt them into `tech-debt-removal`

PR #1706, opened 2025-11-06
fix: do not send nack when rpc request mutex not acquired + prep at_client v3.9.2 for release

srieteja:
**- What I did**
- Prevent a NACK/ACK race when running multiple parallel instances of AtRpc.
- In this case: Having multiple Policy Servers running, and when one of the Policy Servers sends a NACK(because it was unable to acquire the mutex), before the Policy Server that acquired the mutex sends an ACK, the ACK is being ignored as the NACK completed the future.
- Also update at_client version to v3.9.2 preparing it for release.

**- How I did it**
- removed the code that sends the NACK, always ensuring only a single responder

**- How to verify it**
- Tests pass here
- Manually:
   - Run two Policy Servers with the same atsign(e.g. @policy), and pass this atsign to noports daemon(e.g. @daemon)
   - Allow a third atsign (e.g. @np_client) to connect to the daemon through the Policy
   - Try to connect @np_client to the daemon, and this should work

fix: do not send nack when the AtRpc request mutex is not acquired

### DIFF
--- a/packages/at_client/lib/src/rpc/at_rpc.dart
+++ b/packages/at_client/lib/src/rpc/at_rpc.dart
@@ -367,11 +367,8 @@ class AtRpc {
     bool mutexAcquired = enableRequestMutex &&
         await _tryAcquireSessionMutex(requestId, notification.to);
     if (enableRequestMutex && !mutexAcquired) {
-      var message =
-          'Ignoring request: could not acquire mutex for requestId $requestId';
-      // send NACK
-      await sendResponse(notification, request,
-          AtRpcResp.nack(request: request, message: message));
+      logger.shout(
+          'Ignoring request: could not acquire mutex for requestId $requestId');
       return;
     }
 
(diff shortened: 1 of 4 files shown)


packages/at_client/lib/src/rpc/at_rpc.dart
```diff
@@ -367,11 +367,8 @@ class AtRpc {
     bool mutexAcquired = enableRequestMutex &&
         await _tryAcquireSessionMutex(requestId, notification.to);
     if (enableRequestMutex && !mutexAcquired) {
-      var message =
-          'Ignoring request: could not acquire mutex for requestId $requestId';
-      // send NACK
-      await sendResponse(notification, request,
-          AtRpcResp.nack(request: request, message: message));
+      logger.shout(
+          'Ignoring request: could not acquire mutex for requestId $requestId');
```
gkc, line 371: Yep that's it

gkc (approved):

PR #1710, opened 2025-11-10
fix: expose atKeys as method parameter in OnboardingService interface + ready for publish at_onboarding_cli v1.14.2

srieteja:
**- What I did**
- Updated at_onboarding_cli to v1.14.2 and the changelog along with it
- Exposed the atKeysFile method parameter in AtOnboardingCli.enroll() (already available in AtOnboardingServiceImpl.enroll())
- Fixed a log statement after generating the .atKeys file:
    - The method parameter atKeysFile now correctly takes precedence over the path defined in OnboardingPreference.atKeysFilePath.

**- How to verify it**
- Tests pass here

fix: expose atKeys as method parameter in OnboardingService interface + ready for publish at_onboarding_cli v1.14.2


packages/at_client/lib/src/rpc/at_rpc.dart
```diff
@@ -367,8 +367,6 @@ class AtRpc {
     bool mutexAcquired = enableRequestMutex &&
         await _tryAcquireSessionMutex(requestId, notification.to);
     if (enableRequestMutex && !mutexAcquired) {
-      logger.shout(
```
srieteja, line 370: unrelated change: remove redundant log

packages/at_onboarding_cli/lib/src/onboard/at_onboarding_service.dart
```diff
@@ -55,6 +55,7 @@ abstract class AtOnboardingService implements ProgressPublisher {
     String deviceName,
     String otp,
     Map<String, String> namespaces, {
+    File? atKeysFile,
```
srieteja, line 65: Expose the atKeysFile method parameter through AtOnboardingService. Note that this param is already present in AtOnboardingServiceImpl.

packages/at_onboarding_cli/lib/src/onboard/at_onboarding_service_impl.dart
```diff
@@ -610,7 +610,7 @@ class AtOnboardingServiceImpl implements AtOnboardingService {
     await fileWriter.flush();
     await fileWriter.close();
     stdout.writeln(
-        '${chalk.green('[Success]')} Your .atKeys file saved at ${atOnboardingPreference.atKeysFilePath}\n');
```
srieteja, line 613: Use the path of the atKeysFile provided through method parameters rather than the one defined in OnboardingPreference.

xavierchanth (changes requested):
Whats the intent or purpose for the change. What problem does it solve and why?

Reexports create unnecessary coupling between packages, and also make it harder to maintain forward compatibility in some instances.

We're also breaking the whole purpose of the facade pattern by exposing an implementation in addition to its interface.

srieteja:
> Whats the intent or purpose for the change. What problem does it solve and why?
> 
> Reexports create unnecessary coupling between packages, and also make it harder to maintain forward compatibility in some instances.
> 
> We're also breaking the whole purpose of the facade pattern by exposing an implementation in addition to its interface.

The purpose of these exports is to ensure that the new noports binary, currently in package `sshnoports` has the necessary classes/implementations available to approve enrollments.

1. Exposing the EnrollmentServiceImpl was necessary to run a new EnrollmentService. This has now been resolved by using the DefaultAtServiceFactory (which creates and returns a new EnrollmentService). Hence, I will go ahead and remove the EnrollmentServiceImpl export.
2. The exports from at_auth are still required: To perform any operation on the enrollment, I still need to create an `EnrollmentRequestDecision` object, which is only available in at_auth. 
     - There are three ways to address this: 
        - Import at_auth into whichever package is using it. 
        - Export the necessary classes through at_client. 
        - Create an intermediary `approve` method in OnboardingCli that just takes the enrollment_id and approves the enrollment for you.
     - Note that at the end of the day, at_client has become coupled with at_auth when it comes to approving or denying enrollments.
      - Importing at_auth seems to be the cleaner option for now. However, we should ensure that users don't need to import additional packages to use a service in any given package. What do you think @XavierChanth ?

gkc (commented):
I think the underlying problem with this PR is that we have a legacy debt problem of too much stuff in the wrong places.

I'd like to see an overview of what we are trying to achieve, what is blocking us, and then we can figure out what is the refactoring we need to do to get stuff into the right places.

Apart from all of the above, we should avoid exporting implementation classes.

srieteja:
> I think the underlying problem with this PR is that we have a legacy debt problem of too much stuff in the wrong places.
> 
> I'd like to see an overview of what we are trying to achieve, what is blocking us, and then we can figure out what is the refactoring we need to do to get stuff into the right places.
> 
> Apart from all of the above, we should avoid exporting implementation classes.

I'm trying to approve an enrollment in sshnoports. 

I already have an instance of EnrollmentService, and I need to call it's `approve` method. The blocker here is that to do so, we need to import at_auth into `sshnoports`, as the `EnrollmentRequestDecision` class is only defined in at_auth.

gkc:
> > I think the underlying problem with this PR is that we have a legacy debt problem of too much stuff in the wrong places.
> > I'd like to see an overview of what we are trying to achieve, what is blocking us, and then we can figure out what is the refactoring we need to do to get stuff into the right places.
> > Apart from all of the above, we should avoid exporting implementation classes.
> 
> I'm trying to approve an enrollment in sshnoports.
> 
> I already have an instance of EnrollmentService, and I need to call it's `approve` method. The blocker here is that to do so, we need to import at_auth into `sshnoports`, as the `EnrollmentRequestDecision` class is only defined in at_auth.

1. Where is the PR which would consume these changes? Why and how is it using an EnrollmentService from at_client?
2. Why can't you just import at_auth into sshnoports?
3. Is there refactoring you can do to move the relevant parts from at_client to at_auth?
4. Rather than export implementation classes, you should create a factory in the interface / abstract superclass which creates and returns an instance of an implementation class

srieteja:
> > > I think the underlying problem with this PR is that we have a legacy debt problem of too much stuff in the wrong places.
> > > I'd like to see an overview of what we are trying to achieve, what is blocking us, and then we can figure out what is the refactoring we need to do to get stuff into the right places.
> > > Apart from all of the above, we should avoid exporting implementation classes.
> > 
> > 
> > I'm trying to approve an enrollment in sshnoports.
> > I already have an instance of EnrollmentService, and I need to call it's `approve` method. The blocker here is that to do so, we need to import at_auth into `sshnoports`, as the `EnrollmentRequestDecision` class is only defined in at_auth.
> 
> 1. Where is the PR which would consume these changes? Why and how is it using an EnrollmentService from at_client?
> 2. Why can't you just import at_auth into sshnoports?
> 3. Is there refactoring you can do to move the relevant parts from at_client to at_auth?
> 4. Rather than export implementation classes, you should create a factory in the interface / abstract superclass which creates and returns an instance of an implementation class

1. PR link: https://github.com/atsign-foundation/noports/pull/2304/

2. We can import at_auth, and I have done that in the latest commits.

3. Looking forward, we could consider moving fetchEnrollmentRequests out of at_client/enrollment_service_impl so that all enrollment-related implementation exists in a single place. Aside from fetchEnrollmentRequests, the other methods (approve, deny, revoke, etc.) are simply wrappers around the at_auth implementations with no additional functionality.

4. The implementation-class export issue has been resolved by switching to AtServiceFactory.

srieteja:
Updates:
- removed export for impl classes 
- removed exports for at_auth classes from at_client

packages/at_onboarding_cli/lib/src/onboard/at_onboarding_service_impl.dart
gkc: @srieteja (1) What if atKeysFile is passed as null? Shouldn't it fall back to using atOnboardingPreference.atKeysFilePath?

packages/at_onboarding_cli/lib/src/onboard/at_onboarding_service.dart
gkc: @srieteja You need to document how the `atkeysFile` parameter is used

gkc (changes requested):

packages/at_onboarding_cli/lib/src/onboard/at_onboarding_service_impl.dart
srieteja: There's already a fallback in place for when the `atKeysFile` is not passed through method params. The code below handles that:

```
if (atKeysFile == null) {
      ...

      atKeysFile = File(atOnboardingPreference.atKeysFilePath!);
}
```

packages/at_onboarding_cli/lib/src/onboard/at_onboarding_service.dart
srieteja: The documentation has now been added through [e9f200b](https://github.com/atsign-foundation/at_client_sdk/pull/1710/commits/e9f200b91597be474f209c219979c46ff0349789)

packages/at_client/lib/src/rpc/at_rpc.dart
gkc: Please can you replace it with a logger.info rather than removing entirely

gkc (changes requested):
Nearly there, one last thing

packages/at_client/lib/src/rpc/at_rpc.dart
srieteja: This log was removed because there's a similar message already being logged by `_tryAcquireSessionMutex()`, and I felt this was redundant. But I can go ahead and make this an info.logger if you still feel it's necessary @gkc 

The statement in `_tryAcquireSessionMutex`:
> logger.shout('Will not handle request from $atsign; could not acquire mutex $mutexKey');

xavierchanth:
> > Whats the intent or purpose for the change. What problem does it solve and why?
> > Reexports create unnecessary coupling between packages, and also make it harder to maintain forward compatibility in some instances.
> > We're also breaking the whole purpose of the facade pattern by exposing an implementation in addition to its interface.
> 
> The purpose of these exports is to ensure that the new noports binary, currently in package `sshnoports` has the necessary classes/implementations available to approve enrollments.
> 
>     1. Exposing the EnrollmentServiceImpl was necessary to run a new EnrollmentService. This has now been resolved by using the DefaultAtServiceFactory (which creates and returns a new EnrollmentService). Hence, I will go ahead and remove the EnrollmentServiceImpl export.
> 
>     2. The exports from at_auth are still required: To perform any operation on the enrollment, I still need to create an `EnrollmentRequestDecision` object, which is only available in at_auth.
>        
>        * There are three ways to address this:
>          
>          * Import at_auth into whichever package is using it.
>          * Export the necessary classes through at_client.
>          * Create an intermediary `approve` method in OnboardingCli that just takes the enrollment_id and approves the enrollment for you.
>        * Note that at the end of the day, at_client has become coupled with at_auth when it comes to approving or denying enrollments.
>        * Importing at_auth seems to be the cleaner option for now. However, we should ensure that users don't need to import additional packages to use a service in any given package. What do you think @XavierChanth ?

People can import the other APIs from the libraries that expose them if (and only if) they need them, but exporting other packages from your own means that your own package's public API is no longer fully in your control, even if you use the `show` keyword.

And, even if we own both packages, we should not couple our packages for the same reason we should avoid coupling the rest of the software stack. It is why we use the facade pattern, and are replacing the singleton pattern with dependency injection.

Software and packages should be treated as individual units, so that they prioritize composability over ease of use. Generally, the most common use-case is quite straight forward, so even with a more verbose syntax ease of use is still good for most people. But without a focus on composability, any time you stray beyond the assumptions of the simplifications that were made for ease of use, it is like running into a brick wall.

packages/at_client/lib/src/rpc/at_rpc.dart
gkc: OK

gkc (approved):

xavierchanth (approved):

srieteja:
Thank you @gkc, @XavierChanth — this discussion was quite informative.

PR #1712, opened 2025-11-11
chore: New Flutter stable version

atsignbot:
Automated changes by [create-pull-request](https://github.com/peter-evans/create-pull-request) GitHub action


cpswan:
image

The empty change is likely the result of some weirdness happening with the release channels, which might in turn be a glitch related to the 3.10 release.

@XavierChanth you may consider some error checking to avoid this. Also it would be useful to have the release version in a comment.

PR #1729, opened 2025-11-25
build(deps): Updated archive dependency to ^4.0.7

gkc:
**- What I did**
Upgrade dependency on archive package to ^4.0.7

**- How I did it**
With difficulty, muttering curses to myself. See commits.

**- How to verify it**
Tests pass.

Note 1: the archive package is used in a couple of the at_tools packages but not in conjunction with at_client so there is no dependency clash there.

Note 2: Why am I doing this? Because I wanted to use the archive package in `noports` and noticed we're not using latest major version of `archive` here in at_client

### DIFF
--- a/packages/at_client/CHANGELOG.md
+++ b/packages/at_client/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 3.9.3
+
+- build(deps): Updated archive dependency to ^4.0.7
+
 ## 3.9.2
 
 - fix: AtRpc - prevent NACK/ACK race when handling request mutex acquisition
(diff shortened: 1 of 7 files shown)


athandle (approved):

CurtlyCritchlow (approved):
lgtm. Question though, why did you remove archive from at_onboarding_flutter? I did noticed that you placed it in the global pubspec.yaml.

gkc:
> lgtm. Question though, why did you remove archive from at_onboarding_flutter? I did noticed that you placed it in the global pubspec.yaml.

It's there transitively anyway, and having it there explicitly was causing a dependency hell which I could not fix any other way.

PR #1742, opened 2025-12-05
fix(at_client_flutter): keep an option to write to keychain

Xlin123:
**- What I did**
 - if someone wants to authenticate via file & have the keys save in the keychain they will have to auth twice.
 - so we either expose a method to writeToKeychain or provide an option in authenticate()
 -  I provided an optional param in AuthService.authenticate() called backupToKeychain 
**- How I did it**

**- How to verify it**

fix(at_client_flutter): keep an option to write to keychain


packages/at_client_flutter/lib/src/services/auth_service.dart
```diff
@@ -36,11 +36,20 @@ class AuthService {
   /// Authenticate an atSign with secondary server
   ///
   ///   [atAuthRequest] - AtAuthRequest containing atSign, AtRootDomain, AtKeysIo.
+	///
+	///		[backupOnKeychain] -  
+	///
   /// Returns [AtAuthResponse] which contains keys and status of authentication
-  Future<AtAuthResponse> authenticate(AtAuthRequest atAuthRequest) async {
+  Future<AtAuthResponse> authenticate(AtAuthRequest atAuthRequest, {bool backupOnKeychain = false}) async {
```
xavierchanth: I would consider avoiding the idea of introducing flutter specific things... that is what got us here in the first place.

That begs the question of how to make it generic. One possible answer is to introduce an optional `List<AtKeysIo>` to the dart signature which is for additional places to backup the keys during activation.

The idea behind this comment is that this signature stays the same as at_auth's.

Xlin123: [16d3315](https://github.com/atsign-foundation/at_client_sdk/pull/1742/commits/16d3315dc67c1952e6b0045c9ef49e8943ed05bc)

xavierchanth (approved):

PR #1746, opened 2025-12-08
feat: at_client: Add RemoteLocalPref enum and AtClientPreference.remoteLocalPref field

gkc:
**- What I did**
- feat: Add RemoteLocalPref enum and AtClientPreference.remoteLocalPref field to enable apps to easily specify that they want to use the remote atServer by default.
  - Behaviour is identical to existing unless apps specifically set the value of `AtClientPreference.remoteLocalPref` to something other than the default, which is `localOnly`
  - Up to now we've enabled use of remote atServer via the Get/Put/Delete RequestOptions. It is now possible to override the other way. So for example if AtClientPreference.remoteLocalPref is `remoteOnly`, application code can still force use of local by explicitly using a RequestOptions with `useRemoteAtServer = false`
- build: update pubspec, changelog for release 3.10.0

**- How I did it**
- See file diffs. The changes are fairly self-explanatory - basically, an enhancement to the existing `useRemoteAtServer` logic with regards to whether to use localSecondary or remoteSecondary, plus the inevitable little bits of cleanup, plus additional tests to cover the changes

**- How to verify it**
Tests pass

### DIFF
--- a/packages/at_client/lib/src/client/at_client_impl.dart
+++ b/packages/at_client/lib/src/client/at_client_impl.dart
@@ -345,12 +345,8 @@ class AtClientImpl implements AtClient, AtSignChangeListener {
   Future<String?> executeVerb(
       VerbBuilder builder, RemoteLocalPref prefForOp) async {
     switch (prefForOp) {
-      case RemoteLocalPref.localFirst:
+      case RemoteLocalPref.localOnly:
         return await localSecondary!.executeVerb(builder, sync: true);
-      case RemoteLocalPref.remoteFirst:
-        var retVal = await _remoteSecondary!.executeVerb(builder);
-        await localSecondary!.executeVerb(builder);
-        return retVal;
       case RemoteLocalPref.remoteOnly:
         return await _remoteSecondary!.executeVerb(builder);
     }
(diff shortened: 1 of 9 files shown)


packages/at_client/lib/src/client/at_client_impl.dart
```diff
@@ -46,9 +45,9 @@ class AtClientImpl implements AtClient, AtSignChangeListener {
 
   AtClientPreference? get preference => _preference;
   late final String _atSign;
-  String? _namespace;
```
gkc, line 49: Removed this as obsolete

packages/at_client/lib/src/client/at_client_impl.dart
```diff
@@ -46,9 +45,9 @@ class AtClientImpl implements AtClient, AtSignChangeListener {
...
   SecondaryKeyStore? _localSecondaryKeyStore;
-  LocalSecondary? _localSecondary;
+  @visibleForTesting
+  LocalSecondary? localSecondary;
```
gkc, line 50: made visible for testing purposes

packages/at_client/lib/src/client/at_client_impl.dart
```diff
@@ -268,16 +266,9 @@ class AtClientImpl implements AtClient, AtSignChangeListener {
     // }
   }
 
-  Secondary getSecondary() {
-    if (_preference!.isLocalStoreRequired) {
-      return _localSecondary!;
-    }
-    return _remoteSecondary!;
-  }
-
```
gkc, line 277: removed as obsolete

xavierchanth (approved):

cconstab (approved):

PR #1783, opened 2026-01-19
feat: Introduce Collision aware FileAtKeysIo

### LINKED ISSUE #1793: Add collision handling for .atKeys file creation

  When onboarding or enrolling atSigns, the .atKeys file creation fails ungracefully if a file already exists at the target path. Additionally, there's no way for applications to customize the behavior when collisions occur - they can only catch exceptions after the fact.        

Introduce a collision handling system for .atKeys file operations that allows applications to                                                                                                                                                                                                                                                                                                define custom collision behavior through injectable handlers (abort, prompt user, rename, etc.)                                                                                                                                                                                                                                                                    
                                                                                                                                                                                                                  
The solution should integrate with at_auth's WrittenAtKeysIo for seamless onboarding/enrollment flows

srieteja:
Closes https://github.com/atsign-foundation/at_client_sdk/issues/1793

**- What I did**
- Introduced a defensive check in `at_auth` to avoid overwriting at_keys. 
- Introduced collision handlers to manage `atKeys` file conflicts during creation.
- Created `CollisionAwareFileAtKeysIO` that integrates collision handling with `at_auth`'s onboarding flow.
- Set `chmod 600` permission on the created `atKeys` files for security
- Updated `ActivateCli` and `AtOnboardingServiceImpl` to use this collision-aware system
- Removed `_generateAtKeysFile` method that previously used to write atKeys
- Deprecated `allowOverwrite` parameter from multiple methods in AtOnboardingService

**- How I did it**
- Created `CollisionContext` to be able to pass collision-specific details between layers. 
- Created `AtKeysFileCollisionHandlers` class that holds currently two solutions: 

    1. `abortOnCollisionHandler`: that aborts as soon as a collision is encountered
    2. `interactiveConsoleHandler`: that will iteratively prompt the user to provide an alternate path to store the keys file
- Created collision result types (Abort, UseAlternative) that handlers return to indicate how the conflict should be resolved    
- Extended `WrittenAtKeysIO` in `CollisionAwareFileAtKeysIo` that will allow injecting the above collision handling mechanisms into `at_auth`.
- Created `AtFileUtil` that currently holds one method: `setSecureFilePermission` that will:

    1. sets `chmod 600` permissions for atKeys on POSIX systems
    2. Attempts to set equivalent permissions on Windows using `icacls`

 - Added ExceptionMessageFormatter.extractRootCause() to strip nested "Exception:" prefixes from error messages.

**- How to verify it**
- Unit tests have been added to validate the changes
- Existing unit/functional/e2e tests passing should indicate that backwards compatibility is intact
- Manually:
   - Use at_activate to onboard an atsign and ensure an atKeys file is created (let activate_cli use the default path)
   - Now try to create an enrollment for the same atsign  and approve it(again, let activate_cli use the default path)
   - Now, since the enrollment creation is going to cause a collision, you should be prompted to provide an alternate path to store the new atKeys file.
   
feat: Introduce collision handling mechanisms in AtOnboardingCli


packages/at_onboarding_cli/test/at_onboarding_cli_test.dart
```diff
@@ -201,12 +201,18 @@ void main() {
       RSAKeypair encryptionRsaKeyPair = onboardingService.generateRsaKeypair();
       atEnrollmentResponse.atAuthKeys = AtKeys()
         ..enrollmentId = '123'
-        ..defaultSelfEncryptionKey = AtBytes.fromString(onboardingService.generateAESKey())
-        ..defaultEncryptionPublicKey = AtBytes.fromString(encryptionRsaKeyPair.publicKey.toString())
-        ..defaultEncryptionPrivateKey = AtBytes.fromString(encryptionRsaKeyPair.privateKey.toString())
-        ..apkamPrivateKey = AtBytes.fromString(encryptionRsaKeyPair.privateKey.toString())
-        ..apkamPublicKey = AtBytes.fromString(encryptionRsaKeyPair.publicKey.toString())
-        ..apkamSymmetricKey = AtBytes.fromString(onboardingService.generateAESKey());
+        ..defaultSelfEncryptionKey =
```
srieteja, line 204: ignore: dart format

packages/at_onboarding_cli/test/at_onboarding_cli_test.dart
```diff
@@ -263,12 +275,18 @@ void main() {
...
```
srieteja, line 278: ignore: dart format

packages/at_onboarding_cli/test/at_onboarding_cli_test.dart
```diff
@@ -332,30 +566,37 @@ AtOnboardingPreference getOnboardingPreference() {
...
   if (atChopsKeys.atPkamKeyPair?.atPublicKey.publicKey != null) {
-    atAuthKeys.apkamPublicKey = AtBytes.fromString(atChopsKeys.atPkamKeyPair!.atPublicKey.publicKey);
+    atAuthKeys.apkamPublicKey =
+        AtBytes.fromString(atChopsKeys.atPkamKeyPair!.atPublicKey.publicKey);
   }
   if (atChopsKeys.atPkamKeyPair?.atPrivateKey.privateKey != null) {
-    atAuthKeys.apkamPrivateKey = AtBytes.fromString(atChopsKeys.atPkamKeyPair!.atPrivateKey.privateKey);
+    atAuthKeys.apkamPrivateKey =
+        AtBytes.fromString(atChopsKeys.atPkamKeyPair!.atPrivateKey.privateKey);
   }
```
srieteja, line 583: ignore: dart format changes

packages/at_onboarding_cli/lib/src/at_keys/keys_file_writer.dart
gkc, line 1: Why do we have this class as well as at_keys_io_impl ?

gkc (commented):
This is all fine, I guess, but feels a bit over-engineered. The primary goal is to ensure that atKeys files are never overwritten. How a program handles that is up to the program. The "prompt for alternative" etc stuff feels program-specific (upon catching an exception or finding that there is already an atKeys at the path) rather than something that needs to be in the library. However I am open to argument ...

srieteja:
We are moving away from this impl. See https://github.com/atsign-foundation/at_client_sdk/issues/1793 for more context.

PR #1815, opened 2026-02-10
feat: support for resuming enrollments (requester side)

### LINKED ISSUE #1747: Introduce a resume mechanism for enrollments

The current process for creating a new enrollment involves retrying authentication until approval, followed by writing keys to a file once successful authentication occurs. However, if the CLI handling this process is interrupted or terminated before completion, there is no way to recover the enrollment. As a result, a new enrollment must be created, but this cannot be done using the same device_name/app_name as the previous attempt. This limitation poses an issue for users who encounter interruptions during the enrollment process.

One possible solution would be to introduce a mechanism that allows users to resume the authentication retry loop and create keys on successful authentication. This can be achieved by introducing an internal state file that stores the keys and enrollment details as soon as the enrollment is created. The state file would then be maintained until the enrollment is approved, at which point the keys would be written. If the CLI is terminated during the process, upon running the same command again, it could infer all the necessary details from the state file and continue with the remaining steps of the process. This would eliminate the need for creating a new enrollment and provide a more seamless experience for users.

srieteja:
Closes https://github.com/atsign-foundation/at_client_sdk/issues/1747

**- What I did**
- set `chmod 600` for all atKeys files
- `enroll` and `onboard` in `AtOnboardingService` now perform a pre-check to ensure the atKeys file is creatable.
- `OnboardingService.enroll()` saves checkpoints that can be used to resume interrupted `awaitApproval()` flow.
- Introduced `toJson` and `fromJson` in `AtEnrollmentResponse.`

**- How I did it**
- Introduced `AtFileUtil` that has methods:
   - `setSecureFilePermissions`: sets `chmod 600` for the atKeys file on POSIX. Attempts to do the equivalent on Windows using the `icacls` command.
   - `validateFileIsCreatable`: Ensures file does not already exist. Also writes and deletes a dummy file to ensure the provided path is writable.
 - Introduced `EnrollmentCheckpoint` class that holds all the checkpoint read-write ops.
    - The checkpoint file names use a hash of atsign + appname + deviceName + namespaces. This ensures that the checkpoint is created per atsign-params combo. So sending and enrollment requests with a different atsign or enroll params will not cause collisions.
    - The checkpoint is deleted as soon as the enrollment is approved or denied.
    - The checkpoint has an internal expiry: it uses the `apkamKeysExpiry` or has a fallback value of 24 hours.
    - In case the checkpoint is expired or corrupted, that file is deleted.

**- How to verify it**
- Existing tests pass
- Newly added checkpoint-specific tests pass
- Manually: 
   - Try running an enroll command using activate_cli.
   - Interrupt the process while waiting for approval
   - Try running the same command again
   - activate_cli should pick up the checkpoint and resume awaiting approval

feat: set `chmod 600` for all atKeys files
feat: support resuming enrollments

### DIFF
--- a/packages/at_auth/lib/src/enroll/models/at_enrollment_response.dart
+++ b/packages/at_auth/lib/src/enroll/models/at_enrollment_response.dart
@@ -83,4 +83,21 @@ class AtEnrollmentResponse {
   String toString() {
     return 'AtEnrollmentResponse{enrollmentId: $enrollmentId, enrollStatus: $enrollStatus}';
   }
+
+  Map<String, dynamic> toJson() {
+    return {
+      'enrollmentId': enrollmentId,
+      'enrollStatus': enrollStatus.name,
+      if (atSign != null) 'atSign': atSign,
+    };
+  }
+
+  factory AtEnrollmentResponse.fromJson(Map<String, dynamic> json) {
+    String enrollmentId = json['enrollmentId'];
+    EnrollmentStatus enrollmentStatus = EnrollmentStatus.values
+        .firstWhere((es) => es.name == json['enrollStatus']);
+    String? atSign = json['atSign'];
+
+    return AtEnrollmentResponse(enrollmentId, enrollmentStatus, atSign: atSign);
+  }
 }
--- a/packages/at_onboarding_cli/lib/src/cli/auth_cli.dart
+++ b/packages/at_onboarding_cli/lib/src/cli/auth_cli.dart
@@ -337,6 +337,7 @@ Future<int> wrappedMain(List<String> arguments) async {
                 passPhrase: commandArgResults[AuthCliArgs.argNamePassPhrase]));
       case AuthCliCommand.decrypt:
         await passPhraseDecryptAtKeys(commandArgResults);
+
       case AuthCliCommand.version:
         stdout.writeln('Version: $packageVersion');
     }
@@ -488,17 +489,6 @@ Future<bool> enroll(ArgResults argResults, {AtOnboardingService? svc}) async {
     throw ArgumentError('The --${AuthCliArgs.argNameAtKeys} option is'
         ' mandatory for the "enroll" command');
   }
-
-  File f = File(argResults[AuthCliArgs.argNameAtKeys]);
-
-  if (f.existsSync()) {
-    throw StateError('Error: atKeys file ${f.path} already exists');
-  }
-
-  if (!canCreateFile(f)) {
-    throw StateError('Error: Unable to open $f for writing');
-  }
-
   svc ??= createOnboardingService(argResults);
 
   Map<String, String> namespaces = {};
@@ -511,25 +501,23 @@ Future<bool> enroll(ArgResults argResults, {AtOnboardingService? svc}) async {
     namespaces[namespace] = permission;
   }
 
-  // If apkam Keys expiry is not set, then APKAM keys should lives forever.
+  // If apkam Keys expiry is not set, then APKAM keys should live forever.
   // Therefore set to 0ms (0 milliseconds) and TTL will not be set.
   String apkamKeysExpiry = argResults[AuthCliArgs.argNameExpiry] ?? '0ms';
-  AtEnrollmentResponse er = await svc.sendEnrollRequest(
-      argResults[AuthCliArgs.argNameAppName],
-      argResults[AuthCliArgs.argNameDeviceName],
-      argResults[AuthCliArgs.argNamePasscode],
-      namespaces,
-      apkamKeysExpiryDuration: parseDuration(apkamKeysExpiry));
-  stdout.writeln('Enrollment ID: ${er.enrollmentId}');
 
-  stderr.writeln('Waiting for approval; will check every 10 seconds');
-  await svc.awaitApproval(er,
-      retryInterval: AtOnboardingService.defaultApkamRetryInterval,
-      logProgress: true,
-      maxRetries: int.parse(argResults[AuthCliArgs.argNameMaxRetries]));
+  String? atKeysPath = argResults[AuthCliArgs.argNameAtKeys];
+  File? atKeysFile = atKeysPath != null ? File(atKeysPath) : null;
 
-  stderr.writeln('Creating atKeys file');
-  await svc.createAtKeysFile(er, allowOverwrite: false);
+  await svc.enroll(
+    argResults[AuthCliArgs.argNameAppName],
+    argResults[AuthCliArgs.argNameDeviceName],
+    argResults[AuthCliArgs.argNamePasscode],
+    namespaces,
+    atKeysFile: atKeysFile,
+    apkamKeysExpiryDuration: parseDuration(apkamKeysExpiry),
+    maxRetries: int.parse(argResults[AuthCliArgs.argNameMaxRetries]),
+    retryInterval: AtOnboardingService.defaultApkamRetryInterval,
+  );
 
   return true;
 }
@@ -1168,7 +1156,7 @@ AtOnboardingService createOnboardingService(ArgResults ar) {
     if (stdout.hasTerminal && viewableLength > (stdout.terminalColumns - 3)) {
       output = '${output.substring(0, stdout.terminalColumns - 3 + diff)}...';
     }
-    stderr.write('\r\x1b[K$output');
+    stderr.writeln('\r\x1b[K$output');
   });
 
   return impl;
--- a/packages/at_onboarding_cli/lib/src/cli/auth_cli.dart
+++ b/packages/at_onboarding_cli/lib/src/cli/auth_cli.dart
@@ -463,7 +463,6 @@ Future<bool> onboard(ArgResults argResults, {AtOnboardingService? svc}) async {
       maxRetries: int.parse(argResults[AuthCliArgs.argNameMaxRetries]),
       retryInterval: AtOnboardingService.defaultActivationCheckInterval,
     );
-    stderr.writeln();
     return true;
   } catch (e) {
     await Future.delayed(Duration(milliseconds: 10));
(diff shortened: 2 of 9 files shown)


packages/at_auth/lib/src/enroll/models/at_enrollment_response.dart
```diff
@@ -46,4 +46,26 @@ class AtEnrollmentResponse {
   String toString() {
     return 'AtEnrollmentResponse{enrollmentId: $enrollmentId, enrollStatus: $enrollStatus}';
   }
+
+  Map<String, dynamic> toJson() {
+    return {
+      'enrollmentId': enrollmentId,
+      'enrollStatus': enrollStatus.name,
+      if (atSign != null) 'atSign': atSign,
+      if (atAuthKeys != null) 'atAuthKeys': atAuthKeys!.toJson(),
```
srieteja: should not serialize/de-serialize atAuthKeys

packages/at_onboarding_cli/lib/src/cli/auth_cli.dart
```diff
@@ -488,17 +488,6 @@ Future<bool> enroll(ArgResults argResults, {AtOnboardingService? svc}) async {
     throw ArgumentError('The --${AuthCliArgs.argNameAtKeys} option is'
         ' mandatory for the "enroll" command');
   }
```
srieteja, line 490: This pre-check is now handled in OnboardingService.enroll()

packages/at_onboarding_cli/lib/src/cli/auth_cli.dart
```diff
@@ -514,22 +503,20 @@ Future<bool> enroll(ArgResults argResults, {AtOnboardingService? svc}) async {
   // If apkam Keys expiry is not set, then APKAM keys should lives forever.
   // Therefore set to 0ms (0 milliseconds) and TTL will not be set.
   String apkamKeysExpiry = argResults[AuthCliArgs.argNameExpiry] ?? '0ms';
```
srieteja, line 505: `auth_cli` should use `OnboardingService.enroll()` to make use of the new checkpoint impl

packages/at_onboarding_cli/lib/src/onboard/at_onboarding_service_impl.dart
```diff
@@ -41,8 +43,12 @@ class AtOnboardingServiceImpl implements AtOnboardingService {
 
   AtEnrollment? _atEnrollment;
```
srieteja, line 45: ignore: format only

packages/at_onboarding_cli/lib/src/onboard/helpers/enrollment_checkpoint.dart
```diff
@@ -0,0 +1,111 @@
...
+    AtEnrollmentResponse er,
+    String appName,
+    String deviceName,
+    Map<String, String> namespaces, {
+    Duration? expiry,
+  }) async {
+    expiry ??= defaultCheckpointExpiry;
+
+    final Map<String, dynamic> json = er.toJson();
+    json.remove('atSign'); // do not reveal which atSign this checkpoint belongs to
```
gkc, line 65: Why not reveal which atSign it belongs to?

packages/at_onboarding_cli/lib/src/onboard/helpers/enrollment_checkpoint.dart
```diff
@@ -0,0 +1,111 @@
...
+  }
+
+  @visibleForTesting
+  File getFile(String appName, String deviceName,
+          Map<String, String> namespaces) =>
+      File(path.join(Directory.current.path,
+          '${_paramsHash(_atSign, appName, deviceName, namespaces)}.enrollment.checkpoint'));
+
+  /// Writes the checkpoint to disk with secure file permissions (chmod 600).
+  /// Any existing checkpoint for different params is deleted first.
```
gkc: Why removing other checkpoint files?

gkc: Also I don't see where the checkpoint file is being removed

packages/at_onboarding_cli/lib/src/onboard/at_onboarding_service_impl.dart
```diff
@@ -216,6 +223,9 @@ class AtOnboardingServiceImpl implements AtOnboardingService {
     await atAuth!.completeActivation();
   }
 
+  @visibleForTesting
+  late EnrollmentCheckpoint enrollCheckpoint = EnrollmentCheckpoint(_atSign);
```
gkc: It's not late if you're initializing it here

gkc: If you want to create it at construction, add into the constructor function

packages/at_onboarding_cli/lib/src/onboard/at_onboarding_service_impl.dart
```diff
@@ -490,6 +534,9 @@ class AtOnboardingServiceImpl implements AtOnboardingService {
   }
 
   /// Retries PKAM auth until an enrollment is approved/denied/expired
+  ///
+  /// This method is no longer needed here. Use AtAuthImpl.validateAtServer()
+  /// ToDo: Marking this for removal
```
gkc: Why not just remove it?

packages/at_onboarding_cli/test/at_onboarding_cli_test.dart
```diff
@@ -275,6 +291,11 @@ void main() {
 
       File file = File(f.path);
       expect(file.existsSync(), true);
+      if (!Platform.isWindows) {
+        final stat = await FileStat.stat(file.path);
+        expect(stat.mode & 0x1FF, equals(0x180),
+            reason: 'atKeys file must have chmod 600 (owner read/write only)');
+      }
```
gkc: Is there any way to check on Windows platform?

srieteja: I have added some logic to be able to check that on Windows through [56a0678](https://github.com/atsign-foundation/at_client_sdk/pull/1815/commits/56a0678b83764c9b3607d46e7cced142363b2dbe). But, note that I haven't tested this on a windows yet. I will be able to do that later today.

packages/at_onboarding_cli/lib/src/onboard/helpers/enrollment_checkpoint.dart
srieteja: That is a stale doc belonging to an intermittent impl. Now updated through [e271512](https://github.com/atsign-foundation/at_client_sdk/pull/1815/commits/e271512283437439ddf46964d70f1794ab8ec7de).

packages/at_onboarding_cli/lib/src/onboard/helpers/enrollment_checkpoint.dart
srieteja: This way, even if the checkpoint file (containing the atAuthKeys) is compromised, it won't reveal which atsign those keys belong to. Just a tad more secure.

packages/at_onboarding_cli/lib/src/onboard/at_onboarding_service_impl.dart
srieteja: > It's not late if you're initializing it here

That's true, but combining late with an initialization enables a lazy init. It won't instantiate `EnrollCheckpoint` until the first time it's called. Also, if I remove late, the compiler throws an error, as `_atSign` is only available after the constructor body runs.

packages/at_onboarding_cli/lib/src/onboard/at_onboarding_service_impl.dart
srieteja: I tagged the wrong method. It was actually _waitTillSecondaryCreated that was supposed to be removed. I have created a PR to do the same: https://github.com/atsign-foundation/at_client_sdk/pull/1860.

packages/at_onboarding_cli/lib/src/cli/auth_cli.dart
```diff
@@ -511,73 +500,27 @@ Future<bool> enroll(ArgResults argResults, {AtOnboardingService? svc}) async {
...
+    argResults[AuthCliArgs.argNamePasscode],
+    namespaces,
+    atKeysFile: atKeysFile,
+    apkamKeysExpiryDuration: parseDuration(apkamKeysExpiry),
+    maxRetries: int.parse(argResults[AuthCliArgs.argNameMaxRetries]),
+    retryInterval: AtOnboardingService.defaultApkamRetryInterval,
+  );
 
   return true;
 }
```
srieteja, line 523: This impl has been moved to AtFileUtil.ensureWritable()

packages/at_onboarding_cli/test/at_onboarding_cli_test.dart
srieteja: Update: I tested this on Windows, and it works. The tests pass, too.

packages/at_onboarding_cli/lib/src/onboard/at_onboarding_service_impl.dart
gkc: Please move this initialization into the constructor body ... it's then clearer to a casual reader what is happening.

srieteja: I have moved the init to the constructor through [cac70d7](https://github.com/atsign-foundation/at_client_sdk/pull/1815/commits/cac70d795fcd7074efe20fcaac8ffd3a02802084).

packages/at_onboarding_cli/lib/src/onboard/at_onboarding_service_impl.dart
gkc: You could make this `final Atsign atSign` and then use `this.atSign` in the constructor. Constructor users will pass Atsigns - to make an Atsign from a String you use `Atsign alice = 'alice'.toAtsign()`

packages/at_onboarding_cli/test/at_onboarding_cli_test.dart
gkc: Great!

gkc (changes requested):

packages/at_onboarding_cli/lib/src/onboard/at_onboarding_service_impl.dart
srieteja: But wouldn't that break stuff? What we can do instead is make the `final Atsign atSign` change and handle that through the constructor, like
```
AtOnboardingServiceImpl(
    String atsign,
    this.atOnboardingPreference, {
    this.atServiceFactory,
    String? enrollmentId,
  }): atSign = atsign.toAtsign() {
```

What do you think @gkc?

gkc: Yes that's fine for now (because changing the type does cause a ripple effect) but please take opportunity as you're making changes to move to the `Atsign` type

srieteja: I will keep that in mind.

srieteja: Changed the atSign type from String -> Atsign in [08b9eb2](https://github.com/atsign-foundation/at_client_sdk/pull/1815/commits/08b9eb2feff91890ec0c578b8fb830bddb1855c0).

gkc (approved):

PR #1839, opened 2026-02-23
fix: monitor delayed start is cancelled when stopAllSubscriptions is called (NotificationServiceImpl)

### LINKED ISSUE #1838: Zombie monitor start after atsign switch due to un-cancellable delayed start
### Describe the bug

  NotificationService.subscribe() delays the monitor start by 30 seconds for statsNotification           
  subscriptions. However, the delay uses Future.delayed(), which cannot be cancelled.                    

  When switching atSigns via AtClientManager, the previous atSign's notification service is
  closed, but its delayed monitor start still fires, creating background monitors that consume resources
  and may interfere with the new atSign's notifications.

### Steps to reproduce

  1. Service calls subscribe(regex: 'statsNotification') → 30-second delay starts
  2. Before the delay completes, stopListening() is called (e.g., during atSign switch)
  3. Monitor stops as expected
  4. 30 seconds after step 1: The uncancellable callback fires and calls startListening()
  5. Result: Zombie monitor starts for the OLD atSign, even though the service was already closed

### Expected behavior

The old monitor should not run/start in the background after switching atsigns.

srieteja:
Closes https://github.com/atsign-foundation/at_client_sdk/issues/1838

**- What I did**
- When stopAllSubscriptions() is called on NotificationServiceImpl, cancel the delayed start monitor call

**- How I did it**
- Introduced a timer to store the delayed start call 
- Cancel the timer when trying to stop the NotificationService

**- How to verify it**

fix: monitor delayed start is cancelled when stopAllSubscriptions is called (NotificationServiceImpl)

### DIFF
--- a/packages/at_client/lib/src/service/notification_service_impl.dart
+++ b/packages/at_client/lib/src/service/notification_service_impl.dart
@@ -377,7 +377,12 @@ class NotificationServiceImpl extends NotificationService
     return status;
   }
 
+  @visibleForTesting
   Timer? delayedStartListeningTimer;
+
+  @visibleForTesting
+  Duration? delayedStartListeningTimerDuration;
+
   @override
   Stream<AtNotification> subscribe(
       {String? regex, bool shouldDecrypt = false}) {
@@ -420,8 +425,9 @@ class NotificationServiceImpl extends NotificationService
     // much better clear control over the notification listening lifecycle.
     if (atClient.getPreferences()?.monitorAutoStart == true) {
       if (regex == 'statsNotification') {
+        delayedStartListeningTimerDuration ??= Duration(seconds: 30);
         delayedStartListeningTimer =
-            Timer(Duration(seconds: 30), startListening);
+            Timer(delayedStartListeningTimerDuration!, startListening);
       } else {
         startListening();
       }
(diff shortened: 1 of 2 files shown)


gkc:
looks good @srieteja ... for testing purposes you may wish to make the Duration customizable (but marked `@visibleForTesting`) ... e.g. in a unit test you could do something like
- set `delayedStartListeningDuration` (which should also be marked as `@visibleForTesting`) to `Duration(milliseconds:10)`
- simulate the sequence of events
  - subscribe to `statsNotification`
  - verify that the `delayStartListeningTimer` is not null
  - verify that the monitor targetState is still `notConnected`
  - call `stopAllSubscriptions(stopNotificationsListener: true)` 
  - verify that the monitor targetState is still `notConnected`
  - verify that the `delayStartListeningTimer` is null
  - pause for 20 milliseconds
  - verify that the monitor targetState is still `notConnected`

gkc (changes requested):
Please add the additional verification as described in my comment @srieteja ... the real proof is that after the delay period, the monitor isn't then started.

srieteja:
> Please add the additional verification as described in my comment @srieteja ... the real proof is that after the delay period, the monitor isn't then started.

@gkc the test has been added through [7eb85c6](https://github.com/atsign-foundation/at_client_sdk/pull/1839/commits/7eb85c6e6f25004b69af694d768747985bbe97c2)

gkc (approved):

PR #1845, opened 2026-02-25
fix(ai): automatically prepend @ symbol into text box

JeremyTubongbanua:
**- What I did**
- Use `at_utils` `.fixAtSign` function
- The "typable dropdown" widget now accepts a prop called "valueNormalizer" which will take the input, then adjust it depending on the functions implementation. Upon normalizing the value, onChanged is then called with the normalized value.
- The dialog defines the "valueNormalizer" to be `_normalizeAtSignInput` which will adjust the string upon every input change which guarantees the @ symbol to always be the first character in the atSign string.
- Alongside that, the `_handleSubmit` function will call `_normalizedAtSignForSubmit` which double guarantees that an @ symbol is present.

**- How I did it**
- Using Codex gpt 5.3 medium

**- How to verify it**
- See comment: https://github.com/atsign-foundation/at_client_sdk/pull/1845#issuecomment-3961228201
- I've also manually tested:
  1. Upload via .atKeys with no keychain keys
  2. Upload via .atKeys with keychain keys present
  3. Switch atSign
  4. Clearing keychain then re-adding
  5. Exporting the atKeys then using those keys to reauthenticate

fix(ai): automatically prepend @ symbol into text box

### DIFF
--- a/packages/at_client_flutter/lib/src/widgets/shared/typable_dropdown.dart
+++ b/packages/at_client_flutter/lib/src/widgets/shared/typable_dropdown.dart
@@ -80,8 +80,7 @@ class TypableDropdown extends StatelessWidget {
         return TextField(
           controller: controller,
           focusNode: focusNode,
-          decoration:
-              decoration ??
+          decoration: decoration ??
               InputDecoration(
                 hintText: hintText ?? 'Type to search...',
                 hintStyle: TextStyle(fontSize: 12),
--- a/packages/at_client_flutter/lib/src/widgets/shared/typable_dropdown.dart
+++ b/packages/at_client_flutter/lib/src/widgets/shared/typable_dropdown.dart
@@ -5,6 +5,7 @@ class TypableDropdown extends StatelessWidget {
   final List<String> items;
   final String? initialValue;
   final ValueChanged<String?>? onChanged;
+  final String Function(String)? valueNormalizer;
   final String? hintText;
   final InputDecoration? decoration;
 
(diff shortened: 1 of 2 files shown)


JeremyTubongbanua:
The text box guarantees the @ symbol is prepended.

https://github.com/user-attachments/assets/d0979fb7-b4a5-438b-a99e-feda48e8c508

packages/at_client_flutter/lib/src/widgets/shared/typable_dropdown.dart
```diff
@@ -5,6 +5,7 @@ class TypableDropdown extends StatelessWidget {
   final List<String> items;
   final String? initialValue;
   final ValueChanged<String?>? onChanged;
+  final String Function(String)? valueNormalizer;
```
Xlin123, line 8: is this normalizing the value on change / selection?

JeremyTubongbanua: Yes

```
onChanged: (value) {
  var normalized = valueNormalizer?.call(value) ?? value;
  if (normalized != value) {
    controller.value = TextEditingValue(
      text: normalized,
      selection: TextSelection.collapsed(offset: normalized.length),
    );
  }
  onChanged?.call(normalized.isEmpty ? null : normalized);
},
```

Xlin123: should we do both or just normalize on submission?

JeremyTubongbanua: We do both

1. We normalize as they are changing the text
2. Then do a one last normalize before submission

First gives a good user experience
Second guarantees it is functional

Xlin123 (approved):

PR #1842, opened 2026-02-23
feat(at_auth): improve at_auth tests

Xlin123:
**- What I did**
 - at_auth tests have been improved focusing on using `AtServerStatus`
 - introduced tests for `AtKeys` behaviours
 - added at_auth into at_libraries workflow for CI
   - [see action](https://github.com/atsign-foundation/at_client_sdk/actions/runs/22444744163/job/64996062443?pr=1842)
 - ran dart format 
 
**- How I did it**
 - AtAuthImpl now contains an injectable AtServerStatus that's visibleForTesting
   - allows for better more clearly defined state for our test cases
 - Added tests for AtKeys regarding json/atChops conversions
 
**- How to verify it**
 - run at_auth unit tests locally
 
**- Description for the changelog**
feat(at_auth): improve at_auth tests

PR #1847, opened 2026-02-26
feat(at_auth): introduce all v4 registrar endpoints

Xlin123:
**- What I did**
 - implemented v4 registrar endpoints into at_auth
 
**- How I did it**
 - claude and then Srie and I combed through all the code to double check functionality lined up with our registrar description
 
**- How to verify it**
 - tests pass (including the new ones)
 
feat(at_auth): introduce all v4 registrar endpoints


packages/at_auth/lib/src/registrar/registrar.dart
```diff
@@ -1,26 +1,113 @@
...
+  final HttpMethod method;
+  const RegistrarApiEndpoint(this.path, this.method);
+}
+
+/// Legacy enum name for backward compatibility
+@Deprecated('Use RegistrarApiEndpoint instead')
+typedef ActivateApiEndpoint = RegistrarApiEndpoint;
+
+/// Backward compatibility extension for legacy enum values
+extension ActivateApiEndpointLegacy on RegistrarApiEndpoint {
```
Xlin123, line 28: we renamed the enum for endpoints, so I'll leave this here for now

Xlin123:
@tinashe404 waiting for #1842 to merge to ensure we didn't break anything

packages/at_auth/lib/src/registrar/registrar_service.dart
```diff
@@ -59,24 +78,167 @@ class RegistrarService implements Registrar {
...
     }
     var payload = jsonDecode(res.body);
     if (payload["message"] != "Verified") {
-      // The toString is for typesafety & to prevent unexpected crashes
       throw Exception('Verification failed: ${payload["message"].toString()}');
     }
-    String cramkey = payload["cramkey"]?.split(':').last ?? '';
-    return cramkey;
+
+    String? cramKey = payload["cramkey"]!.split(':').last;
```
srieteja: `payload["cramkey"]!` will throw if the API response doesn't contain a cram key. Maybe use ? instead of !

packages/at_auth/lib/src/registrar/registrar_service.dart
```diff
@@ -59,24 +78,167 @@ class RegistrarService implements Registrar {
...
-      'atsign': atSign,
-      'otp': otp,
-    });
+    var res = await registrarApiRequest(
+      RegistrarApiEndpoint.validateOtp,
+      {'atsign': atSign, 'otp': otp},
+    );
     if (res.statusCode != 200) {
       throw Exception(
           'Failed to verify activation: ${res.reasonPhrase} - ${res.body}');
```
srieteja: I don't really like the idea of having err body in the exception mesage. Should we leave the reason phrase in and log the error body separately ?

packages/at_auth/lib/src/registrar/registrar_service.dart
```diff
@@ -59,24 +78,167 @@ class RegistrarService implements Registrar {
...
+      }
+    }
+    return atSigns;
+  }
+
+  @override
+  Future<String> getFreeAtSignByCategory(List<String> categories) async {
+    Uri url = Uri.https(registrarUrl,
+        "$apiBase${RegistrarApiEndpoint.getFreeAtsignByCategory.path}");
+
```
srieteja, line 135: Why isn't this using the `registrarApiRequest()` here?

packages/at_auth/lib/src/registrar/registrar_service.dart
```diff
@@ -14,40 +14,59 @@ class RegistrarService implements Registrar {
...
   Future<bool> sendActivationOtp(String atsign) async {
-    var res = await registrarApiRequest(ActivateApiEndpoint.login, {
-      'atsign': atsign,
-    });
+    var res = await registrarApiRequest(
+      RegistrarApiEndpoint.requestOtp,
+      {'atsign': atsign},
+    );
     if (res.statusCode != 200) {
       return false;
```
srieteja, line 74: We always throw when statusCode != 200, in the other methods. Why return false here ?

packages/at_auth/lib/src/registrar/registrar_service.dart
```diff
@@ -59,24 +78,167 @@ class RegistrarService implements Registrar {
...
+    );
+
+    if (res.statusCode != 200) {
+      throw Exception(
+          'Failed to get free Atsign by category: ${res.reasonPhrase} - ${res.body}');
+    }
+    var payload = jsonDecode(res.body);
+    if (payload["data"] == null) {
+      throw Exception(
+          'Failed to get free Atsign by category: payload data is null');
```
srieteja, line 153: It might be worth logging the error message from the response here.

packages/at_auth/lib/src/registrar/registrar_service.dart
```diff
@@ -59,24 +78,167 @@ class RegistrarService implements Registrar {
...
+      },
+    );
+    if (res.statusCode != 200) {
+      throw Exception(
+          'Failed to validate person: ${res.reasonPhrase} - ${res.body}');
+    }
+    var payload = jsonDecode(res.body);
+
+    // Check if validation was successful and return appropriate data
+    if (payload["success"] == true) {
```
srieteja: payload["success"] is not always guaranteed to be present in the response.

packages/at_auth/lib/src/registrar/registrar_service.dart
```diff
@@ -59,24 +78,167 @@ class RegistrarService implements Registrar {
...
+      // New user - return cramkey
+      return {
+        'success': true,
+        'cramkey': payload["cramkey"]?.split(':').last ?? '',
+      };
+    } else if (payload["data"] != null) {
+      // Existing user - return list of atSigns
```
srieteja, line 238: Case: when the max atsign limit for an atsign is reached. The current code would just return the atsigns and not display the error message. See API docs, page 8.

srieteja (changes requested):

packages/at_auth/lib/src/registrar/registrar_service.dart
```diff
@@ -59,24 +78,167 @@ class RegistrarService implements Registrar {
...
+      return {
+        'atsigns': payload["data"]["atsigns"],
+        'newAtsign': payload["data"]["newAtsign"],
+      };
```
srieteja: If there needs to be a call to the same method with confirmation set to true, that should be communicated, either through method docs or a log.

packages/at_auth/lib/src/registrar/registrar_service.dart
```diff
@@ -59,24 +78,167 @@ class RegistrarService implements Registrar {
...
+    if (payload["success"] == true) {
+      return payload["data"]["atsign"];
+    }
+    throw Exception(
+        'Failed to get free Atsign: ${payload["message"] ?? "Unknown error"}');
+  }
+
+  /// Returns a `Future<List<String>>` containing free available atSigns of count provided as input.
+  Future<List<String>> getFreeAtSigns({
+    required int amount,
```
srieteja: Should we add this to the interface?

Xlin123: can we actually just remove this, I think the user can do this themselves if they need

packages/at_auth/lib/src/registrar/registrar_service.dart
Xlin123: im guessing because of the list<String> 

our data map for that is Map<String, String>

i can refactor the registrarApiRequest for it if you'd like

packages/at_auth/lib/src/registrar/registrar_service.dart
srieteja: Fair enough. Let's do that.

packages/at_auth/lib/src/registrar/registrar_service.dart
Xlin123: it'll be a breaking change if I change it, I'll make a todo.

packages/at_auth/lib/src/registrar/registrar_service.dart
Xlin123: payload probably won't have an error message at this point, these are 200 RES that for some reason don't have a data key (shouldnt be possible)

packages/at_auth/lib/src/registrar/registrar_service.dart
Xlin123: [22ea1db](https://github.com/atsign-foundation/at_client_sdk/pull/1847/commits/22ea1db089f52670353c6483ec92a7cae94d86ca)

packages/at_auth/lib/src/registrar/registrar_service.dart
Xlin123: [22ea1db](https://github.com/atsign-foundation/at_client_sdk/pull/1847/commits/22ea1db089f52670353c6483ec92a7cae94d86ca)

packages/at_auth/lib/src/registrar/registrar_service.dart
Xlin123: [22ea1db](https://github.com/atsign-foundation/at_client_sdk/pull/1847/commits/22ea1db089f52670353c6483ec92a7cae94d86ca)

packages/at_auth/lib/src/registrar/registrar_service.dart
Xlin123: [22ea1db](https://github.com/atsign-foundation/at_client_sdk/pull/1847/commits/22ea1db089f52670353c6483ec92a7cae94d86ca)

packages/at_auth/lib/src/registrar/registrar_service.dart
Xlin123: [22ea1db](https://github.com/atsign-foundation/at_client_sdk/pull/1847/commits/22ea1db089f52670353c6483ec92a7cae94d86ca)

packages/at_auth/lib/src/registrar/registrar_service.dart
srieteja: Fair enough, let's leve this be for now.

packages/at_auth/lib/src/registrar/registrar_service.dart
```diff
@@ -59,24 +81,177 @@ class RegistrarService implements Registrar {
...
+  }
+
+  /// validates A Person through atSign, an attached email and otp from the authentication request.
+  ///
+  /// For a new user, returns cramkey
+  /// returns: {
+  ///		'success':
+  ///   'cramkey':
+  /// }
+  /// For an existing user, returns existing atsigns, and the new one
```
srieteja, line 199: Conflicting docs. There's some in the interface and some here. This might override the interface docs.

packages/at_auth/lib/src/registrar/registrar.dart
```diff
@@ -1,26 +1,113 @@
...
+    required String email,
+    String? oldEmail,
+  });
+
+  /// Validates the person registration using the OTP
+  ///
+  /// [atSign] - The atSign being validated
+  /// [email] - Email address used for registration
+  /// [otp] - 4-character OTP sent to email
+  /// [confirmation] - Set to true if validating previously validated atSign
```
srieteja, line 102: nit: Could we add a note that says "if you have existing atsigns, you need to re-send the same request with confirmation set to true".

packages/at_auth/lib/src/registrar/registrar_service.dart
Xlin123: [7926d02](https://github.com/atsign-foundation/at_client_sdk/pull/1847/commits/7926d02da78bf75e8e0b8db883cdff48928ffb9a)

packages/at_auth/lib/src/registrar/registrar.dart
Xlin123: [7926d02](https://github.com/atsign-foundation/at_client_sdk/pull/1847/commits/7926d02da78bf75e8e0b8db883cdff48928ffb9a)

srieteja (approved):

PR #1859, opened 2026-03-10
fix: ensure the subservices can be garbage collected after client is stopped

srieteja:
**- What I did**
- AtClient.stop() now closes enrollment service (re-assigned by AtClientManager when resuming)
- Ensure the closed sub-service can be garbage collected
   - After closing the sub-services, set them to null (NotificationService, SyncService, EnrollmentService)
   - Attach the above services to a finalizer so that whenever they get garbage collected, it gets logged
- Make AtClient.syncService nullable
   - This allows assigning null so that it can be garbage collected
   - The getters defined in the interface expect that the SyncService is non-nullable, so when the client is stopped, it throws
- atLookup.close() doesn't try to close the connection when the connection is null

**- How to verify it**
- Tests should pass
 
fix: ensure the subservices can be garbage collected after client is stopped


packages/at_client/lib/src/client/at_client_impl.dart
```diff
@@ -89,28 +89,40 @@ class AtClientImpl implements AtClient {
...
   @override
   set syncService(SyncService syncService) {
     _syncService = syncService;
+    _finalizer.attach(_syncService!, 'SyncService for $_atSign');
   }
 
   @override
-  SyncService get syncService => _syncService;
+  SyncService get syncService {
+    if (_isStopped) throw StateError('AtClient for $_atSign has been stopped');
```
gkc: Shouldn't the above logic be added to the notificationService and enrollmentService getters as well?

packages/at_client/lib/src/client/at_client_impl.dart
```diff
@@ -269,17 +285,21 @@ class AtClientImpl implements AtClient {
...
     }
 
     try {
       await (notificationService as NotificationServiceImpl).stop();
+      _notificationService = null;
     } catch (e) {
       _logger.warning('Error while closing notification service: $e');
     }
 
+    _enrollmentService = null;
```
gkc: Is there no `stop()` on enrollmentService?

packages/at_client/lib/src/client/at_client_impl.dart
srieteja: Yes. It has now been added through [2506506](https://github.com/atsign-foundation/at_client_sdk/pull/1859/commits/2506506f343ed17a53ce8d5591382f26ac0d5d3f).

packages/at_client/lib/src/client/at_client_impl.dart
srieteja: Yes. EnrollmentService is stateless, so no stop() needed. That makes me think we don't really need to set it to null. What do you think @gkc ?

gkc: Best to keep things consistent I think

gkc:
@srieteja re the failures in the functional tests - I'd suggest two changes to handleNotificationReceipt
(1) Check isStopped and ignore the notification if it is stopped
(2) Modify the catch from `} on Exception catch (e) {` which doesn't catch the StateError to just `} catch (e) {`

packages/at_client/lib/src/service/notification_service_impl.dart
```diff
@@ -224,12 +229,16 @@ class NotificationServiceImpl extends NotificationService {
...
+          } catch (e) {
+            logger.severe('Caught: while processing notificationJson $e');
           }
         });
       }
     } on Exception catch (e) {
       logger.severe('unexpected error:${e.toString()}'
           ' while processing notificationJson: $notificationJSON');
+    } catch (e) {
+      logger.severe('handleNotificationReceipt() - Caught: $e');
```
gkc: Please just have one catch block, we don't care what sort of thing was thrown

srieteja: removed through [f162fed](https://github.com/atsign-foundation/at_client_sdk/pull/1859/commits/f162fedc24c4ff7821719f89bc48b6a072242239).

gkc (approved):

PR #1860, opened 2026-03-10
feat: added pre-auth connectivity probe and progress events to `AtAuthImpl.validateAtServer()`

srieteja:
**- What I did**
- Removed _waitTillSecondaryCreated from AtOnboardingServiceImpl. This is now a duplicate of AtAuthImpl.validateAtServer()
- Added the missing progress events in AtAuthImpl.validateAtServer()
- OnboardingServiceImpl now subscribes to AtAuth's progress stream

**- How to verify it**
- Tests pass here
- `activate_cli onboard` should work without any issues


packages/at_auth/lib/src/at_auth_impl.dart
```diff
@@ -165,14 +166,18 @@ class AtAuthImpl implements AtAuth {
       atOnboardingRequest.atKeys = await atOnboardingRequest.atKeysIo?.read(
         atOnboardingRequest.atSign,
       );
```
srieteja, line 174: Potential exception from `atKeysIo.read()` should be swallowed. `AtAuthenticationException` should not be swallowed.

packages/at_auth/lib/src/at_auth_impl.dart
```diff
@@ -371,7 +376,13 @@ class AtAuthImpl implements AtAuth {
     int retryCount = 0;
 
     while (retryCount < atRequest.retryOptions.maxRetries) {
```
srieteja: added progress events while finding and connecting to secondary server

packages/at_auth/lib/src/at_auth_impl.dart
```diff
@@ -417,16 +428,46 @@ class AtAuthImpl implements AtAuth {
           }
         }
```
srieteja, line 446: Moved: pre-auth secondary probe logic from `AtOnboardingService._waitTillSecondaryCreated()`.

packages/at_onboarding_cli/lib/src/onboard/at_onboarding_service_impl.dart
```diff
@@ -956,8 +867,18 @@ class AtOnboardingServiceImpl implements AtOnboardingService {
   @Deprecated('AtChops will be created in AtAuth')
   AtChops? atChops;
 
+  AtAuth? _atAuth;
+
   @override
-  AtAuth? atAuth;
+  AtAuth? get atAuth => _atAuth;
+
```
srieteja, line 875: Forward atAuth progress events to the atOnboardingCli stream controller.

packages/at_onboarding_cli/test/at_onboarding_cli_test.dart
```diff
@@ -40,13 +40,15 @@ void main() {
     reset(mockAtLookup);
     reset(mockAtAuth);
     registerFallbackValue(FakeAtAuthRequest());
```
srieteja, line 42: Added mocking to ensure tests pass with the atAuth stream being forwarded to onboarding_cli.

packages/at_auth/lib/src/at_auth_impl.dart
```diff
@@ -41,10 +42,16 @@ class AtAuthImpl implements AtAuth {
 
   PkamAuthenticator? pkamAuthenticator;
```
srieteja, line 43: Just moved the atEnrollment variable up.

packages/at_auth/lib/src/at_auth_impl.dart
```diff
@@ -417,18 +444,41 @@ class AtAuthImpl implements AtAuth {
...
+
+        secondaryAddressFinder ??= CacheableSecondaryAddressFinder(
+          atRequest.rootDomain.rootDomain,
+          atRequest.rootDomain.rootPort,
+        );
+        SecondaryAddress secondaryAddress =
+            await secondaryAddressFinder!.findSecondary(atRequest.atSign);
+
+        await (probeSocket ?? _defaultProbeSocket)(
+            secondaryAddress.host, secondaryAddress.port);
```
gkc, line 461: I think a usability improvement here could be that if the probeSocket throws, then catch it, log a warning, and continue

gkc (changes requested):
What's happening with versions here - i.e. what depends on what, and what is your plan for publishing new versions so we don't get breakages in consuming packages e.g. noports?

packages/at_auth/lib/src/at_auth_impl.dart
srieteja: Added a catch for socket exception through [51d0aa3](https://github.com/atsign-foundation/at_client_sdk/pull/1860/commits/51d0aa368aa0c16d301c6fd868185b1eb34888ac)

srieteja:
> What's happening with versions here - i.e. what depends on what, and what is your plan for publishing new versions so we don't get breakages in consuming packages e.g. noports?

at_onboarding_cli depends on at_auth. We publish `at_auth v3.0.2` first and then `at_onboarding_cli v1.15.1`. To avoid duplicate progress events or missing progress events, we should specify `at_auth >= 3.0.2` while releasing the latest version of `at_onboarding_cli`. 

This [PR](https://github.com/atsign-foundation/noports/pull/2597) in noports, uses overrides for at_auth and at_onboarding_cli from the current branch to ensure nothing is breaking.

packages/at_auth/lib/src/at_auth_impl.dart
```diff
@@ -466,8 +466,13 @@ class AtAuthImpl implements AtAuth {
             ProgressEventType.success);
 
         break; // Exit loop if no exception occurs
+      } on SocketException catch (e) {
+        _logger.warning('Attempt #[$retryCount/$maxRetries] : '
+            'Probe socket failed: $e');
+        _addProgress('Connect', '#[$retryCount/$maxRetries] : '
+            'Probe socket failed : $e', ProgressEventType.error);
```
gkc: you need to add the logic about rethrowing if retryCount >= maxRetries in the same way as it's done in the next catch block, otherwise the function could return without throwing an exception

or (better)

in the catch block, have
```
if (e is SocketException) {
  // logger.warning for socket exception as this is kinda normal
} else {
  // logger.severe for other exceptions which we don't expect
}
```

gkc (changes requested):

packages/at_auth/lib/src/at_auth_impl.dart
srieteja: Addressed through [e875ec1](https://github.com/atsign-foundation/at_client_sdk/pull/1860/commits/e875ec1d455bfaa8d7fec4c01d88de3af90e1165)

gkc (approved):

Issue #1840, opened 2026-02-23
feat(at_commons): add AtBytes equality operator


Xlin123:
### Is your feature request related to a problem? Please describe.

Across many e2e/functional and unit tests we're performing `AtBytes?.toString() == AtBytes?.toString()` 

It was necessary when we moved to at_auth 3.0 and at_chops 3.0

but now I'd like to move to an actual equality operator

### Describe the solution you'd like

Add an equality operator for `AtBytes` in `at_commons`.

Only introducing this functionality, leave the rest of the class alone.

### Describe alternatives you've considered

change strEquals method to strEquals(String? other) but that's a breaking change

### Additional context

Code example from AtKeys in at_auth
```
  @override
  bool operator ==(Object other) {
    if (identical(this, other)) return true;
    if (other is! AtKeys) return false;
    return enrollmentId == other.enrollmentId &&
        apkamPublicKey?.toString() == other.apkamPublicKey?.toString() &&
        apkamPrivateKey?.toString() == other.apkamPrivateKey?.toString() &&
        defaultEncryptionPublicKey?.toString() ==
            other.defaultEncryptionPublicKey?.toString() &&
        defaultEncryptionPrivateKey?.toString() ==
            other.defaultEncryptionPrivateKey?.toString() &&
        defaultSelfEncryptionKey?.toString() ==
            other.defaultSelfEncryptionKey?.toString() &&
        apkamSymmetricKey?.toString() == other.apkamSymmetricKey?.toString();
  }
```

== fails since we have no override, we do have a strEquals, but it expects a non-nullable making it challenging to use in our tests..

PR #1878, opened 2026-04-01
feat: introduce AtBytes.equals

Xlin123:
**- What I did**
- #1840
- TLDR; AtKeys have a lot of nullable AtBytes, cannot preform strEquals reliably due to null case.
 
**- How I did it**
- created an equals() method that compares two (AtBytes? a, AtBytes? b)

**- How to verify it**
 - see at_keys_test.dart in at_auth
 - added tests in at_bytes_test.dart
 
feat: introduce AtBytes.equals

### DIFF
--- a/packages/at_commons/lib/src/key/atbytes.dart
+++ b/packages/at_commons/lib/src/key/atbytes.dart
@@ -19,7 +19,11 @@ class AtBytes {
   bool operator ==(Object other) {
     if (other is! AtBytes) return false;
     if (identical(this, other)) return true;
-    return strEquals(other.toString());
+    if (bytes.length != other.bytes.length) return false;
+    for (int i = 0; i < bytes.length; i++) {
+      if (bytes[i] != other.bytes[i]) return false;
+    }
+    return true;
   }
 
   @override
--- a/packages/at_commons/lib/src/key/atbytes.dart
+++ b/packages/at_commons/lib/src/key/atbytes.dart
@@ -15,10 +15,13 @@ class AtBytes {
 
   bool strEquals(String str) => toString() == str;
 
-  /// Equality check on two AtBytes?, if both are null it will return true.
-  static bool equals(AtBytes? a, AtBytes? b){
-    if (identical(a, b)) return true;
-    if (a == null || b == null) return false;
-    return a.strEquals(b.toString());
+  @override
+  bool operator ==(Object other) {
+    if (other is! AtBytes) return false;
+    if (identical(this, other)) return true;
+    return strEquals(other.toString());
   }
+
+  @override
+  int get hashCode => Object.hashAll(bytes);
 }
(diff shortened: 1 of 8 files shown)


Xlin123:
will be publishing `at_commons 5.9.0`

no need to publish at_auth yet, so I placed the changes for next release.

packages/at_commons/lib/src/key/atbytes.dart
```diff
@@ -14,4 +14,11 @@ class AtBytes {
   String toString() => base64Encode(bytes);
 
   bool strEquals(String str) => toString() == str;
+
+  /// Equality check on two AtBytes?, if both are null it will return true.
+  static bool equals(AtBytes? a, AtBytes? b){
```
gkc: Please change this to be an "==" operator for AtBytes. Don't need to worry about nulls because in code if you are saying `if (a == b)` then by definition the `==` only gets called if both objects are non-null

gkc (changes requested):

packages/at_commons/lib/src/key/atbytes.dart
Xlin123: resolved in [a150e71](https://github.com/atsign-foundation/at_client_sdk/pull/1878/commits/a150e71a62f4cad017ff1ef9522f221bfa6b9b12)

packages/at_commons/lib/src/key/atbytes.dart
```diff
@@ -15,10 +15,13 @@ class AtBytes {
...
-  /// Equality check on two AtBytes?, if both are null it will return true.
-  static bool equals(AtBytes? a, AtBytes? b){
-    if (identical(a, b)) return true;
-    if (a == null || b == null) return false;
-    return a.strEquals(b.toString());
+  @override
+  bool operator ==(Object other) {
+    if (other is! AtBytes) return false;
+    if (identical(this, other)) return true;
+    return strEquals(other.toString());
```
gkc: Should not have the overhead of converting to strings. Either use `ListEquality.equals` from the [collection](https://pub.dev/packages/collection) package, or do a comparison loop here, like
```
if (bytes.length != other.bytes.length) return false;
  for (int i = 0; i < bytes.length; i++) {
    if (bytes[i] != other.bytes[i]) return false;
  }
  return true;
```

gkc (changes requested):

packages/at_commons/lib/src/key/atbytes.dart
Xlin123: resolved in [5a1aeba](https://github.com/atsign-foundation/at_client_sdk/pull/1878/commits/5a1aeba140ae38840867aa8888cc9c0679b48bd0)

gkc (approved):

PR #1887, opened 2026-04-10
build: upgrade pointycastle dependency to 4.0.0

gkc:
build: upgrade pointycastle dependency to 4.0.0


gkc:
dependency hell because the encrypt package still depends on pointycastle 3.x.x

PR #1888, opened 2026-04-11
feat: atClient API enhancements, more sample code

gkc:
**- What I did**
- feat: some api enhancements
  - add `AtCollection`s for a better, much simpler, programming experience for 90% of what an application will typically want to do with regards to data persistence and sharing
  - add factory `AtRpc.server` for easier construction of AtRpc servers (just pass a request handler)
  - add `send` and `subscribeFiltered` functions to NotificationService to provide support for the most canonical usages
  - add `Atsign get atSign` to AtClient and NotificationService
- docs: Added canonical usage examples for
  - notifications
  - rpcs
  - collections

**- How I did it**

**- How to verify it**

PR #1914, opened 2026-04-27
feat(commons): at_cli_commons 3.1.0 + at_commons 5.10.0

gkc:
## Summary

Chunk A of 7 carving up the monolithic `gkc-enhance-api` branch (PR #1888).

- `at_cli_commons` → 3.1.0: CLIBase accepts AtOnboardingPreference injection.
- `at_commons` → 5.10.0: AtKey helpers.

## Chain

- **A (this)** → B → C → D → E → F → G. Each subsequent PR targets the previous one's branch.

## Test plan

- [ ] `dart analyze packages/at_commons packages/at_cli_commons`
- [ ] `dart test -d packages/at_commons`
- [ ] `dart test -d packages/at_cli_commons`

### DIFF
--- a/packages/at_commons/lib/src/keystore/at_key.dart
+++ b/packages/at_commons/lib/src/keystore/at_key.dart
@@ -110,6 +110,10 @@ class AtKey {
     }
   }
 
+  String get fullKey => '$key${_dotNamespaceIfPresent()}';
+
+  String get fullKeyAndOwner => '$fullKey$sharedBy';
+
   @override
   String toString() {
     if (key.isNullOrEmpty) {
--- a/packages/at_commons/pubspec.yaml
+++ b/packages/at_commons/pubspec.yaml
@@ -1,6 +1,6 @@
 name: at_commons
 description: A library of Dart and Flutter utility classes that are used across other components of the atPlatform.
-version: 5.9.0
+version: 5.10.0
 repository: https://github.com/atsign-foundation/at_client_sdk/tree/trunk/packages/at_commons
 homepage: https://atsign.com
 documentation: https://docs.atsign.com/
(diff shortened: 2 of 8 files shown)


packages/at_commons/pubspec.yaml
```diff
@@ -1,6 +1,6 @@
 name: at_commons
 description: A library of Dart and Flutter utility classes that are used across other components of the atPlatform.
-version: 5.9.0
+version: 5.10.0
```
srieteja, line 3: The last published version is 5.8.0. Leaving it at v5.9.0 should be fine.

packages/at_commons/lib/src/keystore/at_key.dart
```diff
@@ -110,6 +110,10 @@ class AtKey {
     }
   }
 
+  String get fullKey => '$key${_dotNamespaceIfPresent()}';
+
+  String get fullKeyAndOwner => '$fullKey$sharedBy';
```
srieteja, line 115: We might need a null guard for `sharedBy`. Something like `$fullKey${sharedBy ?? ''}`

packages/at_commons/pubspec.yaml
gkc: Thanks @srieteja good spot. I will follow up with a PR specifically for this but in order to avoid a whole load of rebasing etc of the subsequent chunks' PRs I'd like to get this merged and then follow up once all of the chunks have been merged.

packages/at_commons/lib/src/keystore/at_key.dart
gkc: Not needed right now but in any event I think I would prefer a StateError if sharedBy is not set. Another good catch, thank you

srieteja (approved):

PR #1920, opened 2026-04-28
chore: New Flutter stable version

atsignbot:
Automated changes by [create-pull-request](https://github.com/peter-evans/create-pull-request) GitHub action


gkc:
There's a real issue here, am investigating.

gkc:
OK the issue here is that the latest trunk at_server has a bug (very minor, irrelevant to everything except our test harnesses). I will push a change for that in at_server and then we should be good to rerun here in at_client_sdk

gkc:
Closing and reopening to trigger a full re-run

gkc:
aha! it is working as expected; the problem is the APKAM enrollment doesn't have sufficient permission to use `config:set`. I'll modify the apkam setup in another PR until it's working then merge trunk into here

gkc:
> aha! it is working as expected; the problem is the APKAM enrollment doesn't have sufficient permission to use `config:set`. I'll modify the apkam setup in another PR until it's working then merge trunk into here

https://github.com/atsign-foundation/at_client_sdk/pull/1921

gkc:
Success.

Xlin123 (approved):

PR #1930, opened 2026-05-04
feat: pluggable encryption/decryption

Xlin123:
**- What I did**

Add a **pluggable, per-value encryption seam** to `at_client`. Every encrypted value carries an `appMetadata.providerId` naming the scheme that wrote it, and the SDK routes each read/write to a matching `CryptoProvider`. The built-in `legacy` provider preserves today's behaviour and is the default, so existing data stays readable forever and apps can drop in new schemes per-value with no flag-day. This is the foundation the post-quantum / group-encryption work builds on (design docs included under `docs/`).

The public surface is small — `CryptoProvider`, `CryptoConfig`, `CryptoContext`, plus `appMetadata` on the wire:

```dart
abstract class CryptoProvider {
  String get id;
  Future<String> encrypt(CryptoContext context, AtKey atKey, String plaintext);
  Future<String> decrypt(CryptoContext context, AtKey atKey, String ciphertext);
}
```

- **Stateless providers.** Everything a provider needs arrives per call via `CryptoContext` (the client) and the `AtKey`, so a single instance is safe across atSigns. No `initialize`, no factory.
- **Declared in the preference.** `AtClientPreference.crypto = CryptoConfig(defaultProviderId, providers: [...])`. There is no separate registry: `CryptoRuntime` resolves each put/get/notify/sync against the **live** `preference.crypto` by `appMetadata.providerId`, falling back to the built-in `legacy` provider. A re-used cached client adopts the new preference's crypto config.
- **SDK owns routing metadata.** After a successful `encrypt`, the runtime stamps `appMetadata.providerId` + `isEncrypted`; a provider only contributes `appMetadata.additional` (its own per-record data — IV, key id, …), which is opaque to the SDK and visible to the atServer as plaintext metadata.
- **Per-operation override:** `PutRequestOptions.cryptoProviderId` / `NotificationParams.cryptoProviderId`.
- **Wire field:** `Metadata.appMetadata = AppMetadata{providerId, additional}` (needs `at_commons ^5.11.0`, already published).

**- How I did it**

- New `lib/src/crypto/`: `crypto.dart` (the public types), `crypto_runtime.dart` (provider resolution + the `dynamic → String` boundary + per-operation exception chaining), and `legacy/` (the pre-existing scheme wrapped as `LegacyCryptoProvider`).
- `CryptoRuntime` is invoked from the put/get/notify/sync transformers: it resolves the provider, calls it, and stamps the `AtKey`'s metadata. An unknown `providerId` throws `CryptoProviderNotRegistered`; a non-String plaintext is rejected with `AtEncryptionException` at the boundary; a null ciphertext is handled gracefully rather than throwing a raw `TypeError`.
- `AtClientImpl` fails fast at construction if the configured default provider can't be resolved, and adopts the new crypto config when a cached client is re-used (covered in `AtClientManager`'s idempotency short-circuit too).
- Includes the crypto roadmap, implementation plan, and ADR-0001 under `docs/` (merged from trunk and aligned to this API), plus the slim public surface, an updated `example/bin/custom_crypto_provider.dart`, and a rewritten README "Crypto providers" section.

**- How to verify it**

```bash
cd packages/at_client && dart analyze lib test && dart test --concurrency=1   # 573 green
# functional (live atServer):
tests/at_functional_test/runLocal.sh                                          # 80/80
# e2e (cross-atSign, incl. the custom-provider machinery tests):
tests/at_end2end_test/runLocal.sh 26000                                       # 32/32
```

CI on this PR runs all of the above. `tests/at_end2end_test/test/encryption_test.dart` covers the custom-provider put/get round trip, notification delivery, the SDK stamping `providerId`/`isEncrypted` on the provider's behalf, adopt-config-on-reuse (replacing the old config), and routing by `providerId` among multiple providers.

feat: pluggable per-value encryption — apps configure custom `CryptoProvider`s via `AtClientPreference.crypto`; the SDK routes each value by `appMetadata.providerId`, with the built-in `legacy` provider as the default and fallback. Old data stays readable; new schemes drop in per-value.


packages/at_commons/lib/src/verb/update_verb_builder.dart
gkc, line 1: @Xlin123  You need to make similar changes to the NotifyVerbBuilder

Xlin123: good catch, will do

gkc: @Xlin123 and of course the unit tests

Xlin123: notify currently doesn't have a `getBuilder(String command)` method, and buildCommand is handled via `metadata.toAtProtocolFragment()`

should I add one? would we be creating/handling notify from raw command?

packages/at_client/lib/src/crypto/legacy/legacy_decryption.dart
Xlin123, line 1: legacy cleaned up and renamed, no functionality changed.

packages/at_client/lib/src/crypto/legacy/legacy_encryption.dart
Xlin123, line 1: legacy cleaned up and renamed, no functionality changed.

packages/at_client/lib/src/crypto/legacy/legacy_crypto_provider.dart
Xlin123, line 1: example of a `CryptoProvider`

packages/at_commons/lib/src/verb/update_verb_builder.dart
Xlin123: [472ff2d](https://github.com/atsign-foundation/at_client_sdk/pull/1930/commits/472ff2d1e2bd33bc68161df811a72b7a24d6bc08)

packages/at_client/lib/src/crypto/crypto.dart
Xlin123, line 1: new models

packages/at_client/lib/src/crypto/crypto_runtime.dart
Xlin123, line 1: `CryptoRuntime` will handle all routing logic for providers, and exposes the basic enc/dec functions of our sdk.

ie: decryptForNotification, etc.

packages/at_client/lib/src/crypto/crypto_storage.dart
Xlin123, line 1: implementation of the CryptoStorage contract but specifically for secondary servers (local and remote)

leaves options open for if we (or the user) ever spin up different types of storages.

packages/at_client/lib/src/crypto/crypto.dart
```diff
@@ -0,0 +1,232 @@
...
+
+class ThrowCryptoError extends CryptoFailureResolution {
+  const ThrowCryptoError();
+}
+
+class RetryCryptoOperation extends CryptoFailureResolution {
+  const RetryCryptoOperation();
+}
+
+class CryptoProviderNotFoundContext {
```
Xlin123: Error context, provides at_client / key details and contains method to registerProvider in order to support registered after not found.

packages/at_client/lib/src/crypto/crypto.dart
```diff
@@ -0,0 +1,232 @@
...
+    required this.owner,
+    required this.recipient,
+    required this.namespace,
+    required this.name,
+  });
+}
+
+/// Main implementation and contact point for pluggable encryption
+/// this is the contract which all CryptoProvider implementations must follow.
+abstract class CryptoProvider {
```
Xlin123, line 66: main contact for pluggable encryption

packages/at_client/lib/src/crypto/crypto.dart
```diff
@@ -0,0 +1,232 @@
...
+
+  Future<CryptoEncryptResult> encrypt(CryptoEncryptRequest request);
+
+  Future<CryptoDecryptResult> decrypt(CryptoDecryptRequest request);
+}
+
+// Request & Response models used during runtime routing
+// main consumers: [CryptoRuntime] and [CryptoProvider]
+
+class CryptoEncryptRequest {
```
Xlin123: keeps the request and responses open ended for specific implementations if required

Xlin123 (commented):
LGTM @gkc thank you for the changes 👍 

we'll need another reviewer to approve

packages/at_client/example/bin/custom_crypto_provider.dart
```diff
@@ -0,0 +1,173 @@
...
+      }
+      final key =
+          AtKey()
+            ..key = 'secret'
+            ..namespace = applicationNamespace
+            ..sharedBy = me
+            ..sharedWith = other.toAtsign();
+
+      const plaintext = 'hello from the xor provider!';
+      stdout.writeln('Writing key: @$other:${key.key}.${key.namespace}@$me');
```
JeremyTubongbanua: I think here use key.toString() ?

packages/at_client/example/bin/custom_crypto_provider.dart
```diff
@@ -0,0 +1,173 @@
...
+      }
+      // sharedBy = the writer, sharedWith = me (the reader)
+      final key =
+          AtKey()
+            ..key = 'secret'
+            ..namespace = applicationNamespace
+            ..sharedBy = other.toAtsign()
+            ..sharedWith = me;
+
+      stdout.writeln('Reading key: @$me:${key.key}.${key.namespace}@$other');
```
JeremyTubongbanua: Same thing here?

packages/at_client/lib/src/crypto/crypto.dart
```diff
@@ -0,0 +1,94 @@
+import 'package:at_client/src/client/at_client_spec.dart';
+import 'package:at_commons/at_commons.dart';
+
+/// The id of the built-in legacy (pre-pluggable) encryption scheme — the
+/// default provider and the fallback for records with no `appMetadata`.
+const String legacyCryptoProviderId = 'legacy';
```
JeremyTubongbanua, line 6: Is there a reason we're not making a CryptoProviderId enum?

packages/at_client/lib/src/crypto/crypto_runtime.dart
```diff
@@ -0,0 +1,140 @@
+import 'package:at_client/src/client/at_client_spec.dart';
+import 'package:at_client/src/crypto/crypto.dart';
+import 'package:at_client/src/crypto/legacy/legacy_crypto_provider.dart';
+import 'package:at_commons/at_commons.dart';
+
+/// Routes encryption/decryption to the [CryptoProvider] named by an [AtKey]'s
+/// `appMetadata.providerId`.
+class CryptoRuntime {
+  static const String legacyProviderId = legacyCryptoProviderId;
```
JeremyTubongbanua, line 9: Duplicate variable with what is already in `legacy_crypto_provider.dart` ?

packages/at_client/lib/src/crypto/crypto.dart
Xlin123: making it an enum forces it to become a hard-coded first class object (how are we able to account for new schemes?)

packages/at_client/lib/src/crypto/crypto_runtime.dart
JeremyTubongbanua: Oh if this is for future-proofing what "legacy" is then that's fine

tests/at_end2end_test/test/encryption_test.dart
```diff
@@ -25,14 +27,64 @@ void main() {
...
+    );
+    return atClientManager.atClient;
+  }
+
+  // Like getAtClient, but configures an explicit list of providers (for the
+  // multi-provider / config-adoption machinery tests below).
+  Future<AtClient> getAtClientWith(
+    String atSign,
+    List<CryptoProvider> providers, {
+    String defaultProviderId = 'legacy',
```
JeremyTubongbanua: I think we should use a variable or enum here

JeremyTubongbanua (dismissed):
Small nits but LGTM yay

Xlin123:
I'll go these change requests, thanks @JeremyTubongbanua for the review.

tests/at_end2end_test/test/encryption_test.dart
Xlin123: [6e74c21](https://github.com/atsign-foundation/at_client_sdk/pull/1930/commits/6e74c218934e64820a930ce8ad5c5a89881f9c45)

packages/at_client/example/bin/custom_crypto_provider.dart
Xlin123: [6e74c21](https://github.com/atsign-foundation/at_client_sdk/pull/1930/commits/6e74c218934e64820a930ce8ad5c5a89881f9c45)

packages/at_client/example/bin/custom_crypto_provider.dart
Xlin123: [6e74c21](https://github.com/atsign-foundation/at_client_sdk/pull/1930/commits/6e74c218934e64820a930ce8ad5c5a89881f9c45)

gkc:
Examples are still not using $key instead of $other.....

JeremyTubongbanua (approved):

gkc (approved):

JeremyTubongbanua (approved):

PR #1937, opened 2026-05-12
feat(at_auth): added `requestActivationOtp`

Xlin123:
**- What I did**
 - deprecated bad method sendActivationOtp with `Future<bool>` signature in favour of `Future<void>` with exceptions
 -  will be publishing 3.1.1 patch 
  
**- How I did it**
 - Added `RegistrarException` to at_auth_exceptions.dart
  
**- How to verify it**
  - added unit tests for new method
 
feat(at_auth): added `requestActivationOtp`


gkc (dismissed):

packages/at_auth/lib/src/registrar/registrar_service.dart
```diff
@@ -80,15 +83,32 @@ class RegistrarService implements Registrar {
...
+    var payload = jsonDecode(res.body);
+    if (payload["message"] != "Sent Successfully") {
+      throw RegistrarException(res.body);
+    }
+  }
+
   @override
   //TODO: this really should be Future<String>
   Future<String?> verifyActivation({
     required String atSign,
```
CurtlyCritchlow, line 105: Rename parameter to `atsign`.

CurtlyCritchlow (changes requested):
Have a minor request. it's part of cleaning up our sdk overtime. See comments.

packages/at_auth/lib/src/registrar/registrar_service.dart
Xlin123: most of them are named parameters so would be a BC

PR #1976, opened 2026-06-11
feat(at_client): same-atSign pairwise secret sharing

gkc:
**- What I did**

Added "alice to alice" secret sharing: multiple atClients authenticated as the same atSign (on the same or different APKAM enrollments) can share secrets pairwise, end-to-end encrypted so that only the target client can read them — not every client that holds the selfEncryptionKey.

- Per-client registration: each client publishes an APKAM-signed key bundle as a hidden public key in its enrollment's reserved namespace, with optional namespace-scoped copies for efficient discovery of clients participating in a given application namespace.
- Pairwise secret envelopes: a random content key is wrapped to the recipient's bundle key, the payload encrypted, and the envelope APKAM-signed. Envelopes are self keys whose names end with the secret's application namespace, so the atServer's existing enrollment authorization scopes who can fetch/sync them — no server changes. Receivers verify before decrypting, then delete.
- `SecretStore` plus an approver-side helper (`shareAllSecretsWithEnrollment`) that shares held secrets with a newly approved enrollment's clients, filtered by its approved namespaces. The enroll/approve lifecycle itself is untouched.
- **Post-quantum from day one**: per-client keys are X-Wing hybrid KEM keypairs (X25519 + ML-KEM-768, `draft-connolly-cfrg-xwing-kem-10`, verified against the draft's test vectors byte-exact), and envelope payloads are sealed with AES-256-GCM under the KEM-encapsulated shared secret — confidentiality is harvest-now-decrypt-later resistant, and nothing secret travels except the KEM ciphertext. Both primitives land in at_chops in this PR. Bundle and envelope formats carry explicit algorithm ids for future agility.
- The app-facing API surface is marked `@experimental`: this is the substrate for the group-based encryption direction in `docs/crypto-roadmap.md` (included), where the durable app surface will be `SecureGroup`. Types use the MLS-correspondent KeyPackage naming.
- fix: `get` now respects an explicit `isEncrypted=false` and returns the raw value without attempting decryption; legacy data with the flag absent still takes the try-decrypt fallback.
- fix: dotted application namespaces are preserved intact on the receive side.

**- How I did it**

New `lib/src/secret_sharing/` library in `at_client` plus an `EnvelopeSigning` mixin (signs with the APKAM keypair; verified against the signer enrollment's published `_apsk` key), composed by the ready-made `AtClientSecretSharing` class and exported via `package:at_client/at_client_mixins.dart`. No changes to atServer, at_auth, or at_chops.

**- How to verify it**

- Unit tests: `cd packages/at_client && dart test --concurrency=1`
- Functional (against the virtualenv): `tests/at_functional_test/test/secret_sharing_test.dart` — two clients of one atSign register, discover, share, receive, and clean up end to end
- Manually: run `packages/at_client/example/bin/secret_sharing.dart` as receiver and then sender with the same atSign (see `example/README.md`)

Same-atSign pairwise secret sharing (experimental): post-quantum per-client key packages (X-Wing) with namespace-scoped discovery, per-client AES-256-GCM envelopes, secret store and enrollment-approval sharing helper; at_chops gains X-Wing and AES-256-GCM; get now respects explicit isEncrypted=false.


gkc:
Four additional commits:

- `docs: add encryption/decryption roadmap` — `docs/crypto-roadmap.md` places this PR as step 1 of the broader encryption/decryption direction (pluggable CryptoProvider, rotating-key provider, post-quantum, pq-mls) and records the coordination points with the `xl-pluggable` and `jt-pq` branches.
- `refactor: rename secret-sharing wire markers pq -> ss` — `__sskb-`/`sskb-`/`__sskbns`; the markers name the mechanism, not its algorithms.
- `feat: shared per-AtClient secret sharing instance` — `AtClientSecretSharing.forClient(atClient)` (one client identity per AtClient, shared by the app and future SDK-internal consumers); `registerClient` namespaces are now additive across calls.
- `feat: waitForSecret and reserved system secret names` — race-free await of a secret's arrival; `__`-prefixed secret names reserved for system use (e.g. a future rotating-key provider's `__rk.*` key material).

PR #1981, opened 2026-06-15
feat(at_commons): add Metadata.appMetadata wire field (5.11.0)

gkc:
**- What I did**

Added `Metadata.appMetadata` — an `AppMetadata{providerId, additional}` value carried on the wire as `:appMetadata:` (base64-encoded JSON) on the `update`, `update:meta`, and `notify` verbs, and parsed back by the verb builders. Also added the supporting `CryptoProviderNotRegistered` exception and the `fetchCryptoProvider` `Intent`, and bumped at_commons to 5.11.0.

This is the foundational, dependency-free first step of a series that lands pluggable, post-quantum-ready encryption in the SDK. It adds **only** the at_commons wire field and shared vocabulary — the consuming logic ships in later, dependent PRs (at_chops, at_persistence_secondary_server, at_client). `CryptoProviderNotRegistered`/`fetchCryptoProvider` are not yet referenced in-tree; they live here because at_commons owns the `AtClientException`/`Intent` vocabulary and at_client 3.13.0 will consume them from the published at_commons.

**- How I did it**

- `AppMetadata.providerId` — SDK-owned routing field naming the crypto provider able to decrypt the value; validated as a non-empty string (`FormatException` otherwise).
- `AppMetadata.additional` — provider-owned opaque plaintext metadata the SDK preserves but does not interpret (plaintext on the wire, so no secrets).
- Serialized via `Metadata.encodeAppMetadata` (base64-JSON) only when present; decoded in `Metadata.fromJson` and the verb builders. `==`/`hashCode` on `Metadata` and `UpdateVerbBuilder` extended to include it.
- No behavioural change for existing callers: `appMetadata` is `null` unless set.

**- How to verify it**

New coverage in `at_key_test.dart`, `update_verb_builder_test.dart`, and `notify_verb_builder_test.dart` (encode/decode round trip, verb serialization/parsing, `providerId` validation). Full suite: **486 tests pass**; `dart analyze --fatal-warnings` clean; `dart format` clean.

Add `Metadata.appMetadata` wire field (and `CryptoProviderNotRegistered`/`fetchCryptoProvider`) for pluggable-crypto provider routing.

### DIFF
--- a/packages/at_commons/lib/src/exception/at_client_exceptions.dart
+++ b/packages/at_commons/lib/src/exception/at_client_exceptions.dart
@@ -82,3 +82,10 @@ class InvalidPinException extends AtClientException {
 
   InvalidPinException.message(String message) : super.message(message);
 }
+
+class CryptoProviderNotRegistered extends AtClientException {
+  CryptoProviderNotRegistered(String message,
+      {Intent? intent, ExceptionScenario? exceptionScenario})
+      : super.message(message,
+            intent: intent, exceptionScenario: exceptionScenario);
+}
--- a/packages/at_commons/CHANGELOG.md
+++ b/packages/at_commons/CHANGELOG.md
@@ -1,3 +1,12 @@
+## 5.11.0
+
+- feat: add `Metadata.appMetadata` (`AppMetadata{providerId, additional}`),
+  emitted on the wire as `:appMetadata:` (base64-encoded JSON) on the
+  `update`, `update:meta` and `notify` verbs and parsed back by the verb
+  builders. `providerId` routes pluggable-crypto decryption; `additional`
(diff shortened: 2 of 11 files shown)


packages/at_commons/lib/src/exception/at_client_exceptions.dart
```diff
@@ -82,3 +82,10 @@ class InvalidPinException extends AtClientException {
 
   InvalidPinException.message(String message) : super.message(message);
 }
+
+class CryptoProviderNotRegistered extends AtClientException {
```
gkc, line 86: this isn't used in this PR; it's an AtClient thing but am putting in this PR so we just have one at_commons package bump

Xlin123 (approved):
thanks @gkc

PR #1982, opened 2026-06-15
feat(at_chops): X-Wing KEM, AES-256-GCM, HKDF, key consolidation (3.2.0)

gkc:
**- What I did**

Adds the post-quantum and KDF crypto primitives to at_chops and consolidates the key classes, releasing 3.2.0:

- **`XWingPureDartAlgo`** — the X-Wing hybrid PQ/traditional KEM (draft-connolly-cfrg-xwing-kem-10; X25519 + ML-KEM-768), with `AtXWingKeyPair` and `AtChopsUtil.generateXWingKeyPair()`.
- **`AesGcm256EncryptionAlgo`** — AES-256-GCM authenticated encryption (pure-Dart). This is the AEAD (Authenticated Encryption with Associated Data) that encrypts the secret-sharing envelope payloads and `SecureGroup.seal`/`open` data in the later at_client work; an authenticated cipher is required here (the existing AES-CTR `AESEncryptionAlgo` is not) to detect tampering of peer-supplied ciphertext.
- **`HmacSha256` + `HkdfSha256`** (RFC 2104 / RFC 5869) and `SHA256HashingAlgo`, with `md5`/`sha256` wired into the hashing-algo factory.
- **Key-class consolidation** — `AtPublicKey`/`AtPrivateKey`/`SymmetricKey`/`AsymmetricKeyPair` into a single sealed hierarchy in `src/key/keys.dart`.

This is dependency-independent (it keeps `at_commons: ^5.6.0`, no reliance on unreleased packages) and can merge on its own. It's the second step of a series landing pluggable, post-quantum-ready encryption in the SDK; later PRs (at_persistence_secondary_server, at_client) consume these primitives.

**- How I did it**

- X-Wing follows the draft's combiner over X25519 + ML-KEM-768; `MlKem768PureDartAlgo.encapsulate` gained an optional deterministic-randomness parameter (FIPS 203 `m`) purely so the KEM can be checked against published test vectors.
- AES-256-GCM emits `ciphertext || tag` with an explicit 12-byte nonce and raises `AtDecryptionException` on tag mismatch.
- HKDF is the standard extract-then-expand over `HmacSha256`.
- The key consolidation moves the **same** public classes/constructors into `src/key/keys.dart`; they remain exported from `package:at_chops/at_chops.dart`, so it is source-compatible. Only direct `src/`-path imports of the removed files (`at_key_pair.dart`/`at_private_key.dart`/`at_public_key.dart`) would break — verified that no in-repo package or pub.dev consumer imports those `src/` paths, hence a minor (3.2.0) rather than major bump.

**- How to verify it**

New byte-exact vector coverage: X-Wing draft vectors (`x_wing_algo_test.dart`), AES-256-GCM (`aes_gcm_encryption_algo_test.dart`), HKDF/HMAC RFC vectors (`hkdf_algo_test.dart`). Full suite: **112 tests pass**; `dart analyze --fatal-warnings` clean; `dart format` clean.

Add X-Wing KEM, AES-256-GCM, HKDF/HMAC + SHA-256 hashing, and consolidate the key classes (source-compatible).


packages/at_chops/lib/src/key/keys.dart
Xlin123, line 1: thank you this was bothering me 🥇

packages/at_chops/lib/src/algorithm/x_wing_pure_dart_algo.dart
JeremyTubongbanua, line 1: Do we need an x_wing_dart_ffi_algo.dart as well?

Xlin123 (dismissed):
lgtm

packages/at_chops/lib/src/algorithm/x_wing_pure_dart_algo.dart
gkc: > Do we need an x_wing_dart_ffi_algo.dart as well?

Yes good catch

JeremyTubongbanua: From my little prompting, there doesn't seem to be an XWing implementation in OpenSSL directly, so 
the "x_wing_ffi_algo.dart" implementation would make use of "ml_kem_768_ffi_algo.dart" and "x25519_ffi_algo.dart"

JeremyTubongbanua (commented):
Once x wing ffi is implemented,

we'll need x wing ffi tests that puts this at the top of the file:

```dart
@Tags(['ffi'])
```

This test file would also have interoperability tests with pure Dart.

See "ml_kem_768_ffi_test.dart" and "x25519_ffi_test.dart" for examples

JeremyTubongbanua (changes requested):
- [ ] `x_wing_ffi_algo.dart` (OpenSSL implementation)
- [ ] `x_wing_ffi_test.dart` (contains ffi tests AND interoperability with pure Dart) (don't forget ffi tag at top of file)
- [ ] Update CHANGELOG

packages/at_chops/lib/src/key/at_key_pair.dart
JeremyTubongbanua, line 1: Is this a breaking change?

Xlin123: no, still exported by keys.dart

gkc:
Thanks @JeremyTubongbanua — addressed in 48f0475e5:

- [x] **`x_wing_ffi_algo.dart` (OpenSSL implementation)** — `XWingFfiAlgo` composes `MlKem768FfiAlgo` + `X25519FfiAlgo` (X-Wing has no native OpenSSL primitive). The SHAKE-256 expansion and SHA3-256 combiner are pure-Dart (pointycastle) and byte-identical to `XWingPureDartAlgo`, so the two backends are fully interoperable.
- [x] **`x_wing_ffi_test.dart`** (`@Tags(['ffi'])` at top, ffi tests + interop with pure-Dart) — keygen byte-matches pure-Dart for the same seed, FFI decapsulates the draft vector ciphertext to the published shared secret, round-trip, bidirectional FFI↔pure interop, and implicit-rejection.
- [x] **Update CHANGELOG** — folded into the in-progress 3.2.0 entry.

Two supporting additions were needed since the FFI ML-KEM didn't expose them yet:
- `MlKem768FfiAlgo.generateKeyPairFromSeed` — deterministic ML-KEM-768 keygen from a 64-byte FIPS 203 `d || z` seed via OpenSSL `fromdata`, so X-Wing decapsulation can re-derive its ML-KEM key from the X-Wing seed.
- `X25519FfiAlgo.publicKeyFromPrivate` — derive `pk_X` from the expanded X25519 secret.

Note: there's no derandomized encapsulation on the FFI path — OpenSSL's ML-KEM encapsulation takes no external randomness — so the draft KAT vectors are verified against the pure-Dart backend, while the FFI backend is verified by decapsulating the draft vector ciphertext to the published secret.

All 6 FFI tests pass against OpenSSL 3.6.1 locally; full at_chops suite (118) green; `dart analyze --fatal-warnings` + `dart format` clean.

packages/at_chops/lib/src/algorithm/ml_kem_768_ffi_algo.dart
gkc, line 1: dart format changes

packages/at_chops/lib/src/algorithm/ml_kem_768_ffi_algo.dart
```diff
@@ -122,6 +127,62 @@ final class MlKem768FfiAlgo implements AtKemAlgorithm {
     }
   }
```
gkc, line 129: new method

packages/at_chops/lib/src/algorithm/x25519_ffi_algo.dart
```diff
@@ -34,22 +34,26 @@ final class X25519FfiAlgo implements AtKeyAgreementAlgorithm {
         'EVP_PKEY_CTX_new');
     _ctxFree = _lib.lookupFunction<EvpPkeyCtxFreeNative, EvpPkeyCtxFreeDart>(
         'EVP_PKEY_CTX_free');
```
gkc, line 36: dart format

packages/at_chops/lib/src/algorithm/x25519_ffi_algo.dart
```diff
@@ -124,6 +128,26 @@ final class X25519FfiAlgo implements AtKeyAgreementAlgorithm {
     }
   }
```
gkc, line 130: new method

packages/at_chops/lib/src/algorithm/x_wing_ffi_algo.dart
gkc, line 1: new xwing ffi algo

JeremyTubongbanua (dismissed):
Tests are currently being skipped, but #1983 will fix it

gkc:
@JeremyTubongbanua let's merge this one **after** #1983 in that case

gkc:
marking as draft so we don't accidentally merge until after we've merged #1983

JeremyTubongbanua (approved):
I fixed merge conflicts and made it so that FFI tests fail when libcrypto is unavailable.

Xlin123 (approved):

PR #1986, opened 2026-06-17
fix(at_client_flutter): Force uppercase on OTP input fields

LilianGRD:
- Switch the APKAM dialog keyboard from numeric to visible-password so letters can be typed.
- Force uppercase on both the APKAM and CRAM input fields, so a lowercase is replaced with uppercase


Xlin123 (approved):
please use semantic commits for the future please

LilianGRD:
sorry I made that quickly

PR #1987, opened 2026-06-17
feat(at_auth): introduce AtKeysSet models

Xlin123:
**- What I did**
- This will be the first small change before I make large changes regarding the AtKeysIo formats.
- This PR will also act as a discussion for any comments/concerns related to the new AtKeys model. The actual I/O relating to formats of the JSON and how we read/write legacy and modern implementations will happen in their respective PRs.
- I only added the new AtKeysSet model, none of the classes are picking up the new model yet.
- the at_auth barrel is still exporting the exact same classes, AtKeysSet will be folded in with the proper AtKeysIo changes and the tests regarding those changes.
- first step in : #2005 

**- How I did it**
 Open-ended atKeys format will be shaped by the following:
1. , when we onboard / read keys / apkam
    i. during this time, at_auth will ensure the AtKeysSet contains the necessary items for auth & general conversation.
2. After authentication, during AtClient creation, the CryptoProviders that are being registered will ensure that their respective  keys are ALSO in place in the AtKeySet, and if not will preform the necessary operations for that. 
3. during  cryptography operations the CryptoProviders will pull their validated keys that are required.

AtKeysSet is internally a map, mapping to id or pairId (for symmetric and key pairs respectively). However, externally is remains acting as a list.

**- How to verify it**
- no functionality changes, still relying on legacy / `at_keys_legacy.dart` across all classes.
- so all old tests pass.

**- Description for the changelog**
feat(at_auth): introduce AtKeysSet models

PR #1989, opened 2026-06-17
feat(at_chops): stateless core, deprecating old API

Xlin123:
**- What I did**
 - planning for a minor version of at_chops 3.3.0, roughly half of the work for the outlined #2004
 - deprecate all state in at_chops and move towards direct apis on the algorithms
 - reorganized and renamed the algorithms into their respective types
 - collapsed AtChopsUtil (which generated keys) into the respective key classes as a static generate() class
  
**- How I did it**
- rearrange, deprecate, renamed, reorganize.

**- How to verify it**
 - all tests still pass
 
feat(at_chops): stateless core, deprecating old API


Xlin123:
needs a revert on the tests, my bad.

gkc (changes requested):
**1. at_chops changes — LGTM.** Stateless core + deprecations are clean and non-breaking: old types kept and `@Deprecated` with clear messages, barrel symbols preserved, new `generate()` key classes are tidy, README updated, version correctly folds into the unpublished 3.3.0. CI fully green.

**2. Pull the AtKeys work out into its own PR.** All of it belongs in **at_auth**, where `AtKeys`/`AtKeysIo` already live (not at_commons). I have a number of thoughts on the shape that I'll leave on that PR. It's also currently unused on this branch, so removing it doesn't affect anything here.

**3. Also worth noting:** dropping the at_commons changes resolves a side issue — the new `atkeys.dart` adds public API to at_commons with no version bump or CHANGELOG entry (5.11.0 is already published), which would need fixing if it stayed.

Xlin123:
good catch @gkc, will note that in [this PR](https://github.com/atsign-foundation/at_client_sdk/pull/1987)

and remove it from this PR (I must've based off of the wrong branch)

gkc (approved):
thanks Xavier, LGTM and thanks for the test changes also which ensure we're testing both the new and the deprecated.

PR #1993, opened 2026-06-20
feat: PQ HPKE seal in at_chops

srieteja:
## What

  - Adds post-quantum authenticated encryption to at_chops via pqSeal / pqOpen
  - Construction follows the HPKE shape (RFC 9180 inspired) but uses a custom at_protocol-internal wire envelope — not RFC 9180 wire-compatible
  - Bumps at_chops to 3.3.0
  - Lays the groundwork for PQ-secured shared-key exchange in the atProtocol

## How

  - KEM → X-Wing (AtKemAlgorithm, caller-supplied); KDF → HkdfSha256 (new hkdf_algo.dart); AEAD → AES-256-GCM (AesGcm256EncryptionAlgo)
  - Envelope format: ver(1) || ctLen(2, BE) || kemCt || gcmCipherText || tag(16) — version byte is checked on open for forwards-compat; mismatches throw PqOpenException(PqOpenFailure.versionMismatch, ...)
  - _deriveKeyAndNonce runs HKDF over the 32-byte X-Wing shared secret + optional info binding to produce the AES key + deterministic nonce — no separate randomness needed post-KEM
  - pqOpen validates envelope length before parsing, extracts kemCt via the 2-byte BE length field, decapsulates, re-derives the key, and maps AtDecryptionException → PqOpenException(authFailure) for typed
  error handling
  - HKDF step provides context binding and key/nonce derivation (not extraction — X-Wing shared secret is already uniformly random)
  - Future-proofing: the version byte allows new envelope schemas without breaking readers of old ciphertext

## How to verify

- Tests pass here

## Skills impact

### DIFF
--- a/packages/at_chops/test/pq_hpke_test.dart
+++ b/packages/at_chops/test/pq_hpke_test.dart
@@ -24,7 +24,7 @@ final class _FixedKem implements AtKemAlgorithm {
       _ss;
 }
 
-Uint8List _b(String s) => Uint8List.fromList(utf8.encode(s));
+Uint8List _utf8(String s) => Uint8List.fromList(utf8.encode(s));
 
 Matcher _opensWith(PqOpenFailure reason) => throwsA(isA<PqOpenException>()
     .having((e) => e.reason, 'reason', reason));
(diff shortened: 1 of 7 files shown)


gkc (changes requested):
Direction here is great — HPKE-base over X-Wing is exactly the standardized PQ public-key-encryption primitive we want, and it replaces the hand-rolled "encapsulate + AES-GCM" I've currently coded directly in the secret-sharing substrate in gkc-pqmls-spike. The stateless, keys-as-parameters shape is also exactly right — a model citizen for the AtChops-stateless direction. Requesting changes only on quality/scoping, not direction:

**1. Tests.** This adds ~195 lines of new crypto with no test file. Please add at least: a `seal`→`open` round-trip; tamper (flip a ciphertext/tag byte) → `authFailure`; `info`/`aad` mismatch → `authFailure`; version-mismatch and malformed-envelope cases. Known-answer vectors would be even better (RFC 9180 has HPKE vectors — adapt, since this is HPKE-*style*, not wire-compatible). at_chops carries ~128 tests and the bar here is round-trip + vectors byte-exact. (See for example the x_wing_test_vectors.dart file in at_chops/test which takes example vectors from the spec draft and is used by a couple of sets of tests)

**2. Reuse the existing at_chops primitives.** This PR imports `AesGcm` / `Hkdf` / `Hmac` directly from `package:cryptography`, but at_chops already ships `AesGcm256EncryptionAlgo`, `HkdfSha256`, and `HmacSha256` (vector-verified). Two HKDF and two AES-GCM implementations in one package is a maintenance smell — please reuse the existing ones, or note why they don't fit here. (`package:cryptography` is fine to import in at_chops — it's the wrapper — so this is consistency within at_chops, not a crypto-import-gate issue.)

**3. Keep at_chops protocol-agnostic.** `pqDeriveConfirmationTag` / `pqDeriveSessionKey` bake atsign Protocol semantics into a primitive library (the `atPQv1-polconfirm` / `atPQv1-session` labels, "from/pol handshake", `sessionID || fromAtSign`). Either make them generic (an HKDF-derive taking a caller-supplied label/context), or move the named helpers up to the at_client or at_server layer. The `pqSeal`/`pqOpen` primitive itself is correctly generic.

**4. Dartdoc: "HPKE-style", not HPKE-interop.** It mirrors the construction but uses a custom envelope (`ver || ctLen || kemCt || ct || tag`), not the RFC 9180 wire format. Fine for internal envelopes — just make the dartdoc clear it is not wire-compatible HPKE, so nobody expects interop.

Follow-up (for @gkc, not blocking this PR): once this lands, the secret-sharing `__ssenv` envelopes and the upcoming `nskey` provider in the gkc-pqmls-spike branch should seal/open through this primitive rather than open-coding encapsulate+GCM again — one audited PQ-PKE path.

gkc:
@srieteja PS my review is in the context of the overall crypto roadmap and my impl to date

gkc:
@srieteja PPS I've incorporated the changes proposed in this PR into the overall roadmap / plan at https://github.com/atsign-foundation/at_client_sdk/tree/gkc-pqmls-spike/docs

srieteja:
Thank you for the really detailed review @gkc.
The latest commits introduce tests and ensure that the methods are atPlatform-agnostic. I've ensured that the impl uses existing primitives.

JeremyTubongbanua (dismissed):
LGTM

packages/at_chops/test/pq_hpke_test.dart
```diff
@@ -0,0 +1,232 @@
...
+      expect(() => pqOpen(xwing, XWingVector1.seed, bad),
+          _opensWith(PqOpenFailure.malformedEnvelope));
+    });
+  });
+
+  group('known-answer vectors (byte-exact)', () {
+    // Golden envelopes are pinned, computed once by composing the
+    // already-vector-verified HkdfSha256 + AesGcm256EncryptionAlgo. They lock
+    // the seal construction (HKDF key schedule + AES-256-GCM body) against
+    // accidental drift. Regenerate via tool_kat_bootstrap.dart if the
```
gkc: What / where is `tool_kat_bootstrap.dart`?

gkc (changes requested):
Great stuff! LGTM except for one nit about a reference that confused me, for which I left a comment

packages/at_chops/test/pq_hpke_test.dart
srieteja: That is an outdated comment, has been removed now in [efcef53](https://github.com/atsign-foundation/at_client_sdk/pull/1993/commits/efcef53416d61ce75ca94db9ebe1483d92efb437)

gkc (approved):

Xlin123:
completes #2004

PR #2022, opened 2026-06-23
fix: allow at_chops to use pqcrypto 0.3.0+

srieteja:
## What
- update at_chops deps `pqcrypto: 0.3.0` -> `pqcrypto: ^0.3.0`

## Skills impact

### DIFF
--- a/packages/at_chops/pubspec.yaml
+++ b/packages/at_chops/pubspec.yaml
@@ -24,7 +24,7 @@ dependencies:
   meta: ^1.9.0
   better_cryptography: ^1.0.0+1
   ffi: ^2.1.0
-  pqcrypto: 0.3.0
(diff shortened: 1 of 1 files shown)


gkc (approved):

JeremyTubongbanua:
Do you think we need to upgrade pub dev version @srieteja

srieteja:
> Do you think we need to upgrade pub dev version @srieteja 

Do you mean the package version @JeremyTubongbanua?

srieteja:
> > Do you think we need to upgrade pub dev version @srieteja
> 
> Do you mean the package version @JeremyTubongbanua?

That was already updated. This was flagged in the dry run, so I fixed it before publishing.

Issue #1889, opened 2026-04-13
Implement post quantum (PQ) cryptography along with crypto agility


cpswan:
### Is your feature request related to a problem? Please describe.

The RSA2048 asymmetric encryption that we presently use is vulnerable to [Shor's algorithm](https://en.wikipedia.org/wiki/Shor%27s_algorithm), implying that it will no longer be fit for purpose as we approach 'Q day' when quantum computers become available that can substantially reduce the time needed to break RSA2048.

### Describe the solution you'd like

We should initially implement a hybrid of a NIST approved Elliptic-curve cryptography ([ECC](https://en.wikipedia.org/wiki/Elliptic-curve_cryptography)) algorithm such as [X25519](https://en.wikipedia.org/wiki/Curve25519) along with a [NIST approved PQ algorithm](https://en.wikipedia.org/wiki/NIST_Post-Quantum_Cryptography_Standardization#First_release) such as [ML-KEM](https://en.wikipedia.org/wiki/ML-KEM).

But we should engineer in [cryptographic agility](https://en.wikipedia.org/wiki/Cryptographic_agility) such that algorithms can be easily replaced if found to be vulnerable (e.g. [SIKE was found to be insecure](https://csrc.nist.gov/csrc/media/Projects/post-quantum-cryptography/documents/round-4/submissions/sike-team-note-insecure.pdf) during the NIST evaluation process),

### Describe alternatives you've considered

We might give some consideration to how we continue to use [IoT SAFE](https://www.gsma.com/solutions-and-impact/technologies/internet-of-things/iot-safe/) for storage of keys in SIM cards.

### Additional context

[Atsigns and Q day discussion doc](https://docs.google.com/document/d/17_TlNJOTPBU-VAZR1OoX1d9R3B4OfAdx_aReK1rzgXI/edit?tab=t.0)

gkc:
@srieteja @JeremyTubongbanua @Xlin123 some notes after some thought and some discussion with Srie last week 

**1. Where the q-day vulnerabilities are**

Symmetric crypto already survives quantum:
AES-256 gives ~128-bit PQ security, SHA-256/512 keep ~128/256-bit preimage
resistance, so every encrypted atKey value and everything at rest under the
self-encryption key needs no change. Only two asymmetric operations break under
Shor:

- **RSA key-wrap of the per-pair AES shared key** (`sharedKeyEnc`, the
  `shared_key.*@*` records). This is the urgent one because of
[...]

PR #2030, opened 2026-06-25
feat(at_chops): add at_chops_ffi.dart barrel and PqcFfi auto-resolver

srieteja:
### What
- Adds `at_chops_ffi.dart` — a new FFI barrel export that extends `at_chops.dart` with OpenSSL-backed algorithm implementations and the `AtPqc` namespace
-`at_chops.dart` no longer exports the FFI algorithm classes at all —`at_chops_ffi.dart` is the only way to reach them
- Adds `AtPqc` — an abstract final namespace that lazily probes `libcrypto` once and exposes immutable singleton backends for X-Wing (`AtPqc.xWing`) and ML-DSA-65 (`AtPqc.mlDsa65`), auto-selecting FFI or pure-Dart fallback for encapsulate/decapsulate/sign/verify and key generation
- `AtSigningAlgorithm` — annotated `@sealed` with a doc note directing new code to `AtSignatureAlgorithm`
- Adds `AtSignatureAlgorithm` — a new stateless signing interface (`signBytes`/`verifyBytes`, key material passed per call). `MlDsa65PureDartAlgo` and `MlDsa65FfiAlgo` implement it directly.
- Adds `XWingKeyPair.publicKeyBytes`/`privateKeyBytes` and `MlDsa65KeyPair.publicKeyBytes`/`privateKeyBytes` — callers no longer manually `base64Decode` the `AtPublicKey`/`AtPrivateKey` strings

### How
- `at_chops.dart` remains web-safe (pure-Dart only, verified via `dart compile js` on a barrel-only entrypoint); `at_chops_ffi.dart` is native-only (`dart:ffi` dependency)
- `MlDsa65FfiAlgo`/`MlDsa65PureDartAlgo` implement `AtSignatureAlgorithm` with matching `(message, key)` order on both backends.
- Keygen stays pure-Dart in the key pair classes (`XWingKeyPair.generate`, `MlDsa65KeyPair.generate`, `AtChopsUtil.generateXWingKeyPair`, `AtChopsUtil.generateMlDsa65KeyPair`) — these are exported from the pure-Dart `at_chops.dart` barrel and must not import `AtPqc`, since `AtPqc` pulls in `dart:ffi` transitively and would break `dart compile js`/wasm for web consumers. `AtPqc.xWing`/`AtPqc.mlDsa65` (FFI-accelerated keygen) are only reachable via the opt-in `at_chops_ffi.dart` barrel.

### How to verify
[...]

PR #2039, opened 2026-06-30
feat(at_chops): aes gcm ffi + tests

JeremyTubongbanua:
## What

Adds an OpenSSL-backed **AES-256-GCM** FFI implementation (`AesGcm256FfiAlgo`) to
at_chops, wire-compatible with the existing pure-Dart `AesGcm256EncryptionAlgo`,
and wires it into the `AtPqc` auto-resolver as `AtPqc.aesGcm256(key)`.

AES-GCM is the AEAD layer of the PQ-HPKE construction, so it is now resolved
through `AtPqc` exactly like `AtPqc.xWing` and `AtPqc.mlDsa65` — callers get the
FFI backend when libcrypto is available and the pure-Dart backend otherwise,
without having to pick manually.

## How

- `AesGcm256FfiAlgo.fromLib(lib, key)` — binds `EVP_aes_256_gcm` /
  `EVP_CIPHER_CTX` via `dart:ffi`. Same wire format (`ciphertext || 16-byte tag`),
  12-byte nonce, and AAD semantics as the pure-Dart algo, so the two interoperate.
- `AtPqc.aesGcm256(key)` auto-selects FFI vs pure-Dart via the new
  `libCryptoSupportsAesGcm` probe. It's a **factory method** (not a static field
  like `xWing`/`mlDsa65`) because AES-GCM needs a key at construction time.
- Exported from `at_chops_ffi.dart` only (not the web-safe `at_chops.dart`
  barrel), since it carries `dart:ffi` bindings.
- New FFI bindings for `EVP_CIPHER_CTX` and the GCM control constants.

## How to verify

- `cd packages/at_chops && dart test` — all 208 tests pass (requires libcrypto).
- Coverage includes: encrypt/decrypt round-trip, empty plaintext, tampered
  ciphertext/tag, wrong key/nonce, AAD match/mismatch, NIST GCM vectors, and
  FFI↔pure-Dart interop in both directions, plus `AtPqc.aesGcm256` selection.

## Skills impact

- [x] No public API change in `at_client`/`at_client_flutter`/`at_auth` — skill unaffected

---
**PQ plan conformance (§10(e)):** project **P-2** — [implementation-plan.md](https://github.com/atsign-foundation/at_client_sdk/blob/trunk/docs/projects/pq/implementation-plan.md), tracking epic #1889.


JeremyTubongbanua:
Will wait for #2030 to be merged first

gkc:
Per the PQ plan's §10(e) conformance rule, this PR is project **P-2** (see [implementation-plan.md](https://github.com/atsign-foundation/at_client_sdk/blob/trunk/docs/projects/pq/implementation-plan.md) and tracking epic #1889).

packages/at_chops/lib/src/algorithm/at_pqc.dart
```diff
@@ -26,4 +30,18 @@ abstract final class AtPqc {
...
+  /// treated as a PQ-adjacent primitive and resolved through [AtPqc] like the
+  /// other backends.  Because the algorithm requires a key at construction time
+  /// this is a factory method rather than a static field.
+  ///
+  /// Do not downcast to the concrete type — it is an implementation detail.
+  static SymmetricEncryptionAlgorithm<Uint8List, Uint8List> aesGcm256(
+          AESKey key) =>
+      (_lib != null && libCryptoSupportsAesGcm(_lib!))
+          ? AesGcm256FfiAlgo.fromLib(_lib!, key)
+          : AesGcm256EncryptionAlgo(key);
```
JeremyTubongbanua, line 50: @srieteja FFI is used when available, otherwise falls back to pure-Dart. 

Just like XWing and MLDSA65

packages/at_chops/lib/src/algorithm/at_pqc.dart
```diff
@@ -26,4 +30,18 @@ abstract final class AtPqc {
...
+
+  /// AES-256-GCM authenticated encryption — FFI when available, else pure-Dart.
+  ///
+  /// AES-256-GCM is the AEAD layer in the PQ-HPKE construction, so it is
+  /// treated as a PQ-adjacent primitive and resolved through [AtPqc] like the
+  /// other backends.  Because the algorithm requires a key at construction time
+  /// this is a factory method rather than a static field.
+  ///
+  /// Do not downcast to the concrete type — it is an implementation detail.
+  static SymmetricEncryptionAlgorithm<Uint8List, Uint8List> aesGcm256(
```
gkc, line 46: **`AtPqc.aesGcm256` can't serve the AEAD-with-AAD use case it exists for (medium).** The factory returns `SymmetricEncryptionAlgorithm<Uint8List, Uint8List>`, whose interface `encrypt`/`decrypt` declare only `{InitialisationVector iv}` — no `aad`. But AES-GCM was added (per the CHANGELOG/README) as the AEAD layer of PQ-HPKE, and HPKE seal/open pass AAD. So `AtPqc.aesGcm256(key).encrypt(plain, iv: iv, aad: header)` is a compile error on the static type, and `aad` is reachable only by downcasting to the concrete class — which this method's own dartdoc (line 41) forbids. As a result `pq_hpke.dart` `pqSeal`/`pqOpen` construct the pure-Dart `AesGcm256EncryptionAlgo` directly, and nothing in the tree calls `AtPqc.aesGcm256` — the FFI backend is never exercised for its one intended workload.

You should pick either: (a) wire `pq_hpke` through `AtPqc.aesGcm256` and expose `aad` (widen the interface, or change the factory's return type to one that carries `aad`); or (b) drop the PQ-HPKE-routing rationale from the docs and describe this as an available-but-not-yet-wired backend.

packages/at_chops/README.md
```diff
@@ -172,6 +172,8 @@ ML-DSA-65, ML-KEM-768, and X25519 each have an OpenSSL FFI backend (`MlDsa65FfiA
 
 X-Wing (`XWingFfiAlgo`) composes the FFI backends for maximum performance when `libcrypto` is available.
 
+AES-256-GCM also has an OpenSSL FFI backend (`AesGcm256FfiAlgo`) alongside its pure-Dart counterpart (`AesGcm256EncryptionAlgo`); the two are fully interoperable. As the AEAD layer of the PQ-HPKE construction, it is resolved through `AtPqc` like the other PQ backends — prefer `AtPqc.aesGcm256(key)`, which auto-selects FFI or pure-Dart, over constructing `AesGcm256FfiAlgo.fromLib(lib, key)` directly.
```
gkc: Consequence of the `at_pqc.dart` comment (low): this says AES-GCM "is resolved through `AtPqc`" and to "prefer `AtPqc.aesGcm256(key)`", but `pq_hpke.dart` hardcodes the pure-Dart `AesGcm256EncryptionAlgo` and there is no caller of `AtPqc.aesGcm256` anywhere — so on a libcrypto-present host PQ-HPKE still runs pure-Dart AES-GCM, and the "prefer `AtPqc.aesGcm256`" advice doesn't compile when AAD is needed. Please reconcile this with whichever option you pick above.

packages/at_chops/lib/src/algorithm/encryption/aes_gcm_ffi_algo.dart
```diff
@@ -0,0 +1,312 @@
...
+      final Pointer<Uint8> keyPtr = calloc<Uint8>(keyBytes.length);
+      final Pointer<Uint8> ivPtr = calloc<Uint8>(nonce.length);
+      keyPtr.asTypedList(keyBytes.length).setAll(0, keyBytes);
+      ivPtr.asTypedList(nonce.length).setAll(0, nonce);
+      try {
+        if (_encryptInitEx(ctx, nullptr, nullptr, keyPtr, ivPtr) <= 0) {
+          throw StateError('EVP_EncryptInit_ex (key/IV) failed');
+        }
+      } finally {
+        calloc.free(keyPtr);
```
gkc, line 106: Secret zeroization (low): the raw 32-byte AES key (`keyPtr` here and at line 220) and the recovered plaintext (`outBuf`, returned at line 279) are `calloc.free`'d without wiping, so the key/plaintext linger in freed native heap (exposed by a later core dump / swap / heap reuse). This matches the sibling pattern (x25519 frees its private key the same way), so it's consistency-level hygiene rather than a regression — but this is the first raw *symmetric*-key FFI buffer, and native memory is trivial to wipe (`keyPtr.asTypedList(len).fillRange(0, len, 0)` before free). Worth doing here, with a follow-up sweep for the siblings.

packages/at_chops/lib/src/algorithm/ffi/openssl_ffi_bindings.dart
```diff
@@ -167,3 +182,58 @@ const int evpPkeyKeypair = 0x87;
...
+typedef EvpEncryptInitExDart = int Function(Pointer<EVP_CIPHER_CTX>,
+    Pointer<EVP_CIPHER>, Pointer<Void>, Pointer<Uint8>, Pointer<Uint8>);
+
+typedef EvpDecryptInitExNative = Int32 Function(Pointer<EVP_CIPHER_CTX>,
+    Pointer<EVP_CIPHER>, Pointer<Void>, Pointer<Uint8>, Pointer<Uint8>);
+typedef EvpDecryptInitExDart = int Function(Pointer<EVP_CIPHER_CTX>,
+    Pointer<EVP_CIPHER>, Pointer<Void>, Pointer<Uint8>, Pointer<Uint8>);
+
+typedef EvpEncryptUpdateNative = Int32 Function(Pointer<EVP_CIPHER_CTX>,
+    Pointer<Uint8>, Pointer<Int32>, Pointer<Uint8>, Int32);
```
gkc: Int32 length bound (low): the `inl` (input length) parameter of `EVP_EncryptUpdate`/`EVP_DecryptUpdate` is bound as `Int32`, so a plaintext/ciphertext/AAD ≥ 2 GiB fed in a single call is silently truncated to 32 bits — data loss or a spurious failure, and a divergence from the pure-Dart backend. Not realistic for at_chops payloads, but `IntPtr`/`Size` matches OpenSSL's `size_t` and removes the cliff.

packages/at_chops/lib/src/algorithm/encryption/aes_gcm_ffi_algo.dart
```diff
@@ -0,0 +1,312 @@
...
+  @override
+  Future<Uint8List> decrypt(Uint8List encryptedData,
+      {InitialisationVector? iv, List<int> aad = const []}) async {
+    if (encryptedData.length < tagLength) {
+      throw AtDecryptionException(
+          'AES-256-GCM (FFI) input shorter than the $tagLength-byte tag');
+    }
+
+    final Uint8List keyBytes = _keyBytes();
+    final List<int> nonce = _nonceBytes(iv);
```
gkc: Exception-type consistency (low): on the decrypt path, `_keyBytes()`/`_nonceBytes()` throw `AtEncryptionException` (lines 298/306) for a non-256-bit key or a missing/wrong-length IV. A caller handling attacker-influenced input with the natural `try { await decrypt(...) } on AtDecryptionException` pattern would let these escape. Consider throwing `AtDecryptionException` for the decrypt-side validation (or a shared base type) so the decrypt failure surface is one catchable type.

packages/at_chops/lib/src/algorithm/ffi/openssl_loader.dart
```diff
@@ -80,6 +81,20 @@ bool libCryptoSupportsMlDsa65(DynamicLibrary lib) {
   return _libCryptoSupportsAlgorithm(lib, 'ML-DSA-65');
 }
 
+/// Returns `true` when [lib] exposes `EVP_aes_256_gcm` (available in every
+/// OpenSSL 1.0+ build, so this essentially just confirms the symbol resolves).
+///
+/// Call this before constructing [AesGcm256FfiAlgo] to confirm the symbol is
+/// present — it is effectively always true for any usable libcrypto.
+bool libCryptoSupportsAesGcm(DynamicLibrary lib) {
```
gkc, line 94: Probe strength (low): `libCryptoSupportsAesGcm` returns true whenever the `EVP_aes_256_gcm` symbol resolves, without checking the cipher is usable — unlike the ML-DSA/ML-KEM probes, which fetch the algorithm by name. On a FIPS-/policy-restricted libcrypto where the symbol is exported but the cipher is disabled at init, `AtPqc.aesGcm256` would select the FFI backend and then throw a bare `StateError` at encrypt/decrypt instead of falling back to pure-Dart. Matching the fetch-by-name probe pattern would make the fallback robust.

gkc (changes requested):
Reviewed the AES-256-GCM FFI addition — the crypto itself is solid: decrypt fails closed on tag mismatch (returns plaintext only after `EVP_DecryptFinal_ex` verifies), the nonce is mandatory (no fixed-nonce fallback), the key is validated at 32 bytes, the wire format matches the pure-Dart backend byte-for-byte, and every EVP return is checked with all buffers/ctx freed in `finally`. All CI is green.

Requesting changes on one substantive item: the new FFI backend and the `AtPqc.aesGcm256` factory aren't reachable for their stated purpose (AEAD-with-AAD / PQ-HPKE) — see the comment on `at_pqc.dart`. The rest are low-severity polish (secret zeroization, an `Int32` length bound, a decrypt-side exception-type inconsistency, and a probe that only checks symbol resolution). Details inline.

packages/at_chops/lib/src/algorithm/ffi/openssl_ffi_bindings.dart
JeremyTubongbanua: Changed to IntPtr https://github.com/atsign-foundation/at_client_sdk/pull/2039/commits/5040be10b09b342af165ffd6a910a44e9a88fcc4

packages/at_chops/lib/src/algorithm/at_pqc.dart
JeremyTubongbanua: Went with option (b)

See https://github.com/atsign-foundation/at_client_sdk/pull/2039/commits/a43a0378b81c3c5a6b6cd48514694cae1cbcbc54

packages/at_chops/README.md
JeremyTubongbanua: Since we went with option (b) (above) 

See https://github.com/atsign-foundation/at_client_sdk/pull/2039/commits/fbd42d8725e793d50cbfa1ca82968ce2ba2d69cf

packages/at_chops/lib/src/algorithm/encryption/aes_gcm_ffi_algo.dart
JeremyTubongbanua: See https://github.com/atsign-foundation/at_client_sdk/pull/2039/commits/6ca384a01bf17f33d7c9be1fcdab4201f7928775

RAW AES key is now wiped before free

packages/at_chops/lib/src/algorithm/ffi/openssl_loader.dart
JeremyTubongbanua: Fixed in https://github.com/atsign-foundation/at_client_sdk/pull/2039/commits/cbea58e6e8181d7ffa508a6cb2679d3851992be3

fetch by name now probe pattern now matches what ML DSA/ML KEM does

packages/at_chops/lib/src/algorithm/encryption/aes_gcm_ffi_algo.dart
JeremyTubongbanua: Fixed in https://github.com/atsign-foundation/at_client_sdk/pull/2039/commits/6ca384a01bf17f33d7c9be1fcdab4201f7928775

decrypt-side key/IV validation now throws `AtKeyDecryption` instead of `AtEncryptionException` . Decrypt failure is now caught by one type./

srieteja (approved):

gkc (approved):
LGTM

Xlin123 (approved):

Issue #2045, opened 2026-07-02
AtKeys decision change


Xlin123:
## Finalized decision for AtKeys
AtKeys will be extended additionally with pq methods and the legacy methods will be deprecated.
AtKeysIo will be extended to introduce runtime persistence (ie: append(), save() etc)
AtKeysIo remains the contact point of AtKeys to keep runtime objects and persisteted objects in-line.

## What this allows us to do moving forward
- Less code churn, code contract will remain the same; contains deprecated fields and methods.
- Providers expect 3 items: an AtClient, an AtKeysIo and access to AtChops
- Deprecation path is CLEAR CUT, and in-line with the changes we're trying to accomplish:
  - Keep legacy keys around, introduce new pq-safe methods
  - deprecate usage of legacy keys, to push call sites to use pq-safe methods

## Future questions to be answered
 - What is the actual point of keeping AtKeys inside the AtClient scope?
   - at encrypt/decrypt time we have the AtKeysIo to inject the provider with our necessary keys
   - at auth time, we have the required AtKeysIo to inject our authentication with necessary keys as well
   - at enrollment time, is an odd case for us since the enrollment structure is a bit immature in comparison to auth, but should be easily fixed.
   - is there any other times we need to think about? Otherwise, the AtClient needs 0 concept of AtKeys at all outside of the provider...

Xlin123:
tagged you all for visibility

gkc:
Am updating docs pack to reflect this decision

gkc:
Closing — the decision record's deliverable is complete. The finalized AtKeys direction (extend the existing `AtKeys` / `AtKeysIo` in place with PQ-safe additive methods, deprecating the legacy fields) is captured in `docs/projects/pq/decisions.md` (commit d662f71b7), matching the 2026-07-06 discussion. The follow-on *code* work — actually extending `AtKeys`/`AtKeysIo` and deprecating the legacy fields — is tracked separately as S-1 (#2005). Closing this decision-record issue as recorded.

PR #2047, opened 2026-07-03
feat(at_auth): deprecate hard-coded atkeys fields and introduce new methods / serialization

Xlin123:
## Summary

This PR introduces the next AtKeys persistence shape while keeping legacy
`.atKeys` files fully working. `AtKeys` now carries two kinds of state:

- the legacy hard-coded fields, preserved for current auth/onboarding code paths
- typed `AtKeysMaterial` records — self-describing (keyId, enrollmentId,
  `CryptographicKeyType`, `KeyAlgorithmType`, bytes, operations, status),
  PQ-ready, and the only key type the new `AtKeys` API deals in

Persistence is a versioned JSON document produced/consumed directly by
`AtKeys.toJson()`/`fromJson()` — the codec/document/resolver layer from
earlier iterations of this branch was removed; validation lives in
`AtKeysAssurance` instead.

## What changed

**Typed key API on `AtKeys`**
- `addKey` / `getKey(keyId, type)` / `keysForKeyId` / `keysForEnrollment`
- `retireKey(keyId, {to})` — the delete operation. Material is never removed
  (retired bytes are still needed to decrypt old data); status moves
  forward-only: `active` → `retired` → `dead`
- equality/hashCode no longer depend on insertion order of materials

**Serialization**
- `toJson()`/`fromJson()` emit/accept the versioned typed-keys document
  (`version`, `atSign`, `keys[].keyParts[]`, legacy fields flat at the top
  level). `fromJson` falls back to the legacy flat shape when there is no
  `version` field, and throws `AtKeysUnsupportedVersionException` on an
  unknown version
- `encodeAtKeysDocument`/`parseAtKeysDocument` carry the wire shape and
  cross-material validation
- at rest, a typed-keys document self-encrypts its legacy RSA fields exactly
  like a legacy-only file (same key, same deterministic IV) — the legacy
  portion is byte-identical, so legacy readers still work

**IO**
- new `flush(Atsign, AtKeys)` — the runtime persist: mutate in memory, then
  flush. On an existing file it safety-checks the rewrite via
  `AtKeysAssurance.validateMapUpdate` (nothing may be lost; statuses may only
  move forward; new materials may be added) and upgrades legacy files in
  place to the typed-keys document. `write()` stays the create-only initial
  persist
- IO implementations moved under `src/keys/io/`; added `InMemoryAtKeysIo`
  for in-memory/test flows
- passphrase encryption moved from `at_chops` into `at_auth`
  (`AtKeysPassphraseEnvelopeCodec`, argon2id)

**Legacy compatibility preserved**
- bare legacy `.atKeys` JSON still decodes as legacy
- legacy self-encrypted files still go through the old decrypt path

## Testing

60 new tests (63 cases counting passphrase parameterization) across
`assurance_test`, `at_keys_io_test`, `at_keys_test`, `legacy_at_keys_test`,
`memory_io_test`, and `passphrase_envelope_test`, including 10 round-trip
tests covering typed keys, legacy files, passphrase-protected files, legacy →
typed-keys upgrades, and the passphrase envelope. 

## Compatibility notes

This is a transition step, not a clean break. Legacy fields and read/write
behavior are preserved; newer typed material flows through the document path
while existing auth/onboarding call sites migrate. Some legacy fields are now
deprecated for the eventual v4 cleanup but are not removed here. In v4 the
`.atKeys` shape will hold legacy in its own field and write all new keys
natively.

## Publishing order
1. at_chops 3.4.1
2. at_auth 3.2.0

## Skills impact

---
**PQ plan conformance (§10(e)):** project **S-1** — [implementation-plan.md](https://github.com/atsign-foundation/at_client_sdk/blob/trunk/docs/projects/pq/implementation-plan.md), tracking epic #1889.


packages/at_auth/lib/src/keys/io/at_keys_io.dart
```diff
@@ -1,37 +1,47 @@
...
 import 'package:at_auth/src/auth_constants.dart' as auth_constants;
-import 'package:at_chops/at_chops.dart';
+import 'package:at_chops/at_chops.dart' hide AtKeysCrypto;
 import 'package:at_commons/at_commons.dart';
 import 'package:at_utils/at_utils.dart' show AtSignLogger;
 
 /// An interface that defines methods for reading AtKeys.
 /// It can be implemented by classes that read AtKeys from different sources,
 sealed class AtKeysIo {
+  final resolver = const AtKeysDocumentResolver();
```
Xlin123: providing all serialization related classes in the AtKeysIo for consumption

packages/at_auth/lib/src/keys/io/file_io.dart
```diff
@@ -28,18 +29,18 @@ class FileAtKeysIo extends WrittenAtKeysIo {
...
       throw AtException(
           'provided keys file does not exist. Please check whether the file path $file is valid');
     }
     String atAuthData = await File(file).readAsString();
-    decodedAtKeysData = jsonDecode(atAuthData);
-    decodedAtKeysData =
-        await decodeAtKeys(decodedAtKeysData, passPhrase: passPhrase);
-    return decryptAtKeysWithSelfEncKey(
-        decodedAtKeysData, PkamAuthMode.keysFile);
+    Map<String, dynamic> json = jsonDecode(atAuthData);
```
Xlin123: in essence, these classes allow IO to ONLY worry about IO not about codec/resolution/validation

packages/at_auth/lib/src/keys/io/file_io.dart
```diff
@@ -28,18 +29,18 @@ class FileAtKeysIo extends WrittenAtKeysIo {
...
+    if (passwordCodec.isEnvelope(json)) {
+      json = await passwordCodec.decode(json, passPhrase: passPhrase);
+    }
+    AtKeysDocument document = codec.decodeDocument(json);
+    return resolver.resolve(document);
```
Xlin123: can follow the code path between these two methods:

decode document: looks for legacy and new format, decodes and provides both formats in the document object.

resolve document: turns the document fields into AtKeys, and ensures legacy and new are provided in the right ways.

packages/at_auth/lib/src/keys/at_keys.dart
```diff
@@ -1,18 +1,56 @@
...
+    T key,
+  ) {
+    final keysForType = _keysByType.putIfAbsent(T, () => {});
+    if (keysForType.containsKey(id)) {
+      throw ArgumentError.value(id, 'primaryId',
+          'Duplicate $T with their unique privateId, see AtKeys ctor for the respective id');
+    }
+    keysForType[id] = key;
+    _keyMaterials.add(key);
+  }
```
Xlin123: these generic methods are also the interaction layer for the "snapshot" of the atkeys at runtime

so if I'm appending a new cryptographic key, I do AtKeysIo.append() and inside append() we can update the runtime object as well.

packages/at_auth/lib/src/keys/serialization/atkey_material.dart
Xlin123, line 1: all the code types for Keys, 

AtPublicKey, AtPrivateKey, AtSymmetricKey and AtKeyPackage

keypairs can be assembled at provider level as they only act as a collective object ie kp.public / kp.private

packages/at_auth/test/legacy_at_keys_test.dart
Xlin123, line 1: test for: does legacy survive round trip

gkc:
Per the PQ plan's §10(e) conformance rule, this PR is project **S-1** (see [implementation-plan.md](https://github.com/atsign-foundation/at_client_sdk/blob/trunk/docs/projects/pq/implementation-plan.md) and tracking epic #1889).

gkc:
Heads-up: the AtKeys direction is now ratified in #2045 (decisions.md, 2026-07-06) and reflected in the plan (project **S-1**): **keep the existing `AtKeys` hierarchy and extend in place** — additive PQ-safe methods + deprecate the legacy fields/methods; extend `AtKeysIo` with runtime persistence (`append()`/`save()`) as the runtime↔persisted contact point; providers are injected `(AtClient, AtKeysIo, AtChops)`. No new `WritableAtKeys` holder. Worth aligning this PR's naming/shape with that before it leaves draft — see [implementation-plan.md](https://github.com/atsign-foundation/at_client_sdk/blob/trunk/docs/projects/pq/implementation-plan.md) S-1 and decisions.md.

gkc (changes requested):
I've not looked at this super deeply yet but a few things

(1) don't update existing tests. Where we were creating AtKeys(), continue to do so. Write new tests to cover new stuff
(2) safety first. Worst possible outcome is bricking or deleting a customers AtKeys. You will need lots of round trip tests; and I would suggest we need a standard safety-assured process when writing which is : write to "temporary" file or keychain entry or whatever; then assert that everything in the existing file (or whatever) is 100% preserved identical in the new one, unless we were explicitly deleting something, and that the new one parses correctly etc, and only once that's all done do we save the new version. Should consider an archive of previous versions also (for files, keep copy with utc date time suffix yyyymmdd.hhmmss.mmmmmm)

(3) make sure you are keeping the existing api no breaking changes ... copyWith being removed is breaking ...

gkc:
PS the safety assurance could prob be done in memory

packages/at_auth/lib/src/keys/io/at_keys_io.dart
```diff
@@ -10,14 +13,32 @@ import 'package:at_utils/at_utils.dart' show AtSignLogger;
...
+
+  /// Persists [atKeys] as the complete new state for [atsign].
+  ///
+  /// This is the runtime counterpart to [write]: mutate the in-memory
+  /// [AtKeys] (e.g. [AtKeys.addKey]), then flush the whole object.
+  /// Implementations backed by durable storage must not lose data: when a
+  /// target already exists, validate that everything in it is preserved in
+  /// [atKeys] (see [AtKeysAssurance.validateMapUpdate]), then rewrite. When
+  /// no target exists, flush creates it — there is nothing to lose.
+  FutureOr<void> flush(Atsign atsign, AtKeys atKeys) {
```
Xlin123, line 47: throw Unimplemented for flush as it would cause BC downstream, in 4.0.0 will be standard.

packages/at_auth/lib/src/keys/at_keys_io_impl.dart
Xlin123, line 1: lost on rename to io/at_keys_io.dart

packages/at_auth/lib/src/keys/at_keys.dart
```diff
@@ -1,36 +1,214 @@
...
+      }
+      return _toLegacyJson();
+    }
+    return {
+      ..._toLegacyJson(),
+      'version': supportedVersion,
+      'atSign': atsign.toString(),
+      'keys': encodeAtKeysDocument(keys),
+    };
+  }
```
Xlin123, line 147: key note: if we're able to write the new fields we will during write. however during flush we HAVE to be able to serialize the typed atKeys.

file_io is routing the atsign through to the atkeys, so "theoretically" shouldn't produce any runtime errors before 4.0.0

pointing this out for visibility cause it feels icky

Xlin123:
@gkc @srieteja fine to review whenever, and adding @JeremyTubongbanua in as well.

gkc:
@Xlin123 can you resolve the changelog conflicts please

gkc (commented):
This is really good @Xlin123 ... I have a lot of pretty small changes to recommend so rather than do it all via iterating on this PR, I'll do a PR to this PR with comments explaining my thinking

Xlin123:
@gkc make sure to publish at_auth first before merging this

once published we i'll squash on merge and then have a follow up PR for rc1

gkc:
> @gkc make sure to publish at_auth first before merging this
> 
> once published we i'll squash on merge and then have a follow up PR for rc1

@Xlin123  ack

gkc (approved):

srieteja (approved):

JeremyTubongbanua (approved):

PR #2063, opened 2026-07-11
fix(at_client): resolve received (cached:) collection items (#2032)

### LINKED ISSUE #2032: Query.watch() never re-emits for received/shared items (delta path getOrNull misses cached: keys)
### Describe the bug

**Package:** `at_client` **3.12.0**
**Area:** `AtCollection<T>` / `Query<T>.watch()` real-time event pipeline
**File:** `packages/at_client/lib/src/collections/collections.dart`

A live `Query.watch()` stream **fails to re-emit when a received (shared-in) item changes**, even though the underlying notification arrives and the collection fires a `CItemUpdated` event. The result: in a collaborative app, when atSign **B** shares/updates an item with atSign **A**, **A's `watch()`-driven UI never updates** for that item (until the app is restarted or the watch is re-created with an empty cache). Items the current atSign **owns** update fine; only **received/shared** items are affected.

### Root cause

`Query.watch()` has two code paths in its `onUpdate` handler (`collections.dart:3944`):

```dart
Future<void> onUpdate(Atsign owner, String id) async {
  if (!usesDeltaPath || cache == null) {
    return refresh();                                   // FULL refetch via getItems()
  }
  ...
  fresh = await _collection.getOrNull(id, owner);       // DELTA path — collections.dart:3953
  final passes = fresh != null && matchesPredicates(fresh);
  if (passes) { ...add/replace... }
  else if (wasInCache) { ...remove... }
  else { return; }                                      // <-- silently no-ops here
}
```

`usesDeltaPath` is true whenever the query has no `limit`/`skip`. Once the initial fetch has populated `cache`, **every** subsequent event takes the delta path and calls `getOrNull(id, owner)`.

`getOrNull` → `getItemsAsStream(id:, owner:)` → `_getKeysInternal(id:, owner:)` (`collections.dart:1121`) builds a scan regex via `_directKeyRegex` (`collections.dart:1143`):

```dart
String _directKeyRegex({String? id, String ownerSuffix = '@'}) {
  final idPattern = id != null ? RegExp.escape(id) : '[^.]+';
  return '$_atKeyScanPrefix$idPattern\\.$_escapedNamespace$ownerSuffix';
}
```

with the shared prefix (`collections.dart:316`):

```dart
const String _atKeyScanPrefix = r'^(?!local:)(?:[^:]*:)?';
```

The prefix `(?:[^:]*:)?` permits **at most one** `xxx:` segment. The three on-disk key shapes for a collection item are:

| Relationship | Local key shape | leading `xxx:` segments |
|---|---|---|
| Self-owned | `<id>.<ns>@<self>` | 0 |
| Shared **by** me | `@<recipient>:<id>.<ns>@<self>` | 1 |
| **Received** (owned by other, shared to me) | `cached:@<self>:<id>.<ns>@<owner>` | **2** (`cached:` + `@<self>:`) |

For a **received** item with an explicit `id`, the regex consumes only `cached:` via `(?:[^:]*:)?`, then expects `<id>` but instead sees `@<self>:<id>…` → **no match** → `getAtKeys` returns nothing → `getOrNull` returns `null` → `onUpdate` hits the `else { return; }` branch → **the watch never re-emits**.

### Why it's intermittent / easily missed

- It only manifests on the **delta path** (`cache != null`). If the item arrives *before* the initial fetch completes, `onUpdate` takes the **`refresh()`** path (full `getItems()`), and `getItems()` is called with **`id == null`**, so `idPattern` is the wildcard `[^.]+`, which *does* absorb the `@<self>:<id>` segment and matches the `cached:` key. So the bug hides whenever the watch happens to be unprimed when the event lands.
- The **event itself fires correctly**: the event-dispatch filter `_regexObjAny` (`collections.dart:605`) uses `[^.]+(\.[^.]+)*` for the id portion, whose `[^.]+` *does* swallow `@<self>:<id>`, so `cached:` keys match it. Hence `collection.updates` emits `CItemUpdated` as expected — making it look like delivery works while `watch()` stays silent.

This split is the smoking gun: the **event filter** matches `cached:` keys but the **targeted read** does not.

### Proof (regex level)

```python
import re
key = "cached:@batman87:q127dgul.boards.teamboard@blizzard81"   # received item on @batman87
prefix = r'(?!local:)(?:[^:]*:)?'
ns = r'boards\.teamboard'

re.search('^'+prefix+r'[^.]+(\.[^.]+)*\.'+ns+'@', key)          # _regexObjAny (event filter)  -> MATCH
re.search('^'+prefix+r'q127dgul\.'+ns+'@blizzard81', key)       # _directKeyRegex(id, owner)    -> None  ❌
re.search('^'+prefix+r'[^.]+\.'+ns+'@', key)                    # _directKeyRegex(id=null)      -> MATCH (masks bug)
```

Output: event filter → match, `getOrNull(id, owner)` → **no match**, `getItems()` (no id) → match.

### Steps to reproduce

Two atSigns, **A** (`@blizzard81`) and **B** (`@batman87`), both running an app that uses `AtCollection<T>` with `EventSource.both` and renders a `query().watch()` stream.

1. On **B**, create the collection and start a `query().watch()` (e.g. in a `StreamBuilder`). Let the initial fetch complete so the watch's internal `cache` is **non-null** (B already has ≥1 item, or simply wait a moment after first frame).
2. On **A**, create a new item in the same collection and share it with **B** (`create(obj: …, sharedWith: {B})`), or update an item already shared with B.
3. Observe **B**:
   - The notification arrives: `NotificationServiceImpl (B)|Received B:<id>.<ns>@A`.
   - The collection ingests it: `AtCollection<…> | Updating cached:B:<id>.<ns>@A`.
   - `collection.updates` **fires** a `CItemUpdated(owner: A, id: <id>)`.
   - **But `query().watch()` does NOT re-emit**, so B's UI does not update.
4. Restart B's app (or otherwise re-create the watch so its cache starts empty). On the first event after that, the watch takes the `refresh()` path and the item **does** appear — confirming the data is present locally and only the delta-path read is at fault.

### Minimal programmatic reproduction (no UI)

On B, after the watch is primed, the targeted read used by the delta path returns null for a received item while the unfiltered read returns it:

```dart
final all = await boards.getItems();                 // includes the received item ✅
final one = await boards.getOrNull(receivedId, ownerAtSign); // null ❌  <-- the bug
// (watch().onUpdate uses getOrNull, so the live stream silently no-ops)
```

### Expected behavior

`Query.watch()` should re-emit when a **received/shared** item changes, exactly as it does for self-owned items. Equivalently, `getOrNull(id, owner)` and `get(id, owner)` should resolve an item that is present in the local keystore as a `cached:@<self>:<id>.<ns>@<owner>` key. Real-time collaboration (the whole point of sharing a collection) should update the recipient's `watch()` UI without a restart.

Allow the scan prefix to match the **two** leading segments a received/shared `cached:` key carries (`cached:` + `@<self>:`), instead of only one.

**File:** `packages/at_client/lib/src/collections/collections.dart:316`

```dart
// before
const String _atKeyScanPrefix = r'^(?!local:)(?:[^:]*:)?';

// after — allow up to two colon-prefixed wrapper segments (e.g. `cached:@self:`)
const String _atKeyScanPrefix = r'^(?!local:)(?:[^:]*:){0,2}';
```

Verification of all three key shapes with the fixed prefix:

```python
import re
prefix = r'(?!local:)(?:[^:]*:){0,2}'
ns = r'boards\.teamboard'
tests = {
  "self":        "q127dgul.boards.teamboard@blizzard81",
  "shared-by-me":"@batman87:q127dgul.boards.teamboard@blizzard81",
  "received":    "cached:@batman87:q127dgul.boards.teamboard@blizzard81",
}
for name, key in tests.items():
    print(name, bool(re.search('^'+prefix+r'q127dgul\.'+ns+'@', key)))
# self True / shared-by-me True / received True   (all now match)
```

`{0,2}` keeps the match bounded (no catastrophic backtracking) and the trailing `\.<namespace>@<owner>` anchor is unchanged, so it does not widen matching to other namespaces or owners. The same prefix is shared by `_regexObjAny`, `_descendantKeyRegex`, and `_directKeyRegexForComposed`; the change is strictly more permissive in the wrapper region only and leaves their namespace/owner anchors intact.

### Suggested regression test

Add a test that primes a `query().watch()` cache, then injects a **received** item event (via the existing `collectionWithInjectedNotifications` / data-event test hooks) and asserts the watch **re-emits** including that item — i.e. exercise the delta path (`cache != null`) with a `cached:@<self>:<id>.<ns>@<owner>` key, not just the unprimed `refresh()` path.

### Screenshots

_No response_

### Smartphones

- Device: MacOS
- Package: at_client

_No response_

### Additional context

Real device logs showing the split (event fires, watch silent) on the **receiver** (`@batman87`), for a board `id=q127dgul` shared by `@blizzard81`:

```
NotificationServiceImpl (@batman87)|Received @batman87:q127dgul.boards.teamboard@blizzard81
AtCollection<Board> boards.teamboard |Updating cached:@batman87:q127dgul.boards.teamboard@blizzard81
boards.updates -> CItemUpdated(owner: @blizzard81, id: q127dgul)   <-- event fires
   ...no query().watch() re-emission follows...                    <-- bug
```

On the **creator** (`@blizzard81`, self-owned), the same sequence ends with the watch re-emitting and the UI updating — confirming the asymmetry is specific to received/`cached:` keys.

Affected public APIs (all route through `_getKeysInternal` → `_directKeyRegex` with an explicit `id`):
- `Query.watch()` (delta path) — the main user-visible failure.
- `getOrNull(id, owner)` / `get(id, owner)` for received items.

Not affected (and thus masks the bug):
- `getItems()` (no `id`) and the watch's `refresh()` path — the id wildcard `[^.]+` absorbs the `@<self>:` segment.
- `collection.updates` / `subUpdates` event streams — they use `_regexObjAny`, whose `[^.]+` id portion matches `cached:` keys.

gkc:
## What I did

**Note to reviewers** TL;DR : Some small code changes, largely regex tightening, better validation, and
simple logic fixes. Lots of new tests for coverage.

Fixes #2032 — a live `Query.watch()` never re-emitted when a **received**
(shared-in) item changed. Items stored locally as
`cached:@<self>:<id>.<ns>@<owner>` were missed by the id-scoped read path
(`Query.watch()` delta path, `getOrNull`, `get(id, owner)`), so a recipient's
UI never updated for a shared item until the app restarted.

While fixing it, an adversarial review of `AtCollection` surfaced **six further
defects**, all fixed in this PR. 

- **Owner cross-match / delete data-loss.** Owner-scoped scans and deletes were
  not end-anchored on the owner, so a concrete owner (`@bob`) prefix-matched a
  longer atSign (`@bobby`) — returning another atSign's items, and on delete
  destroying a same-id received copy. This one is **coupled to the #2032 fix**:
  widening the scan prefix newly exposed it, so it ships here to avoid trading
  one bug for another.
- **`exists(id, owner)` routed received items to a remote lookup** — it diverged
  from `getOrNull` and returned the wrong answer offline.
- **`cleanupOrphans()` could delete a live item** under a received parent when
  offline, because the ancestor presence check routed to a remote lookup
  instead of reading the local cached copy.
- **Dotted top-level ids** were stored intact but read back truncated at the
  first dot (`create(id: '2024.01')` came back as `2024`).
- **A reader's own `markReadByMe`** fired a spurious self `CReadReceipt` on the
  data-event path.
- **`availableEvents` never fired** for a future-dated item that arrived via an
  update event — the scheduler read filtered those items out, and the
  value-less placeholder crashed for a non-nullable item type.
- **A key expiring or being deleted mid-read** duplicated the preceding item.

## How I did it

Brief — see the commit for details:

- Widened the shared scan prefix to allow the two wrapper segments a received
  key carries, and end-anchored concrete owners.
- Made `exists` and the orphan ancestor-check read the local `cached:` copy.
- Guarded write-time ids and the self read-receipt; routed the availability
  scheduler through the not-yet-available read path and gave `CItem` a
  value-less placeholder constructor; reset the read buffer on a mid-stream skip.

## How to verify it

Tests pass — `dart test --concurrency=1` in `packages/at_client` (full unit
suite green). New / added test cases:

**`test/at_collections_received_items_test.dart` (new)**
- #2032: `getOrNull` resolves a received item; the id-scoped scan regex matches
  the `cached:@self:` shape; `getItems()` (wildcard) documents the masking
  asymmetry; `watch()` re-emits a received item that arrives after priming.
- Owner anchor: `getItems(owner:)` does not cross-match a prefix-extended
  atSign; deleting a self item spares a same-id received copy.
- `exists`: resolves a received item from the local store; agrees with
  `getOrNull`.
- Dotted ids: `draft` and `create` reject an id containing `.`.
- Scheduler: a future-dated item arriving via an update event still fires
  `CItemAvailable`.
- Read dedup: an expiry/delete race on a middle key does not re-yield its
  predecessor.

**`test/at_collections_sub_test.dart`**
- A self-owned descendant under a received parent survives cleanup when the
  parent's cached copy is present locally.

**`test/at_collections_data_events_test.dart`**
- A reader's own outgoing `__rr` write fires no `CReadReceipt`; an incoming
  receipt from another atSign still fires.

`AtCollection`: resolve received (shared-in) items in id-scoped reads
(`Query.watch()`, `getOrNull`, `get`, `exists`); end-anchor owner-scoped
scans/deletes; fix `cleanupOrphans` offline over-deletion, dotted-id
truncation, a spurious self read-receipt, missed `availableAt` firings, and a
duplicate-yield read race.


CurtlyCritchlow (dismissed):
I tested this branch on my teamboard app and the UI is updating in realtime when CItem is shared with another atsign or a shared CItem is deleted.

gkc:
@CurtlyCritchlow can you re-approve this PR please? I had to resolve a merge conflict and bump the version number because we've now published at_client 3.13.0

srieteja (approved):

athandle (approved):

PR #1923, opened 2026-04-29
fix(at_client_flutter): prevent dialogs form hanging and improve api key error handling

Xlin123:
**- What I did**
- Fixed dialog failure handling so PKAM/CRAM/APKAM flows don’t hang.
- Added clear registrar API key errors (blank key + 401/403 auth failures).
- Removed example app’s global error modal so widget snackbar handling is used.

**- How I did it**
- Wrapped dialog async calls in `try/catch`; on error: show snackbar, dismiss dialog, return `null`.
- Added registrar API key validation and mapped 401/403 to actionable `AtException`.
- Updated tests for registrar auth errors and dialog failure flows.

**- How to verify it**
- run new tests
- try and reproduce using #1909 

**- Description for the changelog**
- fix: prevent auth/onboarding dialogs from hanging on exceptions and improve registrar API key error handling

Issue #1909, opened 2026-04-21
PkamDialog, CramDialog, and ApkamActivationDialog silently hang forever when authentication throws


CurtlyCritchlow:
### Describe the bug

Three dialog widgets in at_client_flutter start async operations but do not handle the failure case, leaving the dialog permanently open with a loading spinner:

`PkamDialog` — calls `auth.authenticate()` with `.then()` only. If `authenticate()` throws (e.g. `SocketException`, "Max retries reached"), the exception propagates unhandled and `Navigator.of(context).pop()` is never called. The dialog stays open indefinitely.

```
// pkam_dialog.dart — current behaviour
auth.authenticate(request, backupKeys: backupKeys).then((response) {
  Navigator.of(context).pop(response); // ← only reached on success
});
// No .catchError() / try-catch — exception escapes silently
```

- `CramDialog` — identical pattern: _authService.onboard(...).then(...) with no error handler.

- `ApkamActivationDialog` — the submit button onPressed calls await _sendEnrollment(otp) with no try/catch. An exception leaves the dialog stuck in its _isLoading = true spinner state.

Because `showDialog()` only resolves when the dialog is explicitly popped, the caller's `await PkamDialog.show(...)` never completes, and the calling screen cannot reset its loading state or show an error message.

### Steps to reproduce

1. Add a valid atSign and its keys to the device keychain.
2. Disable all network connectivity (airplane mode or Wi-Fi/cellular off).
3. Open an app that uses `PkamDialog.show(...)`.
4. Select the atSign and proceed — the PKAM dialog appears with a spinner.
5. Wait. The dialog never dismisses and no error is shown.
6. The unhandled exception appears in the debug console.
8. The same steps apply to CramDialog and ApkamActivationDialog under network failure conditions.

### Expected behavior

Each dialog should handle the failure case explicitly:

Wrap the async operation in try/catch.
On failure, call Navigator.of(context).pop() and show a `Snackbar` with an error message

### Smartphones

- Device: iPhone17 Pro
- OS: iOS26.2

### Were you using an atApplication when the bug was found?

at_client_flutter

Xlin123:
if the registrarService is missing the apikey, it doesn't throw a helpful exception

PR #2070, opened 2026-07-14
fix(at_client_flutter): onboarding dialogs for the bounded-timeout world

### LINKED ISSUE #1905: Unhandled Exception when using CramDialog
### Describe the bug

When using the cram dialog I go the error: `[ERROR:flutter/runtime/dart_vm_initializer.cc(40)] Unhandled Exception: Exception: unable to connect to atServer for @lunar4bc01_np on 46ab7523-d4ec-5c4c-aa52-f72bb274ae7d.swarm0003.atsign.zone:22918` show in debug console but the app was stuck on the screen in the image.

Image

### Steps to reproduce

1. First I called `final onboardingResponse = await CramDialog.show(...);`
2. Then I got the dialog in the image with the code shown above in debug console

Reproducing this bug might be difficult since it requires connecting to the atServer to fail.

### Expected behavior

When connection to the atServer failed I expected so see a non technical error message for the user.

### Screenshots

_No response_

### Smartphones

- Device: Iphone17Pro
- OS: IOS26.2

at_client_flutter

### Additional context

_No response_

gkc:
## What I did

Hardened the onboarding/authentication dialogs in `at_client_flutter` so they behave correctly now that `onboard`/`authenticate` are **time-bounded** by the #1923 network-timeout work (merged as #2072/#2073/#2074).

The original hang fix (#1905/#1909) only helped when the underlying call *completed*. Before the timeout work, a dead network left `onboard`/`authenticate` in a ~50-minute retry stack, so the future never resolved and the dialog hung regardless. Now those calls throw `AtTimeoutException` (auth 30s, onboarding 5 min), the dialogs' error paths fire, and `.show()` always completes. This PR both lands that error handling and tunes the dialogs for the new bounded-wait behaviour:

- **Dialogs always complete (#1905/#1909).** `CramDialog` / `PkamDialog` catch onboarding/auth errors (incl. `AtTimeoutException`), surface a SnackBar, and pop with `null`; `ApkamActivationDialog` catches enrollment errors and keeps the dialog open for retry.
- **Timeout-aware messaging.** An onboarding `AtTimeoutException` is reported as "your atSign may still be provisioning — try again," not a hard failure; PKAM/APKAM distinguish a reachability timeout from a genuine failure.
- **Live progress by default.** Both dialogs already subscribed to `progressStream` but dropped the events unless a custom `progressBuilder` was supplied — a multi-minute provisioning wait showed static text. They now feed the latest `ProgressEvent.msg` into the default loading view.
- **Start once, not per `build`.** `CramDialog` / `PkamDialog` launched the async call from `build()`, so a rebuild during the (now up-to-5-min) wait could spawn a second `onboard`/`authenticate`. Both are now `StatefulWidget`s that start the call once in `initState`.
- **Correctness nit.** Removed a `return`-in-`finally` in `ApkamActivationDialog` (flagged by the analyzer) that could swallow an in-flight exception; the `setState` is now guarded by `if (mounted)`.

## How I did it

- Converted `CramDialog` and `PkamDialog` from `StatelessWidget` to `StatefulWidget`; the onboarding/auth future is started in `initState` and its `try`/`catch` pops the dialog (success → response, failure → `null` + SnackBar), guarded by `mounted`.
- Branched the SnackBar text on `e is AtTimeoutException` in all three dialogs (imported `AtTimeoutException` from `at_commons` in cram/pkam; it's already in scope via the barrel in apkam).
- In the default `StreamBuilder` branch (no custom `progressBuilder`), the `LoadingDialog` description now shows `snapshot.data!.msg` once progress events arrive, falling back to the provided/default description beforehand.
- Replaced `ApkamActivationDialog`'s `finally { if (!mounted) return; setState(...) }` with `finally { if (mounted) setState(...) }`.

## How to verify it

    cd packages/at_client_flutter
    flutter analyze --no-fatal-infos          # clean (1 pre-existing info in an untouched file)
    dart format . -o none --set-exit-if-changed
    flutter test --concurrency=1              # 14 pass

Manual: show `CramDialog` / `PkamDialog` against an unreachable atServer → the dialog closes with an error SnackBar within the bounded timeout instead of hanging; against a live server, the loading view shows live progress messages during the wait.

- fix: onboarding/auth dialogs always complete (surface a message + pop) when `onboard`/`authenticate` throw or time out; onboarding timeouts read as "still provisioning," default loading view shows live `progressStream` progress, and the async call is started once in `initState` rather than from `build`.

## Also in this PR

Corrects a misattributed reference: the five merged network-timeout CHANGELOG entries (at_commons, at_lookup, at_auth, at_client, at_client_flutter) cited **#1923** — which is a separate draft PR, not the tracking issue. They now point at **#1909**, the "dialogs hang forever" issue the timeout work root-cause-fixes.

---

Closes #1905, #1909. Builds on the merged timeout work (#2072/#2073/#2074, tracked by #1909). Folds under the in-progress `at_client_flutter` **1.1.4** (the version #2074 already published to trunk) — no bump.

Note: draft PR #1923 overlaps this one on the three dialogs (#2070 supersedes that part) but also carries registrar API-key error handling + a dialog test that are out of scope here — see the coordination comment there.

### DIFF
--- a/packages/at_auth/CHANGELOG.md
+++ b/packages/at_auth/CHANGELOG.md
@@ -10,7 +10,10 @@
   spent, then throws `AtTimeoutException`; each inner network call (the atDirectory
   lookup and the connectivity probe) is bounded by the remaining budget and capped
   at 60s. **`RetryOptions.maxRetries` no longer bounds this loop** (the deadline
-  does) (#1909). Requires `at_commons ^5.13.0`.
+  does) (#1923). Requires `at_commons ^5.13.0`.
+- chore(deps): `at_lookup: ^3.6.0` — `validateAtServer` passes the `timeout`
+  parameter that `SecondaryAddressFinder.findSecondary` gained in at_lookup
+  3.6.0, so this version does not compile against at_lookup ≤3.5.x.
 
 ## 3.1.1
 - refactor: route enrollment RSA (encrypt/decrypt `apkamSymmetricKey` under the default encryption keypair) through at_chops (`RsaEncryptionAlgo`) — `crypton` no longer imported in `lib` and moved to `dev_dependencies` (only the enrollment test still uses it for RSA keypair fixtures). Same framing, byte-identical by construction.
--- a/packages/at_auth/CHANGELOG.md
+++ b/packages/at_auth/CHANGELOG.md
@@ -10,7 +10,7 @@
   spent, then throws `AtTimeoutException`; each inner network call (the atDirectory
   lookup and the connectivity probe) is bounded by the remaining budget and capped
   at 60s. **`RetryOptions.maxRetries` no longer bounds this loop** (the deadline
-  does) (#1923). Requires `at_commons ^5.13.0`.
+  does) (#1909). Requires `at_commons ^5.13.0`.
 
 ## 3.1.1
 - refactor: route enrollment RSA (encrypt/decrypt `apkamSymmetricKey` under the default encryption keypair) through at_chops (`RsaEncryptionAlgo`) — `crypton` no longer imported in `lib` and moved to `dev_dependencies` (only the enrollment test still uses it for RSA keypair fixtures). Same framing, byte-identical by construction.
--- a/packages/at_auth/pubspec.yaml
+++ b/packages/at_auth/pubspec.yaml
@@ -12,7 +12,7 @@ resolution: workspace
 dependencies:
   at_server_status: ^1.1.0
   at_commons: ^5.13.0
-  at_lookup: ^3.4.2
+  at_lookup: ^3.6.0
   at_chops: ^3.0.0
   at_utils: ^3.0.19
   http: ^1.2.1
--- a/packages/at_client/CHANGELOG.md
+++ b/packages/at_client/CHANGELOG.md
@@ -4,7 +4,10 @@
   (`AtNetworkTimeouts.defaultTimeout`, capped at 60s), bounding every atServer
   connect / atDirectory lookup / operation so a dead network can't hang the SDK.
   Supersedes the misnamed `outboundConnectionTimeout` (a socket idle time).
-  Requires `at_commons ^5.13.0` (#1909).
+  Requires `at_commons ^5.13.0` (#1923).
+- chore(deps): `at_lookup: ^3.6.0`, `at_auth: ^3.2.0` — the bounded socket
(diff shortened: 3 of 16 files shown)


gkc:
Might be the wrong approach but wanted to do something to get the ball rolling

Xlin123 (approved):
@gkc thank you very much for fixing those, I think I had a branch sitting somewhere for this and essentially had written the same code.

LGTM 🚀

Xlin123 (changes requested):
oh quick note actually from #1923 

> Found when testing using repro instructions in ticket (airplane mode and pkamming).
>
> 
> The root cause of the hanging is actually from stacking retries in our libraries (ie: AtLookup is retrying and so is AtAuth)
> I tried adding UI level timeouts but it still leaves the underlying work retrying.


did you run into the same issues?

gkc:
> oh quick note actually from #1923 
> 
> > Found when testing using repro instructions in ticket (airplane mode and pkamming).
> >
> > 
> > The root cause of the hanging is actually from stacking retries in our libraries (ie: AtLookup is retrying and so is AtAuth)
> > I tried adding UI level timeouts but it still leaves the underlying work retrying.
> 
> 
> did you run into the same issues?

I'll look deeper tomorrow

gkc:
> > oh quick note actually from #1923
> > > Found when testing using repro instructions in ticket (airplane mode and pkamming).
> > > The root cause of the hanging is actually from stacking retries in our libraries (ie: AtLookup is retrying and so is AtAuth)
> > > I tried adding UI level timeouts but it still leaves the underlying work retrying.
> > 
> > 
> > did you run into the same issues?
> 
> I'll look deeper tomorrow

Yeah you're right. Sticking plaster won't work here. I'm going to have to go much deeper. Thanks @Xlin123

gkc:
@Xlin123 Thanks again for the heads-up - I'm going to do a three-PR (four including this one) fix: PR 1 for at_lookup (and a little at_commons); PR 2 for at_auth; PR 3 for at_client and at_client_flutter stuff; I'll leave this one for last. I've tagged you for review for [the first PR ](https://github.com/atsign-foundation/at_client_sdk/pull/2072)

gkc:
@Xlin123 ready for re-review. This supersedes **some** of your older PR #1923, but by no means all - I've left a more detailed comment for you there

Xlin123 (dismissed):

Xlin123 (approved):

PR #2072, opened 2026-07-15
feat(at_lookup): bound atDirectory lookup + connect with a timeout deadline (network-timeout stage 1)

gkc:
## What I did

**PR 1 (of 3) of the network-timeout fix** for the "SDK hangs for minutes on a dead network" problem (#1923, surfaced in the #2070 review). This stage bounds the *lowest* network layer; follow-up stages thread the deadline through at_auth's retry loop and expose an `AtClientPreference` knob.

The hang is a nested retry stack with no wall-clock budget: `AtAuthImpl.validateAtServer` (10× / 2s) wraps `SecondaryUrlFinder.findSecondaryUrl` (5×), whose `_findSecondary` busy-waits **30s** for the atDirectory, and the actual `SecureSocket.connect` had **no `timeout`** — worst case ~50 minutes.

PR 2 will be the at_auth deadline loop over its 10× retry, then PR 3 will be to add `AtClientPreference.networkTimeout` + do the at_client/at_client_flutter wiring

## How I did it

**at_commons (5.13.0):**
- `AtNetworkTimeouts` — the process-wide network-timeout policy: `defaultTimeout` = 30s, `maxAllowed` = 60s hard cap, `cap()`. The single place to set the default; every resolved timeout is capped at 60s (a caller may explicitly request up to 60s; the default when none is given is 30s).
- `SecureSocketConfig.connectTimeout`.

**at_lookup (3.6.0):**
- `SecureSocketUtil.createSecureSocket` now passes a `timeout` to `SecureSocket.connect` (precedence: explicit > `config.connectTimeout` > process default, all capped).
- `SecondaryAddressFinder.findSecondary` takes an optional `timeout`, converted to a single absolute deadline threaded through the 5-attempt retry loop and the 30s response busy-wait — both abandon once the budget is spent.
- Defaults to `AtNetworkTimeouts.effectiveDefault` (30s) even when no caller passes a timeout, so existing call sites are bounded now (capped at 60s).

Note: at_commons 5.12.0 is published, so this bumps at_commons to **5.13.0** (additive minor); at_lookup 3.6.0 is in progress, so am adding the changelog entry under there.

## How to verify it
Tests pass

Note New test in `secondary_address_cache_test.dart` — "findSecondary honours its timeout budget when the atDirectory connects but never answers" — proves the finder gives up in <10s instead of the old 30s busy-wait.

### DIFF
--- a/packages/at_auth/test/at_auth_test.dart
+++ b/packages/at_auth/test/at_auth_test.dart
@@ -34,7 +34,8 @@ class FakeEnrollmentRequest extends Fake implements EnrollmentRequest {}
 class FakeSecondaryAddressFinder extends Fake
     implements CacheableSecondaryAddressFinder {
   @override
-  Future<SecondaryAddress> findSecondary(String atSign) async {
+  Future<SecondaryAddress> findSecondary(String atSign,
+      {Duration? timeout}) async {
     return SecondaryAddress('abcd', 123);
   }
 }
--- a/packages/at_commons/CHANGELOG.md
+++ b/packages/at_commons/CHANGELOG.md
@@ -1,3 +1,12 @@
+## 5.13.0
+
+- feat: add `AtNetworkTimeouts` — the process-wide network-timeout policy
+  (`defaultTimeout`, `maxAllowed` = 30s hard cap, `cap()`). The single place to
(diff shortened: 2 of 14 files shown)


Xlin123 (approved):
lgtm, the new at_commons class looks great

PR #2073, opened 2026-07-15
feat(at_auth): bound validateAtServer with an overall timeout deadline (network-timeout stage 2)

gkc:
## What I did

**Stage 2 of the network-timeout fix** (#1923), building on Stage 1 (#2072, merged). Stage 1 bounded the lowest network layer; this stage caps the wall-clock of `at_auth`'s retry loop, which was the outer multiplier of the "SDK hangs for minutes on a dead network" hang.

`AtAuthImpl.validateAtServer`'s retry loop had no wall-clock budget, so it multiplied the (now individually-bounded) atDirectory-lookup waits into a minutes-long hang.

A key nuance surfaced in review: `validateAtServer` serves **two paths that need opposite budgets** — **authentication** of an *existing* atSign must fail fast on a dead network, while **onboarding** of a *new* atSign must **poll for provisioning**, which can legitimately take minutes.

## How I did it

**at_commons (5.13.0, in-progress on trunk):**
- `AtNetworkTimeouts.defaultOnboardingTimeout` = 5 min — the poll budget for waiting on a newly-registered atSign to be provisioned. Deliberately longer than the 60s per-operation cap (`maxAllowed`), because that cap bounds *individual network operations*, not a poll loop.

**at_auth (3.2.0):**
- `RetryOptions.overallTimeout` — optional; when null the default depends on the request type: **authentication** uses `AtNetworkTimeouts.effectiveDefault` (30s) so a dead network fails fast; **onboarding** uses `defaultOnboardingTimeout` (5 min) so a new atSign has time to provision.
- `validateAtServer` computes one absolute deadline up front and is now **deadline-driven** — it retries every `retryDelay` until the budget is spent, then throws `AtTimeoutException`. Each inner call (`atStatus.get`, `findSecondary` (Stage 1's param), the connectivity probe) is bounded by the remaining budget and capped at 60s. **`RetryOptions.maxRetries` no longer bounds this loop** (the deadline does).

## How to verify it

    cd packages/at_auth && dart analyze lib test && dart test --concurrency=1 test/at_auth_test.dart
    cd ../at_commons && dart analyze lib test && dart test --concurrency=1 test/at_network_timeouts_test.dart

[...]

PR #2074, opened 2026-07-15
feat(at_client): AtClientPreference.networkTimeout + flutter auth timeout (network-timeout stage 3)

gkc:
## What I did

**Stage 3 (final) of the network-timeout fix** (#1923), on top of Stage 1 (#2072, merged) and Stage 2 (#2073, merged). The foundation (per-op timeouts) and the at_auth deadline are in; this stage exposes the app-facing knobs, fixes a connection leak, and documents the onboarding-vs-auth timeout distinction for app authors.

## How I did it

**at_client (3.13.0):**
- `AtClientPreference.networkTimeout` — when set on the preference used to create an `AtClient`, it becomes the process-wide `AtNetworkTimeouts.defaultTimeout` (capped 60s), bounding every atServer connect / atDirectory lookup / operation. Supersedes the misnamed `outboundConnectionTimeout` (a socket idle time, not an operation timeout).

**at_client_flutter (1.1.4):**
- `AuthService.onboard` / `authenticate` accept an optional `timeout` that bounds the whole onboarding/auth attempt (sets `RetryOptions.overallTimeout`; otherwise the process-wide `AtNetworkTimeouts` policy applies — auth 30s, onboarding 5 min).
- `FlutterEnrollmentService.enroll` now closes the `AtLookupImpl` connection in a `finally`, fixing a leak when the enrollment submit throws (a timeout or a network failure).
- **Docs:** an "Onboarding, provisioning & timeouts" section in the README, and expanded `onboard`/`authenticate` dartdocs, so app authors know to leave `timeout` null for the 5-minute provisioning poll (and that a short value truncates it). See [Additional context](#additional-context) below.

**Follow-up filed:** the provisioning poll is not cancellable in-flight — `onboard()` has no cancellation token, so its `Future` runs to completion/timeout even if the user navigates away. Tracked as a feature request in #2075 (referenced from the README note).

## How to verify it

    cd packages/at_client
    dart analyze lib test
    dart test --concurrency=1 test/at_client_impl_test.dart
    cd ../at_client_flutter
    dart analyze lib test
    flutter test --concurrency=1

All 20 at_client construction tests and 14 at_client_flutter tests (incl. `auth_service_test`) pass; both analyze clean.

- feat(at_client): `AtClientPreference.networkTimeout` — a single knob to set the process-wide network timeout.
- feat(at_client_flutter): optional `timeout` on `AuthService.onboard`/`authenticate`; fix a connection leak in `FlutterEnrollmentService.enroll`.

---

**Stage 3 of 3 (final).** Builds on #2072 + #2073 (merged) — completes the #1923 network-timeout fix. Refs #1905, #1909. Opens follow-up #2075 (cancellation).

Note: at_client folds under the in-progress 3.13.0; at_client_flutter → **1.1.4** (additive minor). ⚠️ **#2070 (the dialog fix) also targets 1.1.4 and is still open** — whichever of these two merges *second* will need to bump to 1.1.5.

## Additional context

This stage documents a distinction that surfaced in review and isn't obvious from the API alone: **registering a new atSign and having its atServer *provisioned* are two separate steps**, and provisioning can lag registration by anything from seconds to a few minutes. So the two auth entry points get deliberately opposite budgets:

- **`onboard`** (new atSign) polls for the atServer to come up for **5 minutes** by default (`AtNetworkTimeouts.defaultOnboardingTimeout`), retrying every `retryDelay` with each individual network probe capped at 60s. The 5 minutes is a *retry* budget, not a single wait.
- **`authenticate`** (existing atSign) **fails fast at 30s** — it's already provisioned, so a dead network should surface quickly.

Setting `AtClientPreference.networkTimeout` does **not** shrink the provisioning wait: it writes `AtNetworkTimeouts.defaultTimeout` (the per-op / auth default), never `defaultOnboardingTimeout`. The one crossover is `AuthService.onboard(timeout:)` — an explicit value overrides the 5-minute default, so a short value there truncates the provisioning poll (the footgun the new dartdoc calls out).

Guidance added to the at_client_flutter README for app authors — call `onboard()` with no `timeout`, drive the UI from `AuthService.progressStream`, and use an app-level Retry on `AtTimeoutException` rather than a giant single timeout:

```dart
final sub = authService.progressStream.listen((e) {
  setState(() => _status = e.msg); // e.type: info / success / warning / error
});
try {
  await authService.onboard(request, cramSecret); // no timeout → full 5-min poll
} on AtTimeoutException {
  // provisioning still not ready after 5 minutes — offer Retry
} finally {
  await sub.cancel();
}
```

Because `onboard()` can't be cancelled mid-poll today, the README flags it and points at #2075 for true cancellation.

### DIFF
--- a/packages/at_client/CHANGELOG.md
+++ b/packages/at_client/CHANGELOG.md
@@ -1,4 +1,10 @@
 ## 3.13.0
+- feat: add `AtClientPreference.networkTimeout` — when set on the preference used
+  to create an `AtClient`, it becomes the process-wide network-timeout default
+  (`AtNetworkTimeouts.defaultTimeout`, capped at 60s), bounding every atServer
+  connect / atDirectory lookup / operation so a dead network can't hang the SDK.
+  Supersedes the misnamed `outboundConnectionTimeout` (a socket idle time).
+  Requires `at_commons ^5.13.0` (#1923).
 - refactor: migrate the local keystore to `at_persistence_secondary_server`
   5.0.0 — the client is now commit-log-free. The client no longer maintains a
   local commit log or runs commit-log compaction; sync tracks its progress
--- a/packages/at_client/lib/src/client/at_client_impl.dart
+++ b/packages/at_client/lib/src/client/at_client_impl.dart
@@ -351,6 +351,14 @@ class AtClientImpl implements AtClient {
     _logger = AtSignLogger('AtClientImpl ($_atSign)');
     _preference = preference;
     _preference?.namespace ??= namespace;
+    // If the app configured a process-wide network timeout, apply it as the
+    // single default that bounds every atServer connect / atDirectory lookup /
+    // operation. Only mutates the global when explicitly set, so the built-in
+    // default otherwise stands.
+    if (preference.networkTimeout != null) {
+      AtNetworkTimeouts.defaultTimeout =
+          AtNetworkTimeouts.cap(preference.networkTimeout!);
+    }
     _localSecondaryKeyStore = localSecondaryKeyStore;
 
     if (_localSecondaryKeyStore != null && !_preference!.isLocalStoreRequired) {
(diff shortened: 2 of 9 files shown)


gkc:
@Xlin123 note the "additional context" section above

Xlin123 (approved):

PR #2076, opened 2026-07-15
feat(at_client): thread AtKeysIo through CryptoContext

Xlin123:
## Summary

Interface-first step of the PQ/crypto-agility epic (#1889). Per the ratified
#2045 decision, `CryptoProvider`s should source key material from at_auth's
`AtKeysIo` rather than off the client. This PR adds the seam — the field, its
sourcing at `AtClient` construction, and the wiring through `CryptoRuntime` —
without migrating any provider onto it yet.

Additive only. Branches off `trunk`, parallel to #2047 (no file overlap).

## What changed

- **`CryptoContext.io`** (`AtKeysIo?`) — new field a provider reads to get the
  client's `AtKeys` via `io.read(atSign)`. `null` until an app injects one; no
  default source.
- **`AtClient.atKeysIo`** — new getter (spec + impl). Injected via
  `AtClientImpl.create(atKeysIo:)`, stored on first construction and immutable
  after — same policy as `atChops` (a cached client keeps its original key
  source).
- **`CryptoRuntime`** — reads `_atClient.atKeysIo` live per operation into
  `CryptoContext`, matching the existing live `preference.crypto` resolution.
- **Tests** — `crypto_runtime_test`: provider sees `identical(context.io,
  injected)` on encrypt + decrypt, and `context.io` is null when unstubbed.
  `at_client_impl_test`: `create(atKeysIo:)` stores it, plain `create()` leaves
  it null, cached re-use retains the original. Added a `StubAtKeysIo` (throws on
  read/write, proving no accidental key IO) and an example
  `KeysSeamCryptoProvider` showing the intended consumption shape.

## Explicitly deferred

- `LegacyCryptoProvider` still reads keys off the client; it does **not**
  consume `context.io` (decision #E).
- No provider lifecycle/init hook — providers stay stateless; revisit when the
  first stateful provider lands.

## Semver: minor

Optional params on an unpublished class/method; the whole crypto seam is
unpublished (in-tree 3.13.0). The new abstract `AtClient.atKeysIo` getter is
technically source-breaking for hand-written implementers.

## Skills impact

PQ plan conformance: project S-2 — [implementation-plan.md](https://github.com/atsign-foundation/at_client_sdk/blob/trunk/docs/projects/pq/implementation-plan.md), tracking epic https://github.com/atsign-foundation/at_client_sdk/issues/1889.


packages/at_client/lib/src/client/at_client_impl.dart
```diff
@@ -344,6 +355,7 @@ class AtClientImpl implements AtClient {
     EncryptionService? encryptionService,
     AtKeyValueStore<String, AtData, AtMetaData?>? localSecondaryKeyStore,
     AtChops? atChops,
+    AtKeysIo? atKeysIo,
```
Xlin123, line 359: if it helps, AtKeysIo being handed to an AtClient with a successful auth response is the equivalent of the old AtChops being handed to an AtClient.

Xlin123:
> No provider lifecycle/init hook — providers stay stateless; revisit when the
> first stateful provider lands.

to add to this point, I am picturing one of two shapes:
**1. (in favour)** stateless, but during encrypt/decrypt time we use the context to "lazily" ensure we have the right keys in place.
2. stateful with some sort of init method for construction time of the CryptoProviders

packages/at_client/lib/src/client/at_client_impl.dart
gkc: got it. should we deprecate the atChops param on the previous line?

packages/at_client/lib/src/crypto/crypto.dart
```diff
@@ -47,14 +48,18 @@ class CryptoContext {
...
 
-  const CryptoContext({required this.atClient});
+  /// The client's key source (ratified atsign-foundation/at_client_sdk#2045):
+  /// an `AtKeysIo` (`package:at_auth`) whose `read(atSign)` yields the
+  /// client's `AtKeys`. Sourced from [AtClient.atKeysIo], injected at client
+  /// construction (`AtClientImpl.create(atKeysIo:)`). Null until an app
+  /// injects one — store wiring (so this is populated by default) lands in a
+  /// later project. The built-in legacy provider deliberately does not read
+  /// this yet.
+  final AtKeysIo? io;
```
gkc: Please call this `atKeysIo` (sorry)

packages/at_client/lib/src/client/at_client_impl.dart
gkc: oh I guess the answer is no, because we're not actually using this anywhere yet. Maybe worth some dartdoc comments to explain where we're headed.

packages/at_client/test/test_utils/mocks.dart
```diff
@@ -38,4 +41,90 @@ class MockAtClient extends Mock implements AtClient {
...
+/// client), this one sources its key from the injected [AtKeysIo]: it reads the
+/// current atSign's [AtKeys] and uses `defaultSelfEncryptionKey` as a symmetric
+/// key. This is the intended shape for providers once store wiring lands — the
+/// provider never touches the client's atChops, only `context.io`.
+///
+/// The crypto here is a toy (repeating-key XOR) purely to keep the seam usage
+/// legible; a real provider would feed the key material into AES/GCM or a
+/// ratchet. Pair it with a functional `AtKeysIo` (one whose `read` returns a
+/// canned [AtKeys], not the throwing [StubAtKeysIo]) to exercise a round trip.
+class KeysSeamCryptoProvider implements CryptoProvider {
```
gkc: There are no tests using this? Why's this here?

gkc (changes requested):
Mostly LGTM, some maybe-minor-ish nits

packages/at_client/test/test_utils/mocks.dart
Xlin123: removed in [a4e1b1c](https://github.com/atsign-foundation/at_client_sdk/pull/2076/commits/a4e1b1cc6981e7673b58002e57feda5c14f74f26)

packages/at_client/lib/src/crypto/crypto.dart
Xlin123: [a4e1b1c](https://github.com/atsign-foundation/at_client_sdk/pull/2076/commits/a4e1b1cc6981e7673b58002e57feda5c14f74f26)

packages/at_client/lib/src/client/at_client_impl.dart
```diff
@@ -300,7 +306,9 @@ class AtClientImpl implements AtClient {
     RemoteSecondary? remoteSecondary,
     EncryptionService? encryptionService,
     AtKeyValueStore<String, AtData, AtMetaData?>? localSecondaryKeyStore,
+    @Deprecated('replaced by atKeysIo, will be removed in the next release')
```
Xlin123, line 309: noted for deprecation in the AtClientImpl.create factory

gkc (approved):

PR #2077, opened 2026-07-16
feat: configurable cache duration for CacheableSecondaryAddressFinder

srieteja:
## What
- Make cache duration configurable for CacheableSecondaryAddressFinder.

## How
- Add an optional param `cacheDuration` to `CacheableSecondaryAddressFinder`'s constructor.

## How to verify

1. Tests pass
2. Cache should respect the configured duration.

## Skills impact

### DIFF
--- a/packages/at_lookup/CHANGELOG.md
+++ b/packages/at_lookup/CHANGELOG.md
@@ -1,15 +1,7 @@
 ## 3.7.0
 
-- feat: `CacheableSecondaryAddressFinder` accepts an optional `cacheDuration`
-  constructor parameter, so callers that need faster recovery from a stale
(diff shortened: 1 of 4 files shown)


gkc:
@srieteja We should not reduce the default in at_client, that is fine. But in the proxy service you can, before creating the CacheableSecondaryAddressFinder, change its static default timeout.

gkc:
The minimum thing we need to do here is change `defaultCacheDuration` from `static const Duration` to `static Duration`

The ideal is to allow passing of cacheDuration at construction time, defaulting to defaultCacheDuration

gkc (approved):
LGTM

Issue #2055, opened 2026-07-06
Document atSign activation / atDirectory prerequisites


CurtlyCritchlow:
### Is your feature request related to a problem? Please describe.

Auth fails with `No entry in atDirectory for <atsign>` on atSigns that aren't activated/registered, and the skill doesn't explain the prerequisite or the error.

### Describe the solution you'd like

State that an atSign must be activated and registered in the atDirectory before use, explain what `No entry in atDirectory` means, and show how to confirm an atSign is usable.

hiSandog:
The documentation could distinguish an unactivated atSign from transient atDirectory lookup failures, since both may otherwise look like authentication problems. A short preflight example that performs the directory lookup before requesting credentials would give users a concrete diagnostic step. It would also help to state whether activation must be completed on the same environment or root server used by the SDK configuration.

CurtlyCritchlow:
Closed in #2078

Issue #2054, opened 2026-07-06
Document the owner-writes-only collaboration/ownership model


CurtlyCritchlow:
### Is your feature request related to a problem? Please describe.

Designing "a collaborator edits a shared item" hits `ArgumentError`/`StateError` because `update`/`delete` require `owner == self`, and the skill never states this.

### Describe the solution you'd like

State that `AtCollection` is owner-writes-only: an atSign can only mutate keys it owns; you cannot edit another atSign's item in place. The collaboration model is additive — each atSign owns what it creates; sharing grants visibility, not write access. Include guidance for designing shared features within this constraint.

CurtlyCritchlow:
closed in #2078

Issue #2053, opened 2026-07-06
Document the Flutter watch() stream-holding rule; fix StreamBuilder snippets


CurtlyCritchlow:
### Is your feature request related to a problem? Please describe.

A collaborator's change is received and fires `CItemUpdated`, but the `StreamBuilder` UI does not refresh until a hot reload/restart — because `collection.query().watch()` is created inside `build()`, which mints a new stream object every rebuild and tears down the live subscription.

### Describe the solution you'd like

State the rule plainly: create a `watch()` stream once, hold it in `State` (recreate only when its inputs change), and memorise the `Query` across rebuilds; never call `watch()` inside `build()`. Rewrite the skill's `StreamBuilder` snippets (§6 / ref 04) to follow this.

CurtlyCritchlow:
Closed in #2078

Issue #2019, opened 2026-06-22
Add a robust end-to-end test app to validate the at_client agent skill


CurtlyCritchlow:
### Is your feature request related to a problem? Please describe.

We need a deliberately comprehensive app that forces the skill to touch every major SDK area, so building it cleanly start-to-finish becomes a real test of the skill.

### Describe the solution you'd like

A documented "build this app" prompt for a cross-platform (iOS, Android, macOS) collaborative shared task-board app — Boards → Lists → Cards → checklist items — that exercises:

- all auth flows (new account, .atKeys file import, keychain re-login, sign-out) — incl. macOS file picker
- nested hierarchy (sub-collections, deep watch)
- sharing a board with another user + collaborative edits (notifications)
- real-time updates on shared data (event streams)
- "seen by" indicators (read receipts)
- filtering/search by priority, tag, due-date, done state (typed queries)
- offline + reconcile
- unit tests with no live atServer

Success = the skill can build it end-to-end with no errors. Any failure is a skill gap to fix.

CurtlyCritchlow:
closed in PR #2024

PR #2078, opened 2026-07-16
feat(at_client_skills): v1.2.0 — sync, watch()-holding, ownership & atDirectory guidance

### LINKED ISSUE #2052: Document sync scoping (syncRegex) and the sync lifecycle

With default preferences, `SyncService` never completes (`inProgress`, `keysSent=0`, indefinitely) — unscoped sync covers the atSign's entire keystore and stalls on unrelated/dangling keys from other apps, so shares, edits, and read receipts silently never propagate. The skill also gives no way to reason about when a write has reached the server or why a read is stale.

A "sync" reference covering: (a) set `AtClientPreference.syncRegex = '<namespace>'` — sync is NOT namespace-isolated by default; reserved keys (pkam/encryption/public/`shared_key`/`cached:`) sync regardless, so scoping doesn't break sharing or decryption. (b) The lifecycle: reads are local-only, writes sync in the background, cadence is `syncIntervalMins`, and completion is observed via `SyncService` + `SyncProgressListener` (`sync(onDone:)` is deprecated).

Relying on SDK defaults — but unscoped sync wedges on any atSign used by more than one app, so this isn't viable.

### Additional context

_No response_

CurtlyCritchlow:
## What

Closes #2052, #2053, #2054, #2055

v1.2.0 of the `at_client_skills` agent skill. Closes the four "Agentic
Engineering" documentation gaps found by validating the skill against a real
collaborative app (TeamBoard, #2019) — the runtime guidance a real app depends
on that the skill was missing:

- **#2052** — sync scoping (`syncRegex`) + the sync lifecycle
- **#2053** — the Flutter rule to hold `watch()` streams in `State`, never in `build()`
- **#2054** — the owner-writes-only collaboration/ownership model
- **#2055** — the atSign activation / atDirectory prerequisite

## How

- **#2052:** new `references/11-sync.md` (scope sync to your namespace or it
  wedges; reads local / writes background; `SyncProgressListener`; read-receipt
  eventual-consistency). `syncRegex` added to the `_setupAtClient` example in
  `05-flutter-auth.md` + a SKILL.md callout.
- **#2053:** watch-rule callout in SKILL.md; `StreamBuilder` snippets in
  `03-query-api.md` / `04-events-api.md` / `10-architecture-guide.md` rewritten
  to hold the stream in `State` (verified against `examples/todos` +
  `board_screen` — recreate on input change in `setState`, not `didUpdateWidget`).
- **#2054:** owner-writes-only callout in SKILL.md §4 + note in
  `02-atcollection-api.md` (`update`/`updateSharedWith`/`delete` throw
  `ArgumentError` when `item.owner != self`; sharing is additive).
- **#2055:** activation/atDirectory prerequisite section in `05-flutter-auth.md`
  (explains `SecondaryNotFoundException: No entry in atDirectory`).
- 4 new evals (#15–#18); version bumped to **1.2.0** (pubspec + SKILL.md +
  CHANGELOG). Every claim verified against the `at_client` source.

## How to verify

1. Markdown lint passes: `pymarkdown --config packages/at_client_skills/.pymarkdown scan packages/at_client_skills/**/*.md`.
2. `cd packages/at_client_skills && dart pub publish --dry-run` — validates, `validation/` excluded, 0 warnings from a clean checkout.
3. `evals/evals.json` parses and has 18 evals; the four new ones (#15–#18) cover the gaps above.

## Skills impact


packages/at_client_skills/skills/at_client_skills-sdk/evals/evals.json
```diff
@@ -70,6 +70,26 @@
...
+    },
+    {
+      "id": 17,
+      "prompt": "I want a user to edit an item that another atSign shared with them, but calling update() on it throws ArgumentError. How should collaborative editing work?",
+      "expected_output": "Explains AtCollection is owner-writes-only: update/updateSharedWith/delete throw ArgumentError when item.owner is not the current atSign, so you cannot edit another atSign's item in place. The model is additive: each atSign owns what it creates, and sharing grants a readable copy, not write access. To model contribution, the peer creates their own item and shares it back; reading a received item works but mutating it does not."
+    },
```
CurtlyCritchlow, line 88: @gkc please confirm that the explanation of 17 is true.

packages/at_client_skills/skills/at_client_skills-sdk/references/11-sync.md
```diff
@@ -0,0 +1,101 @@
...
+final sync = AtClientManager.getInstance().syncService;
+final listener = _SyncedListener(() => print('synced'));
+sync.addProgressListener(listener);
+// ... later:
+sync.removeProgressListener(listener);
+```
+
+`SyncProgress` carries `syncStatus` (`SyncStatus`:
+`started` / `notStarted` / `inProgress` / `success` / `failure`),
+`localCommitId`, `serverCommitId`, `lastSyncedOn`, and `atClientException`.
```
gkc: `SyncProgress` has no `lastSyncedOn` field — that one lives on `SyncResult` (`sync_service.dart:118`). An agent following this writes `p.lastSyncedOn` and gets a compile error, which is the exact failure this skill exists to prevent.

Swapping in `pendingPushCount` keeps the list the same length and is the more useful field here: it's why commit-id equality alone isn't sufficient to decide "caught up" when writes land between rounds.

```suggestion
`SyncProgress` carries `syncStatus` (`SyncStatus`:
`started` / `notStarted` / `inProgress` / `success` / `failure`),
`localCommitId`, `serverCommitId`, `pendingPushCount`, and `atClientException`.
```

Optional follow-up, not worth blocking on: this section is aimed at tests and headless drivers, and `SyncServiceWaitUntilCaughtUp.waitUntilCaughtUp` does exactly that job — might be worth a line.

packages/at_client_skills/skills/at_client_skills-sdk/references/11-sync.md
```diff
@@ -0,0 +1,101 @@
...
+final acp = AtClientPreference()
+  ..namespace = 'my_namespace'
+  ..syncRegex = 'my_namespace'   // scope sync to THIS app's keys
+  ..commitLogPath = dir.path
+  ..hiveStoragePath = dir.path;
+```
+
+The atServer always syncs **reserved keys** (pkam / encryption / public /
+`shared_key` / `cached:`) regardless of the regex, so scoping does **not** break
+sharing or decryption — it only excludes other apps' unrelated keys.
```
gkc: The always-include set is narrower than this. Per `alwaysIncludeInSync` in at_server (`packages/at_secondary_server/lib/src/utils/regex_util.dart:27-32`) it's: `shared_key` reserved keys, keys starting with the encryption public key, and namespace-less `public:` keys. `cached:` isn't in it, and `pkam` keys are local-only so never sync regardless.

The conclusion still holds, but for a different reason worth stating — the regex match is unanchored (`RegExp(regex).hasMatch(atKey)`), so `cached:@bob:foo.my_namespace@alice` matches `my_namespace` on its own. Mechanism is what an agent reasons from.

```suggestion
The atServer always syncs a small set regardless of the regex: `shared_key`
reserved keys, the encryption public key, and namespace-less `public:` keys.
The regex match is also an unanchored substring test, so your own namespace's
`cached:` copies match it anyway. Scoping therefore does **not** break sharing
or decryption — it only excludes other apps' unrelated keys.
```

packages/at_client_skills/skills/at_client_skills-sdk/references/11-sync.md
```diff
@@ -0,0 +1,101 @@
...
+receipt has synced back to the item's owner. If receipts never appear at all,
+check sync scoping (§1) first — a wedged sync stops receipts like everything
+else.
+
+---
+
+## Canonical example
+
+<!-- pyml disable-num-lines 2 md013-->
+- [packages/at_client_flutter/examples/todos/](../../packages/at_client_flutter/examples/todos/README.md) — an `AtCollection<T>` app driving sync, shares, and read receipts through the widget stack.
```
gkc: Nit, entirely optional: this relative link resolves to `packages/at_client_skills/skills/packages/at_client_flutter/...`, which doesn't exist — from `references/` the depth needs to be `../../../../`.

This is a pre-existing convention bug rather than something you introduced (`05-flutter-auth.md:259` has the identical pattern), so fine to leave and sweep both separately. `pymarkdown` doesn't check relative-link targets, which is presumably how it survived.

```suggestion
- [packages/at_client_flutter/examples/todos/](../../../../at_client_flutter/examples/todos/README.md) — an `AtCollection<T>` app driving sync, shares, and read receipts through the widget stack.
```

gkc (commented):
Two factual corrections below as suggestions, plus one optional nit. Both matter more than usual here because this file's whole job is being right for agents that can't check it.

Happy to approve once the two are in.

gkc (dismissed):

gkc (approved):

PR #2080, opened 2026-07-17
fix(at_auth): atomic flush + open String key-type tokens; pq docs sync

gkc:
**- What I did**

Review tweaks for #2047, from the S-1-vs-PQ-plan conformance review:

- `FileAtKeysIo` writes are now atomic (write-to-temp + rename) and `flush` keeps the previous file as `<file>.bak` — a crash can no longer truncate the keyfile
- `KeyAlgorithmType` / `CryptographicKeyType` are open String tokens (static-const containers) instead of enums, so unknown tokens round-trip and a new algorithm no longer forces lockstep reader upgrades. Adds the missing `xwing` token; drops `des` and the classical/post-quantum/hybrid role prefixes (the algorithm token carries that fact)
- scoped `flush`'s never-lose dartdoc to bootstrap key stores; reworded the default impl's `UnimplementedError`
- `AtKeys.==`/`hashCode` compare `metadata` structurally (nested maps/lists by value)
- merged trunk; at_auth moves to **3.3.0** (3.2.0 was consumed by the network-timeout release)
- `docs/projects/pq/` reconciled with S-1 as built: 2026-07-17 rulings in decisions.md (flush supersedes append/save; retire-never-remove; never-lose scoping + LocalKeystoreAtKeysIo deferral; String tokens; the 3.3.0 slot), P-2 marked SATISFIED, publish ladder renumbered

**- How I did it**

See commits & diffs

**- How to verify it**

Tests pass (at_auth suite, incl. new .bak and unknown-token round-trip tests)

fix(at_auth): atomic keyfile writes with .bak on flush; open String key-type tokens (adds xwing)


packages/at_client_flutter/lib/src/services/enrollment_service.dart
```diff
@@ -58,8 +58,11 @@ class FlutterEnrollmentService {
       atEnrollmentResponse = await _atEnrollment.submit(request, atLookup);
     } catch (e, s) {
       throw Exception('Enrollment failed: $e \n $s');
+    } finally {
+      // Always close the connection, including when submit() throws (a timeout
+      // or a network failure) — otherwise the AtLookupImpl connection leaks.
+      await atLookup.close();
```
Xlin123, line 64: nice catch

Xlin123 (approved):

PR #2037, opened 2026-06-30
feat(at_client): per APKAM secret sharing substrate

JeremyTubongbanua:
closes https://github.com/atsign-foundation/at_client_sdk/issues/2008

  ## What

  Adds the client-side per-APKAM secret-sharing substrate for WP-SS — the
  mechanism that lets multiple atClients of the *same* atSign (different APKAM
  enrollments, or several keypairs of one enrollment) share secrets such that only
  the intended **APKAM keypair** can read them. This is the layer that will convey
  `nskey` privates per-APKAM (D1) and later underpin the `at/pqmls` group provider
  (D2).

  Every secret travels in an X-Wing-sealed (`pqSeal`), APKAM-signed
  `SecretEnvelope`, addressed by `kpid` and scoped by application namespace, put to
  `<msgId>.<kpid>.__ssenv.<ns>@owner`. Key packages are enrollment-internal —
  conveyed via `EnrollParams.metadata` on `enroll:request` and discovered via the
  gated `enroll:listns` verb — and are **never published**.

  New in `packages/at_client/lib/src/secret_sharing/`:

  - `algo_ids.dart` — algorithm-id constants (`x-wing`, `x-wing-hpke-v1`, `x-wing-v1`).
  - `key_package.dart` — `KeyPackage` / `PackageKey`: the per-APKAM X-Wing recipient
    bundle; `kpid` is the SHA-256 prefix of the X-Wing public half.
  - `secret_store.dart` — `Secret`, `SecretStore`, `SecretStorePersistence`.
  - `secret_envelope.dart` — `SecretEnvelope`: the signed, sealed `__ssenv` delivery unit.
  - `enrollment_directory.dart` — `EnrollmentDirectory` seam + `VerbEnrollmentDirectory`
    backed by the gated `enroll:listns:<ns>` verb (built as a raw wire command, so
    it does not require an at_commons grammar change).
  - `key_package_registration.dart` — `KeyPackageRegistration` mixin: generates/loads
    the X-Wing keypair and registers the package in the enrollment record.
  - `pairwise_secret_sharing.dart` — `PairwiseSecretSharing` mixin: full
    send/receive/sweep/pull flow (`pushSecretToNamespaceMembers`,
    `shareAllSecretsWithEnrollment`, `requestSecret`/`waitForSecret`,
    `excludeEnrollmentIds`).
  - `at_client_secret_sharing.dart` — `AtClientSecretSharing`, the ready-made entry point.
  - `secret_sharing.dart` — barrel export.

[...]

PR #2083, opened 2026-07-17
feat: improve at_auth to at_client seam before moving to at_auth 4

Xlin123:
## What

Tightens the `at_auth` → `at_client` hand-off ahead of `at_auth` 4, closing the gap left by #2076 and #2047.

Auth used to hand the client its own live `AtChops`/`AtLookUp`. Now it hands back an `AtAuthSession` — a key *source* (`AtKeysIo`) plus connection details — and the client rebuilds its own state from it. Both the authenticate/onboard and the enrollment paths now speak sessions.

Fully additive: nothing is removed. Superseded fields are `@Deprecated` but stay populated, so existing callers keep working — removal is the `at_auth` 4 step.

## How

**Auth → session hand-off**
- New `AtAuthSession` { atSign, rootDomain, namespace, atKeysIo, enrollmentId, atLookUp? }.
- `AtAuthImpl.authenticate`/`onboard` build and attach `session` on success.
- `AuthResponse` + children `@Deprecated`; `atChops`/`atLookUp`/`atAuthKeys` superseded by `session`.
- `AtClientManager.setFromAuthSession(session, pref)` — client derives its own `AtChops` via `atKeysIo.read()` rather than adopting auth's socket (optional `reuse:` keeps the old connection as a perf escape hatch).

**Enrollment → session**
- `AtEnrollmentRequest` gains `session` as the source of `atSign`/`rootDomain` (both now `@Deprecated` params).
- `AtEnrollmentResponse` gains `session`; `atSign`/`rootDomain`/`atAuthKeys` `@Deprecated` (still populated).
- On approval, `waitForApproval` flushes the completed keyset into the session's `atKeysIo` and hands back a ready-to-use `session`. No session supplied → legacy path (populate `atAuthKeys`, no session).

## How to verify

```bash
cd packages/at_auth   && dart analyze && dart test
cd ../at_client       && dart analyze && dart test
```

End-to-end against a functional atServer: `authenticate` (or `submit` + `waitForApproval`) → `setFromAuthSession(response.session!, pref)` → assert a get/put round-trips on a freshly rebuilt socket.

## Skills impact


Xlin123:
@gkc @srieteja can i get some eyes on this? 

there was some work needed to actually set a good transition state for at_client to consume v4.

1. I'm under the assumption that #2037 will be flushing the keypackage required before any AtEnrollment operation. Is this correct?
2. Do you think I should put a lock on the FileAtKeysIo write?

gkc:
1. The sss will be for steady state communication of secrets between the authorized enrollments; I don't think it is needed for enrollment approval time 

What I had in mind was something like this
Initial enrollment (onboarding) cuts initial stuff. Subsequent enrollments will have enrollment requests which include a keyPackage for the approver to encapsulate the various keys that the approver holds which the requester can the decapsulate from the enrollment record once it has been approved. And writes those decapsulated secrets to AtKeysIo 

2. A lock at which point? What's the potential race you want to guard against?

gkc:
To restate : enrollment and approval is fundamentally the same - but using pq stuff instead of RSA (& AES). 

Will / should the approver use the sss to share the secrets? Maybe - but the sss required the target's key package to already exist on the keystore so its design assumption is limiting here. Maybe a later optimization? I'll need to think about it

Xlin123:
ah okay, much clearer.

so we're comfortable with just using a pqSeal/unseal and shouldn't really have to worry about ever switching what keytypes were using here?

and that then erases my concern(misunderstanding) that was probing me to put a lock in place.

gkc:
... maybe? Can you lay out the sequence of application events you're thinking about?

Xlin123:
1. Request time: sends KeyPackage Public Key
2. Approval time: seals symmetric and encryption key to the keypackage
3. Request Completion: unseals keys, writes then to IO

is roughly the mental model I have

packages/at_client/lib/src/client/at_client_impl.dart
```diff
@@ -1141,6 +1141,15 @@ class AtClientImpl implements AtClient {
   }
 
   Future<AtChops> _createAtChops(String atSign) async {
+    // When the client was handed an AtKeysIo *source* (and no live
+    // AtChops/AtLookUp was injected by auth), derive our own PKAM+encryption
+    // AtChops from that source instead of reading key material out of the local
+    // secondary. This is what lets a freshly rebuilt connection PKAM on its own
+    // socket — parity with the AtChops auth used to inject.
+    if (_atKeysIo != null) {
```
Xlin123, line 1149: will leave when at_client consumes at_auth v4

packages/at_auth/lib/src/enroll/models/at_enrollment_request.dart
```diff
@@ -35,21 +36,39 @@ abstract class EnrollmentRequest {
...
-      required super.deviceName,
-      super.apkamPublicKey,
-      super.rootDomain,
-      required this.otp,
-      required this.namespaces,
-      this.encryptedAPKAMSymmetricKey,
-      this.apkamKeysExpiryDuration});
+  AtEnrollmentRequest({
+    this.session,
+    @Deprecated('Provide `session` instead; its atSign is used.') String? atSign,
```
Xlin123: follow same deprecation pattern across files (either parameters or fields)

packages/at_auth/lib/src/auth/models/at_auth_session.dart
```diff
@@ -0,0 +1,33 @@
...
+///
+/// The confirmed subset of an [AtAuthRequest] that client creation needs,
+/// promoted to its own type so "request" doesn't quietly double as "session".
+/// Keys cross the boundary as an [AtKeysIo] *source* — the client derives its
+/// own [AtKeys] via `atKeysIo.read(atSign)`, rather than adopting
+/// auth's live `AtChops`/`AtLookUp`.
+class AtAuthSession {
+  final String atSign;
+  final AtRootDomain rootDomain;
+  final String? namespace;
```
Xlin123, line 15: namespaces, use NamespacePermissions

packages/at_client/lib/src/manager/at_client_manager.dart
```diff
@@ -174,6 +179,38 @@ class AtClientManager {
...
+  /// Consumes an [AtAuthSession] (the key *source* + confirmed params) and lets
+  /// the client rebuild its own connection: [session.atKeysIo] flows to
+  /// `AtClientImpl.create(atKeysIo:)`, which derives its own PKAM [AtChops] and
+  /// opens a fresh socket that PKAMs itself. Costs one extra PKAM handshake at
+  /// startup — the accepted price of clean auth/client separation.
+  ///
+  /// Set [reuse] to adopt auth's already-authenticated connection
+  /// ([session.atLookUp]) and skip the second handshake — the perf escape hatch.
+  /// When false (default) the client opens its own fresh socket.
+  Future<AtClientManager> setFromAuthSession(
```
Xlin123: fromAuthSession

packages/at_auth/lib/src/auth/models/at_auth_session.dart
Xlin123: actually doesnt make sense for auth we're keeping it a string

packages/at_auth/lib/src/enroll/models/at_enrollment_request.dart
Xlin123: @srieteja, these actually have to be param deprecations because they conflict with the FirstEnrollmentRequest for v3.3.0

so they might give misleading deprecations flags throughout 3.3.0

packages/at_client/lib/src/manager/at_client_manager.dart
Xlin123: [132b2c5](https://github.com/atsign-foundation/at_client_sdk/pull/2083/commits/132b2c5b182844093622e45b0779baff8bfaac9d)

packages/at_auth/lib/src/auth/models/at_auth_session.dart
srieteja, line 15: We decided to call this `namespaces` during the peer review.

packages/at_auth/lib/src/enroll/models/at_enrollment_request.dart
```diff
@@ -35,21 +36,42 @@ abstract class EnrollmentRequest {
...
-      required this.otp,
-      required this.namespaces,
-      this.encryptedAPKAMSymmetricKey,
-      this.apkamKeysExpiryDuration});
+  AtEnrollmentRequest({
+    this.session,
+    @Deprecated('Provide `session` instead; its atSign is used.')
+    String? atSign,
+    @Deprecated('Provide `session` instead; its rootDomain is used.')
+    AtRootDomain? rootDomain,
```
srieteja, line 57: Why not deprecate these params in the parent class?

packages/at_auth/lib/src/auth/models/at_auth_session.dart
Xlin123: unlike enrollments, you can only auth to a singular namespace,  this is a legacy feature before enrollments

packages/at_auth/lib/src/enroll/models/at_enrollment_request.dart
Xlin123: @srieteja, these actually have to be param deprecations because they conflict with the FirstEnrollmentRequest for v3.3.0

so they might give misleading deprecations flags throughout 3.3.0

srieteja (approved):

Issue #2011, opened 2026-06-22
S-5 — at_auth 4.0.0: WASM barrel split (breaking)


gkc:
Part of #1889.

Move `FileAtKeysIo` and the keyfile lock out of the main barrel into `at_auth_io.dart`, drop the `atKeysIo ??= FileAtKeysIo()` default so callers inject one, extract the `dart:io` socket probe, and move the registrar fully onto `package:http`.

This is at_auth's one breaking cut, so it goes out as 4.0.0. It sits off the D1 critical path, and it's a different major at a different time from at_client's 4.0 (#2016).

It builds on S-3 (#2007) rather than S-1, so the extended `AtKeys`/`AtKeysIo` and the updatable stores bake on a published 3.x first.

**Acceptance.** No at_auth source reachable from `at_auth.dart` imports `dart:io`, and the `FileAtKeysIo()` default is gone. Don't gate on a real `dart compile wasm` of the core, since the at_lookup and at_chops WASM splits are a separate piece of work.

**State at 2026-08-18: not started.** `at_auth_impl.dart:238` still carries the default, and 4 files under `lib/` import `dart:io`. The registrar does already use `package:http`, but wraps `HttpClient`, so that part is half done.

gkc:
@Xlin123 I'm going to pass this one to you, you're best positioned to tackle this alongside the AtKeys enhancement work

Xlin123:
can I nail down at_chops 4.0.0 along with this?

a version of at_auth 4.0.0 should have no visibility on the deprecated AtChops model

and nor should the consumers

gkc:
> can I nail down at_chops 4.0.0 along with this?
> 
> a version of at_auth 4.0.0 should have no visibility on the deprecated AtChops model
> 
> and nor should the consumers

SGTM

Xlin123:
just finished a conversation with @srieteja 

I'm going to wait on at_chops 4.0.0, there's nothing I _really_ need it for aside from cleaning up the code. We'll land it later, (#2099).

The plan for the at_auth 4.0 break will be this:
1. let the package gap fix (#2083) through to 3.3.0-rc2  
2. tag the commit for 3.3.0-rc2 so we can publish later
3. because of #2009, we might end up changing the `waitForApproval` and enrollment mechanisms, so I want to break at_auth in small PRs
4. remove at_auth from workspace and start breaking at_auth
[...]

PR #2099, opened 2026-07-21
feat(at_chops): 4.0.0 break

Xlin123:
## What

Part of **#2011** (S-5 — at_auth 4.0.0: WASM barrel split). Stacked on top of
**#2083** (`xl/improve-auth-model`) — that branch is the head this branch is based
on, so review/merge #2083 first.

`at_chops` **3.4.1 → 4.0.0**. Three goals:

1. **Remove the entire deprecated compatibility surface** — the `@Deprecated` API
   accumulated across 3.3.0 / 3.4.0, much of it explicitly tagged "removed in v4".
2. **Make every algorithm stateless** — no key fields, no setters. Key material is
   raw `Uint8List` passed per call, so one instance is safe to share and there is
   no "did I set the key yet?" failure mode.
3. **Guarantee WASM compatibility** — the pure-Dart barrel
   `package:at_chops/at_chops.dart` compiles under dart2wasm, enforced in CI.

Downstream is **not** migrated in this PR. `at_chops` is removed from the monorepo
pub `workspace:`, so consumers (`at_client` / `at_auth` / `at_lookup` /
`at_onboarding_cli`, all on `^3.x`) resolve against **hosted at_chops 3.x from
pub.dev** and the monorepo stays green. Migrating them onto the v4 direct-algorithm
API is separate follow-up work.

### Stateless algorithms

Every algorithm now takes its key material per call. `encrypt`/`decrypt` gained a
second positional `key` argument; the signing algorithms moved to
`signBytes`/`verifyBytes` with key material in required named parameters.

```dart
// before                                    // after
final aes = AESEncryptionAlgo(aesKey);        final aes = AesCtrEncryptionAlgo(32);
await aes.encrypt(data, iv: iv);              final key = aes.generateKey();
                                              //or AtBytes key = AtKeys.getKey().bytes;
                                              await aes.encrypt(data, key, iv: iv);

final rsa = RsaEncryptionAlgo();              final rsa = RsaEncryptionAlgo();
rsa.atPublicKey = pub;                        rsa.encrypt(data, publicKey);
rsa.encrypt(data);
```

- `SymmetricEncryptionAlgorithm` / `ASymmetricEncryptionAlgorithm` are now
  standalone interfaces. The generic `AtEncryptionAlgorithm<T, V>` base is gone —
  it was never used as a type and both subtypes pinned `T`/`V` to `Uint8List`.
- **`iv` is required** on `SymmetricEncryptionAlgorithm.encrypt`/`decrypt`. Nothing
  generates an IV for you and nothing hands a generated one back, so the caller
  must own it; reusing a (key, iv) pair is a security bug, and an IV that can't be
  reproduced can't decrypt. Wrong-length IVs are rejected. AES-CTR call sites that
  relied on the old implicit all-zeroes IV now pass `InitialisationVector.legacy()`
  explicitly — same bytes, visible at the call site.
- **`generateKey()`** is on `SymmetricEncryptionAlgorithm`: the implementation owns
  the length (GCM always 32 bytes, CTR whatever it was constructed for), so a
  caller holding the interface never has to know it. Replaces `AESKey.generate`.
  One internal secure random source (`src/secure_random.dart`) backs all of them.
- `RsaSigningAlgo` / `EccSigningAlgo` / `Ed25519SigningAlgo` implement
  `AtSignatureAlgorithm` — `generateKeyPair()` / `signBytes()` / `verifyBytes()`,
  raw key bytes per call (Ed25519: 32-byte seed; ECC: 32-byte scalar + 65-byte
  uncompressed point; RSA: DER). `RsaSigningAlgo` takes `hashingAlgoType` / `keySize`
  as named constructor parameters.

### Removed (breaking)

- Facade & helpers: `AtChops`, `AtChopsImpl`, `AtChopsKeys`, `AtChopsUtil`,
  `AtEncrypted`, `AtHashingAlgorithmFactory`.
- Metadata/result/input types: `AtSigningInput`, `AtSigningVerificationInput`,
  `AtSigningMode`, `AtSigningResult` (+`Type`), `AtSigningMetaData`,
  `AtEncryptionResult` (+`Type`), `AtEncryptionMetaData`.
- **All key wrapper classes** — both the deprecated `At*KeyPair` set and their
  replacements (`RsaKeyPair`, `X25519KeyPair`, `MlDsa65KeyPair`, `MlKem768KeyPair`,
  `XWingKeyPair`), the `AbstractKeyPair`/`AsymmetricKeyPair` hierarchy, and the
  `RawKeyPairBytes` mixin.
- **Everything under `src/key/`** — `AESKey`, `AtPublicKey`, `AtPrivateKey`,
  `SymmetricKey`, `AbstractKey`, the misnamed `Ed25519Key` (it was an AES symmetric
  key, not an Ed25519 key), `KeyType`, and the `key_names.dart` constants. Key
  material is raw bytes now; `base64Encode`/`base64Decode` at your own storage
  boundary.
- `AtSigningAlgorithm` interface + `DefaultHash` / `DefaultSigningAlgo` /
  `PkamSigningAlgo` wrappers.
- Stateful `secretKey` setter + `sign` / `verify` on `MlDsa65PureDartAlgo` /
  `MlDsa65FfiAlgo`.
- `StringAESEncryptor` — encode to bytes and use `AesCtrEncryptionAlgo`, or
  `AesGcm256EncryptionAlgo` which authenticates.
- `AesCtrFactory` — its key-length dispatch is folded into the
  `AesCtrEncryptionAlgo` constructor.
- `RsaEncryptionAlgo.fromKeyPair` and the `atPublicKey`/`atPrivateKey` fields (and
  with them the "public/private key not set" exceptions).
- `AtKeysCrypto` (relocated to `at_auth`).

### Renamed / moved

- `AESEncryptionAlgo` → **`AesCtrEncryptionAlgo`**, in `src/encryption/aes_ctr.dart`.
  The old name didn't say which mode it was, which mattered once `AesGcm256EncryptionAlgo`
  landed alongside it. Takes its key length (16/24/32) as the constructor argument.
- `AesGcm256FfiAlgo.fromLib(lib, key)` → `fromLib(lib)`; `AesGcm256EncryptionAlgo(key)`
  → `AesGcm256EncryptionAlgo()`.
- `AtPqc.aesGcm256` is a static field rather than a factory method, matching
  `AtPqc.xWing` / `AtPqc.mlDsa65` — it no longer needs a key at construction, so
  `AtPqc.aesGcm256.generateKey()` works through the interface.
- **`lib/src/algorithm/**` flattened up one level to `lib/src/**`** — e.g.
  `src/algorithm/encryption/aes_gcm.dart` is now `src/encryption/aes_gcm.dart`. All
  of it is under `src/`, so only code reaching past the public barrels is affected.

### Packaging / workspace

- Removed `packages/at_chops` from the root `workspace:` list; dropped
  `resolution: workspace`; bumped `version` to `4.0.0`; fixed the pubspec
  description (`at_protocol` → `Atsign Protocol`).
- Pruned the unused `dart_periphery` dependency (FFI/IO, WASM-hostile).
- Deleted `example/zariot` (it was built on the removed facade + `dart_periphery`);
  `example/pq` is unaffected.

### WASM

- Added `tool/wasm_compat_check.dart` (imports only the web-safe barrel — never
  `at_chops_ffi.dart`) and a `wasm_smoke` job (`dart compile wasm`) to
  `.github/workflows/at_libraries.yaml`. `at_chops_ffi.dart` stays non-WASM by
  design.

## How to verify

```bash
cd packages/at_chops
dart pub get
dart analyze                                   # clean
dart format . -o none --set-exit-if-changed    # clean
dart test --exclude-tags ffi                   # 94 tests pass
dart compile wasm tool/wasm_compat_check.dart -o /tmp/at_chops.wasm
```

Monorepo stays green — consumers resolve hosted at_chops 3.x:

```bash
# from repo root
dart pub get                                   # resolves at_chops as hosted 3.x
cd packages/at_client && dart pub get && dart analyze
```

The ML-DSA-65 / ML-KEM-768 / X-Wing FFI tests need **OpenSSL ≥ 3.5** (see the
`build_and_test_ffi` CI job); on a host with older libcrypto they fail with
`libcrypto does not support ML-DSA-65`. The AES-GCM FFI tests only need OpenSSL 3
and do run. A full `dart test` here is 110 passing / 19 failing, all 19 being that
libcrypto version gate — the `build_and_test_ffi` job is the real check on those.

## Skills impact

(This PR only touches `at_chops`; `at_client` / `at_client_flutter` / `at_auth`
consumers stay on published at_chops 3.x, so skills are unaffected.)


Xlin123:
going to wait on this PR until we settle on all the pq stuff

Xlin123:
waiting on some 3.x PRs to pass through then going for the break.

srieteja:
Five issues from a review pass over this PR — two worth fixing before merge, three smaller.

**1. The WASM gate no longer covers at_chops** — `.github/wasm_gates.yaml`

Dropping `packages/at_chops` from the root `workspace:` list means a root `dart pub get` resolves at_chops from the pub cache (3.x, pulled in via `at_client`'s `^3.6.0`) rather than from this tree. The `wasm_ratchet` job therefore walks published 3.x, so the ratchet isn't guarding the barrels this PR rewrites.

The tell is in the same file: the `controls` entry still asserts `at_chops_ffi.dart` reaches `lib/src/algorithm/ffi/openssl_ffi_bindings.dart`, which this PR moved to `lib/src/ffi/`. That control names a path that no longer exists and is still passing — because it's being checked against the cached 3.x copy.

Fix: keep at_chops resolving locally for the gate (workspace entry or a root override), and repoint the control at `lib/src/ffi/openssl_ffi_bindings.dart`.

**2. `at_cli_commons/example` missed the override cleanup** — `packages/at_cli_commons/example/pubspec.yaml:25`

Five example pubspecs had their `at_chops` path override removed with a `TODO(at_chops-4.0.0)` note. This one still has it, so it resolves local 4.0.0 into a graph whose `at_client` still calls the removed 3.x facade (`AtChopsKeys` / `AtChopsImpl`, `at_client_impl.dart:2399-2401`) and the example won't build.

Worth noting `dart analyze` from that directory passes — the analyzer doesn't type-check method bodies of path dependencies — so this only shows up on `dart compile` or a real build.

Fix: drop the two override lines, same as the other five.

**3. `RsaSigningAlgo.name` mislabels any non-2048 key** — `packages/at_chops/lib/src/signing/rsa.dart:52`

`_name = keySize == 2048 ? rsa2048 : rsa4096`, and `keySize` is unvalidated — so `RsaSigningAlgo(keySize: 3072)` announces `rsa4096`. Separately, `verifyBytes` takes the modulus from the DER public key and ignores `_keySize`, so a default-constructed instance verifying a 4096-bit signature announces `rsa2048`.

Since `name` is the wire identifier and `SigningAlgoType.strongestFirst` ranks `rsa4096` above `rsa2048`, a mislabel also moves the entry in preference ordering — so this may be worth more than a cosmetic fix.

Fix: validate `keySize` against the supported sizes, or derive `name` from the actual key.

**4. `Ed25519Key` migration note points at a class this release deletes** — `packages/at_chops/CHANGELOG.md:17`

"remove the misnamed `Ed25519Key` class — use `AESKey`", but `AESKey` is removed in the same release (line 15), leaving that breaking change with no working migration path. The actual replacement is raw `Uint8List` key bytes plus `SymmetricEncryptionAlgorithm.generateKey()`.

**5. README AES-CTR paragraph still describes 3.x IV handling** — `packages/at_chops/README.md:290`

It says the pure-Dart path right-pads a shorter IV and only the FFI path rejects it. In 4.0.0 `AesCtrEncryptionAlgo` throws on any IV that isn't exactly 16 bytes on both paths (`aes_ctr.dart:69`, `:84`) — which is what the doc comment on `at_pqc.dart:66-69` already says.

srieteja:
Thanks, all five are fixed in ac3b8a204. There's one new issue:

**ECDSA signature format changed without a note.** In 3.x, `ecc_secp256r1` signatures were 128 bytes of ASCII hex (`toCompactHex().codeUnits`). `ecc.dart:74-76` now emits 64 raw R‖S bytes under the same name, so 3.x and 4.0 signatures don't verify against each other. `golden_vectors_test.dart:219` is labelled "captured from at_chops 3.x", but its vector is the new 64-byte form, so the test doesn't catch this. `CHANGELOG.md:28` also says "every wire format is unchanged".

Fix: either keep 3.x compatibility (emit hex, accept both lengths in `verifyBytes`, regenerate the vector from real 3.x output), or document the break and fix the test label. In both cases `verifyBytes` should check the signature length.

PR #2110, opened 2026-07-29
refactor(at_auth)!: at_auth 4.0.0-beta1

Xlin123:
## What

Takes **at_auth** to `4.0.0`.

Onboarding and authentication stop trafficking in objects. `AtAuthSession`, `AtAuthRequest`, `AtOnboardingRequest`, `AuthRequest` and the whole `AuthResponse`/`AtAuthResponse`/`AtOnboardingResponse` hierarchy are deleted. Every entry point takes the atsign, its atServer and a key source as direct parameters and returns `void` — **completing without throwing is the success signal**, and failure throws `AtAuthenticationException`.

The second half is a consequence of at_lookup 4: a connection now binds its PKAM key and signing algorithm at construction and keeps them immutable, so the connection cannot exist until the keys have been read or minted. That means **at_auth has to build it**, rather than being handed one.

This new consequence allows us to push at_auth down the stack allowing it to properly become an AtClient concern not a user-concern. 
```dart
// in at_client, name tbd, could be: onboard, enroll and login!
static Future<AtClient> login(
    Atsign atsign, AtRootDomain rootDomain, AtKeysIo io) async {
  final auth = AtAuth.create();
  // no connection to pass in: at_auth reads the keys from io, builds the
  // AtLookUp bound to them, and authenticates it
  await auth.authenticate(atsign, rootDomain, io);
  return AtClient.create(atLookUp: auth.atLookUp!, ...);
}
```

> This branch also carries at_chops 4.0.0 and at_lookup 4.0.0, which at_auth depends on. Those are reviewed separately in #2099 and #2123 — this diff shrinks to at_auth alone once they merge.

### Direct parameters, void returns

```dart
// before — caller built the connection and injected it
final atLookUp = AtLookupImpl(atsign, rootDomain, rootPort);
final atAuth = AtAuth.create(atLookUp: atLookUp);
final response = await atAuth.authenticate(AtAuthRequest(atsign, io));
if (response.isSuccessful) { useChops(atAuth.atChops!); }

// after — at_auth builds the connection; no throw means authenticated
final atAuth = AtAuth.create();
await atAuth.authenticate(atsign, AtRootDomain.atsignDomain, io);
final authenticated = atAuth.atLookUp!;   // hand this to client creation
```

See `example/authenticate.dart` for the full runnable version.

- `authenticate(Atsign, AtRootDomain, AtKeysIo, {enrollmentId})`
- `onboard(Atsign, AtRootDomain, AtKeysIo, cramSecret, {autoCompleteActivation, appName, deviceName})`
- `completeActivation(Atsign, AtRootDomain, AtKeysIo)`
- `validateAtServer(Atsign, AtRootDomain, {onboarding})`

`RetryOptions` moves from the request onto `AtAuth.create({options})`. The request types' `namespace`, `signingAlgoType` and `hashingAlgoType` fields had no consumer at all and are gone with no replacement.

### at_auth builds the connection

`AtAuth.create` takes no `AtLookUp`. Each call builds its own once it has keys in hand. `AtAuth.atLookUp` then exposes the connection the last successful call authenticated — that is how a void-returning call still hands something forward, and it lets a client built from it skip a second PKAM handshake. Null before the first success, and null again after a failure (the connection is closed rather than leaked).

Two flows fall out of the immutability:

- **`onboard` now uses two connections.** A keyless one carries CRAM auth *and* the first enrollment (the atServer auto-approves it over CRAM); it is then closed, and a second one — built from the keypair that did not exist when the first was constructed — does PKAM. The existing docstring already described this shape; it just wasn't true before.
- **`waitForApproval` builds its own** from `pending.atKeys`, since the injected connection belongs to whoever submitted the request and cannot sign for the new enrollment. This is what its previously-unused `rootDomain` parameter is for.

### The signing scheme is the caller's choice

```dart
AtAuth.create(signing: ApkamSigning.postQuantum);   // default: .legacy
// AtEnrollment still takes a connection — the caller's, already authenticated.
// Its `signing` picks the scheme waitForApproval builds the *new* enrollment's
// connection with.
AtEnrollment.create(atLookUp, signing: ApkamSigning.postQuantum);
```

`ApkamSigning.legacy` is RSA-2048/SHA-256, `postQuantum` is ML-DSA-65. It is deliberately **not** inferred from key material: `AtKeys.generate` mints both a classical and a post-quantum APKAM key, so the keyset cannot express which one an atServer expects — that is a property of the deployment, and the application (at_client) owns it. A keyset that cannot satisfy the chosen scheme throws rather than falling back to the other one, which would authenticate as an identity the caller did not ask for.

### Keys and enrollment

- `AtKeysIo` speaks `Atsign` throughout (`read`/`write`/`flush`); previously `read` took `Atsign` and `write` took `String`, and because `Atsign` is a subtype of `String` the analyzer could never flag the drift. `write` moves up to the base type; `flush` is the runtime-persist op.
- `AtKeys.generate(Atsign, {enrollmentId, mintLegacy})` replaces any key-gen methods in the package.
- `AtKeysMaterial.bytes` is a `Uint8List`, not an `AtBytes` — base64 is a property of how a *keyfile* stores material, so it applies only at the JSON boundary.
- `AtEnrollment` takes its `AtLookUp` at construction; `approve(Atsign, AtKeysIo, decision)` reads the approver's own keys from an `AtKeysIo` like every other key consumer. `waitForApproval`'s retry defaults move onto `AtEnrollment.defaultRetryInterval`/`defaultMaxRetries` so the interface and impl can no longer disagree about them (they did — 48 retries/1 min vs 15/2 s, resolved off the static receiver type).
- `PkamAuthenticator.authenticate` drops its `AtKeys` parameter. It never used it; the signing key reaches PKAM through the connection.
- `RetryOptions`, `CramAuthenticator`, `PkamAuthenticator`, `KeyIds` and `ApkamSigning` are now exported — they were public parameter and field types consumers could not name.

### Fixes

Two round-trip bugs in the typed-keys document: `namespaces` was written as a comma-joined string but read back with `optionalStringList`, and was not registered in `KeyIds.reservedTopLevelKeys`, so it also leaked into `AtKeys.metadata` on read.

### Compatibility

**The `.atKeys` on-disk format is unchanged.** Migrating off the at_chops key-pair wrappers swapped `RsaKeyPair`/`AESKey`/`AESEncryptionAlgo`/`StringAESEncryptor` for the stateless algorithms, but it is the same AES-CTR/PKCS7 under the same keys and IVs — and AES strength still follows the length of the keyfile's own `selfEncryptionKey` rather than being fixed at 256. Legacy flat keyfiles stay readable *and* writable, and passphrase-protected files still decrypt.

at_auth is detached from the pub workspace with path overrides (as at_lookup already is), so the rest of the repo keeps resolving hosted at_auth 3.x. `at_server_status` needed a matching 2.0.0 bump for at_lookup 4, since at_auth depends on it.

**Not migrated:** `at_client`, `at_onboarding_cli`, `at_client_flutter` and `tests/**` do not compile against local at_auth 4. That is the next PR, which adds `AtClient.pkam/apkam/cram(Atsign, AtRootDomain, AtKeysIo)` to replace `AtClientManager.fromAuthSession`.

## How to verify

at_auth is its own pub root:

```bash
cd packages/at_auth
dart pub get
dart analyze     # clean across lib/, test/ and example/
dart test        # 154 pass
```

```bash
dart pub get     # at the repo root: must succeed (it failed outright before this branch)
```

No virtualenv needed — the suite is unit-level with a mocked `AtLookUp`. Worth looking at specifically:

- `test/apkam_signing_test.dart` — the scheme is honoured, and a keyset that cannot satisfy it is rejected in **both** directions rather than falling back.
- `test/at_auth_test.dart` — `onboard` CRAMs on a keyless connection and PKAMs on a different, key-bearing one; a failed call leaves no connection exposed.
- `test/at_keys_io_test.dart`, `test/legacy_at_keys_test.dart` — the on-disk format assertions, which pin the compatibility claim above and pass unchanged.

The Docker functional suites (`tests/at_functional_test`, `tests/at_onboarding_cli_functional_tests`) are **expected to be red** until the downstream migration lands.

Manual smoke against a real atsign:

```bash
cd packages/at_auth
dart example/authenticate.dart -a @alice -k /path/to/alice_key.atKeys
dart example/onboard.dart -a @alice -c <cram> -k /path/to/write.atKeys
```

## Skills impact

- [x] Public API changed — I have opened or will open an `at_client_skills` PR to reflect this

at_auth's entire onboarding/authentication surface changed. `packages/at_client_skills/skills/at_client_skills-sdk/` still documents `AtAuthRequest`/`AtOnboardingRequest`/`AuthResponse` in `SKILL.md`, `references/05-flutter-auth.md` and `references/06-package-map.md`. Best done alongside the downstream `at_client` migration, since the skill documents both sides.


Xlin123:
huge diff right now from #2099 and #2123 being dragged into this PR

PR #2114, opened 2026-07-31
feat(at_chops): rsa signing algo under new AtSignatureAlgorithms

Xlin123:
Key changes:

* Introduced `RsaSignatureAlgo`, a stateless RSA signing implementation that takes key material per call and enforces consistency between the algorithm's wire identifier and key size. Supports both 2048-bit (for PKAM) and 4096-bit (for data/envelope signing) RSA, with digest selection limited to SHA-256 or SHA-512.

### Release and Changelog
will be put in 3.5.0 after 3.4.2 -> #2115, abstract names will be in #2099


gkc (commented):
Two things before merge:

**1. The getter has no consumer, and the mislabelling the CHANGELOG describes is still live.** `AtSigningInput.signingAlgorithm` is typed `AtSigningAlgorithm` (the deprecated interface), which didn't gain the getter — so `AtChopsImpl._signBytes` still takes the label from `signingInput.signingAlgoType`. `at_client`'s `envelope_signing.dart:74` never sets it, so `envelope['signingAlgo']` is always `rsa2048` and will mislabel an ML-DSA APKAM signature. Cheap fix here: in `_signBytes`, prefer `signingAlgorithm.signingAlgoType` when `signingAlgorithm is AtSignatureAlgorithm` — both ML-DSA classes implement both interfaces, so it works immediately. If it's deferred, say so in the CHANGELOG; as written it reads as fixed.

**2. Adding an abstract member to `AtSignatureAlgorithm`/`AtHashingAlgorithm` is source-breaking for implementers** - call it out as such in the changelog. @srieteja has #2115 which is also making changes to AtSignatureAlgorithm. 

Question for you both - is it right to have these as abstract interface classes at this stage? should they just be abstract classes?

Nits:
- `verifyBytes` throwing on a size-mismatched *public* key is stricter than the interface contract and than the ML-DSA impls; the mislabelling argument only applies to signing.
- `RsaSignatureAlgo.hashingAlgoType` isn't on the interface, so "both halves come off the algorithm" only holds when statically typed as `RsaSignatureAlgo`.
- CHANGELOG entries are paragraph-length where the rest of the file is one-liners; the rationale is already in the dartdoc.
- PKCS#1 v1.5 is deterministic — asserting `RsaSignatureAlgo` output is byte-identical to `RsaSigningAlgo`/`PkamSigningAlgo` would pin the migration story better than a self-round-trip.

Xlin123:
gonna wait for #2115, so that this PR is strictly for rsa signing algo

packages/at_chops/lib/src/algorithm/signing/rsa.dart
```diff
@@ -6,6 +7,154 @@ import 'package:at_chops/src/key/keys.dart';
...
+    };
+  }
+
+  RSAPrivateKey _parsePrivateKey(Uint8List der) {
+    final RSAPrivateKey key;
+    try {
+      key = RSAPrivateKey.fromString(base64Encode(der));
+    } catch (e) {
+      throw AtSigningException(
+          'Could not parse secretKey as a PKCS#8 RSA private key: $e');
```
srieteja: This exception message is a little confusing

packages/at_chops/test/rsa_signature_algo_test.dart
```diff
@@ -0,0 +1,175 @@
...
+      final small = await RsaSignatureAlgo.rsa2048().generateKeyPair();
+      final sig = await RsaSignatureAlgo.rsa2048()
+          .signBytes(message, secretKey: small.secretKey);
+      expect(
+          () => RsaSignatureAlgo.rsa4096()
+              .verifyBytes(message, signature: sig, publicKey: small.publicKey),
+          throwsA(predicate((e) =>
+              e is AtSigningVerificationException &&
+              e.toString().contains('2048-bit'))));
+    });
```
srieteja, line 157: A tamper test and another test asserting failure on SHA256-SHA512 interop would be a great addition here.

srieteja (changes requested):
A couple nits, everything else LGTM

packages/at_chops/lib/src/algorithm/signing/rsa.dart
Xlin123: [237d3e3](https://github.com/atsign-foundation/at_client_sdk/pull/2114/commits/237d3e3e6f6da8f0b00c6857bda180259e391da2) ty srie

packages/at_chops/test/rsa_signature_algo_test.dart
Xlin123: [237d3e3](https://github.com/atsign-foundation/at_client_sdk/pull/2114/commits/237d3e3e6f6da8f0b00c6857bda180259e391da2) 

2048 has a tamper but 4096 doesn't

gkc (dismissed):

JeremyTubongbanua (dismissed):

Xlin123:
don't merge until #2115 closes thank you

gkc:
> don't merge until #2115 closes thank you

@Xlin123 #2115 has been merged, you have merge conflicts here now

 @srieteja you need to re-approve as you requested changes in your last review

srieteja (dismissed):

gkc (approved):

gkc:
@srieteja can you re-approve this please? I am the last committer, dealt with the merge conflict (changelog, trivial)

srieteja (approved):

PR #2115, opened 2026-08-01
feat: pq algo spec and validation

srieteja:
## Why
- invalid base64: documented contracts not actually enforced. `MlDsa65KeyPair.create` promised `AtSigningException` for invalid key material but let non-base64 input through as a raw `FormatException`.
- no output-length checks existed: a backend returning a wrong-size ciphertext, shared secret, or signature (OpenSSL version mix-up, `pqcrypto`/`cryptography` bug) went undetected and propagated into buffers sized for the expected length.
- pure-Dart ML-KEM-768/X-Wing validation relied solely on `pqcrypto`'s private `KyberKem` validators, reachable only via an `implementation_imports` escape hatch that could silently stop validating in a future patch release.
- `MlDsa65FfiAlgo.verifyBytes` caught `Object` and returned `false`, so a libcrypto build lacking ML-DSA-65 was indistinguishable from a forged signature.

## What
- Output-length invariant checks for ML-KEM-768/ML-DSA-65/X-Wing (FFI + pure-Dart) — wrong-size ciphertext/shared-secret/signature throws a `StateError` unconditionally.
- Input length validation (public key, ciphertext, secret key) owned by `at_chops` itself at the FFI/pure-Dart boundary, rather than delegated to `pqcrypto`'s private validators.
- `MlDsa65KeyPair.create` throws `AtSigningException` for non-base64 key material instead of a raw `FormatException`.
- `MlDsa65FfiAlgo.verifyBytes` drops its catch-all: a signature mismatch still returns `false`, but a backend failure now surfaces as `StateError`.
- X-Wing size constants and byte-assembly/shared-secret-combiner logic consolidated into `XWingSizes`.

## How
- `algorithm/spec/{ml_dsa_65,ml_kem_768,x_wing}_spec.dart` and `spec/output_length.dart` hold the shared size constants, `checkOutputLength`, and validators — one source of truth used by both FFI and pure-Dart backends.
- `ml_dsa_65_key_pair.dart` wraps `base64Decode` and rethrows `AtSigningException` on a `FormatException`.
- `ml_dsa_65_ffi.dart`: nothing inside `verifyBytes`' `try` throws on a genuine signature mismatch — `EVP_DigestVerify` returns `0`, which is already `false`. The catch-all therefore only ever converted misconfiguration into "invalid signature", so it is removed; this also makes the FFI backend match the pure-Dart one, which carries no catch.

## How to verify

1. Test coverage: `ml_dsa_65_algo_test.dart`, `ml_dsa_65_ffi_test.dart`, `ml_kem_768_algo_test.dart`, `ml_kem_768_ffi_test.dart`, `x_wing_algo_test.dart`, `x_wing_ffi_test.dart`, `output_length_test.dart`.
2. `ml_dsa_65_ffi_test.dart` pins the boundary the removed catch used to blur: right-length garbage key/signature return `false`, so only a backend failure can produce a `StateError`.
3. `dart test` — all 237 tests pass.

## Skills impact


gkc (changes requested):
basically lgtm and useful ... a few things

- please add info to PR description as to why this is needed (there's no associated ticket)
- at_chops 3.4.1 is already published; this is a patch bump carrying an entry that says `breaking: AtSignatureAlgorithm gains an abstract String get name`. Adding an abstract member to an exported `abstract interface class` breaks every external `implements`. But is this actually OK? Even if it is, version should be 3.5.0
- `MlDsa65KeyPair.create` validates; but `AtMlDsa65KeyPair.create` doesn't
- `base64Decode` now runs in the constructor, so corrupt key material fails with `FormatException`, not the documented`AtSigningException`
- `verifyBytes` (FFI) swallows "backend unavailable" as "signature invalid".`ml_dsa_65_ffi.dart:166` catches `Object`.
  - On libcrypto < 3.5, `_loadPublicKey` throws and verify returns `false`. Fail-closed, so not a hole — but a misconfiguration becomes indistinguishable from an attack, and invisible in logs.
- Three new `checkOutputLength` guards in `x_wing_ffi.dart` / `x_wing_pure_dart.dart` aren't being directly tested as far as I can tell
- Pure-Dart ML-KEM validation is not being directly added, it is being delegated to `pqcrypto`'s *private* `KyberKem._validatePublicKey`
- Do not use "atProtocol" - use "Atsign Protocol".
- PR description looks out of date
- `validateSecretKey` / `hasValidVerifyLengths` are bare top-level names in `spec/ml_dsa_65_spec.dart` — will collide once a second algorithm wants the same. Statics on `MlDsa65Sizes` would scope them.
- New `spec/*.dart` dartdoc references `[MlDsa65FfiAlgo]`, `[MlDsa65KeyPair]`, `[AtSignatureAlgorithm]`, `[XWingFfiAlgo]` with nothing imported — links won't resolve.

gkc:
@srieteja : @Xlin123 has #2114 which is also making changes to AtSignatureAlgorithm. 

Question for you both - is it right to have these as abstract interface classes at this stage? should they just be abstract classes?

srieteja:
> @srieteja : @Xlin123 has #2114 which is also making changes to AtSignatureAlgorithm.
> 
> Question for you both - is it right to have these as abstract interface classes at this stage? should they just be abstract classes?

It should be an abstract class at this stage. Making the change as part of coming commits

Xlin123:
we're still communicating hashingAlgo over the wire, can we have the HashingAlgorithms also have a name getter?

packages/at_chops/CHANGELOG.md
Xlin123, line 1: I can pull the breaking changes in my at_chops 4 changes and we can go ahead and merge/publish this.

I know we don't really have any external consumers of these members that are breaking so I am _okay_ to publish this, but we should try our best to adhere to semver when we can. (a little more work, but cleaner)

Xlin123 (changes requested):
I'm going to cherry-pick the breaking commits and then you can revert them so we can publish 3.5.0.

packages/at_chops/lib/src/algorithm/signing/ml_dsa_65_ffi.dart
```diff
@@ -49,37 +66,54 @@ final class MlDsa65FfiAlgo implements AtSigningAlgorithm, AtSignatureAlgorithm {
...
+          libCryptoSupportsMlDsa65}) {
+    if (!supportsMlDsa65(_lib)) {
+      throw AtSigningException(
+          'this libcrypto build does not provide ML-DSA-65 — use '
+          'MlDsa65PureDartAlgo, or AtPqc.mlDsa65 to auto-select');
+    }
     _ctxNewFromName = _lib.lookupFunction<EvpPkeyCtxNewFromNameNative,
         EvpPkeyCtxNewFromNameDart>('EVP_PKEY_CTX_new_from_name');
     _ctxFree = _lib.lookupFunction<EvpPkeyCtxFreeNative, EvpPkeyCtxFreeDart>(
         'EVP_PKEY_CTX_free');
```
srieteja, line 63: format only start

packages/at_chops/lib/src/algorithm/signing/ml_dsa_65_ffi.dart
```diff
@@ -49,37 +66,54 @@ final class MlDsa65FfiAlgo implements AtSigningAlgorithm, AtSignatureAlgorithm {
...
-        EvpDigestVerifyInitDart>('EVP_DigestVerifyInit');
-    _digestVerify = _lib
-        .lookupFunction<EvpDigestVerifyNative, EvpDigestVerifyDart>(
+    _digestVerifyInit =
+        _lib.lookupFunction<EvpDigestVerifyInitNative, EvpDigestVerifyInitDart>(
+            'EVP_DigestVerifyInit');
+    _digestVerify =
+        _lib.lookupFunction<EvpDigestVerifyNative, EvpDigestVerifyDart>(
             'EVP_DigestVerify');
   }
```
srieteja, line 91: format only end

packages/at_chops/lib/src/algorithm/signing/ml_dsa_65_ffi.dart
```diff
@@ -223,7 +272,9 @@ final class MlDsa65FfiAlgo implements AtSigningAlgorithm, AtSignatureAlgorithm {
     try {
       final Pointer<EVP_PKEY> pkey =
           _newRawPrivateKeyEx(nullptr, algName, nullptr, buf, keyBytes.length);
```
srieteja, line 253: format only

packages/at_chops/lib/src/algorithm/signing/ml_dsa_65_ffi.dart
```diff
@@ -279,8 +332,7 @@ final class MlDsa65FfiAlgo implements AtSigningAlgorithm, AtSignatureAlgorithm {
     }
   }
```
srieteja, line 312: format only

packages/at_chops/test/ml_dsa_65_ffi_test.dart
```diff
@@ -30,17 +32,25 @@ void main() {
...
+              'protocol sees one identifier regardless of backend');
+      expect(algo.signingAlgoType, equals(SigningAlgoType.mldsa65),
+          reason: 'must match MlDsa65PureDartAlgo.signingAlgoType — both '
+              'backends report the same pkam:/envelope wire identifier');
       final kp = await algo.generateKeyPair();
 
       expect(kp.publicKey.length, equals(1952));
       expect(kp.secretKey.length, equals(4032));
 
       final Uint8List message =
```
srieteja, line 41: format only — start

packages/at_chops/test/ml_dsa_65_ffi_test.dart
```diff
@@ -98,10 +109,102 @@ void main() {
       final kp = await algo.generateKeyPair();
 
       final Uint8List message = Uint8List.fromList('original'.codeUnits);
-      final Uint8List sig = await algo.signBytes(message, secretKey: kp.secretKey);
+      final Uint8List sig =
+          await algo.signBytes(message, secretKey: kp.secretKey);
```
srieteja, line 88: format only — end

packages/at_chops/test/x_wing_ffi_test.dart
```diff
@@ -26,7 +27,8 @@ void main() {
       }
     });
```
srieteja, line 28: format only start

packages/at_chops/test/x_wing_ffi_test.dart
```diff
@@ -109,7 +111,8 @@ void main() {
       expect(ss, equals(enc.sharedSecret));
     });
 
-    test('a tampered ciphertext decapsulates to a different secret, not an error',
+    test(
+        'a tampered ciphertext decapsulates to a different secret, not an error',
         () async {
       if (lib == null) {
```
srieteja: format only end

Xlin123 (dismissed):
LGTM, update the PR description and have @gkc re-review

packages/at_chops/lib/src/algorithm/spec/x_wing_spec.dart
```diff
@@ -0,0 +1,84 @@
...
+  /// — same reasoning as [combineSharedSecret]'s.
+  static Uint8List assemblePublicKey(
+      Uint8List mlKemPublicKey, Uint8List x25519Public) {
+    checkOutputLength(mlKemPublicKey.length, MlKem768Sizes.publicKeyBytes,
+        operation: 'ML-KEM-768 generateKeyPair', label: 'public key');
+    return Uint8List(publicKeyLength)
+      ..setRange(0, MlKem768Sizes.publicKeyBytes, mlKemPublicKey)
+      ..setRange(MlKem768Sizes.publicKeyBytes, publicKeyLength, x25519Public);
+  }
+
```
srieteja: Extracted the common logic out of X-Wing, Pure Dart, and ffi

srieteja:
| Commit | Summary | Addresses |
|---|---|---|
| [`2d5879e5`](https://github.com/atsign-foundation/at_client_sdk/commit/2d5879e5f) | `base64Decode` moved out of the constructor path so corrupt key material raises `AtSigningException`, not a raw `FormatException` | Wrong exception type surfaced on invalid base64 key material |
| [`78865332`](https://github.com/atsign-foundation/at_client_sdk/commit/78865332e) | Scoped `validateSecretKey`/`hasValidVerifyLengths` as statics on `MlDsa65Sizes`; added tests for `checkOutputLength` guards; fixed unresolved dartdoc links | Bare top-level validator names would collide; new length guards had no direct test coverage; dartdoc `[...]` links didn't resolve |
| [`e0ef3463`](https://github.com/atsign-foundation/at_client_sdk/commit/e0ef34639) | Added `MlKem768Sizes.validatePublicKey/validateCiphertext`; added `AtMlDsa65KeyPair.create` validation; closed X-Wing test-coverage gap. *(Its FFI unsupported-libcrypto throw was later reverted by `a3ad3259`.)* | `AtMlDsa65KeyPair.create` skipped validation that `MlDsa65KeyPair.create` did; ML-KEM validation relied on `pqcrypto`'s private, unversioned validator instead of an owned one |
| [`7362fa44`](https://github.com/atsign-foundation/at_client_sdk/commit/7362fa44f) | `atProtocol` → `Atsign Protocol` wording in `at_auth` comments | Incorrect protocol name in doc comments |
| [`a3ad3259`](https://github.com/atsign-foundation/at_client_sdk/commit/a3ad32590) | Reverted all 3 breaking changes and the abstract-class change; version demoted 3.5.0 → 3.4.2 | Breaking API changes shouldn't ship in a patch release |
| [`1076fad8`](https://github.com/atsign-foundation/at_client_sdk/commit/1076fad88) | Consolidated X-Wing's byte-assembly/shared-secret logic into static `XWingSizes` methods | Extends the statics-scoping fix from `78865332` to X-Wing, which had the same duplicated/untested-guard problem |
| [c1f7282](https://github.com/atsign-foundation/at_client_sdk/pull/2115/commits/c1f7282a0ba4bff47b249773d865c38927ab66b6) | Dropped verifyBytes' catch-all so a backend failure raises StateError while a signature mismatch still returns false; added right-length garbage key/signature tests pinning that boundary; version re-bumped 3.4.2 → 3.5.0 | verifyBytes swallowed "backend unavailable" as "signature invalid", making a misconfigured libcrypto indistinguishable from a forged signature. Chosen over a3ad3259's reverted construction-time throw: the catch only ever caught backend StateErrors, since EVP_DigestVerify reports a mismatch as a return code, not an error

srieteja:
@gkc  @Xlin123  all requested changes are in — thanks for the direct fixes via #2132, @gkc . Ready for re-review.

gkc (approved):

PR #2127, opened 2026-08-08
fix: assert the 'from:' challenge format

gkc:
## What I did

- A client now asserts a `from:` challenge has the shape an atServer issues — `_<uuid><atSign>:<uuid>` — and names the atSign it asked for, before signing it. It previously signed the response verbatim.
  - Both signing paths are guarded: `pkamAuthenticate` and the deprecated `authenticate(privateKey)`.
  - Fails closed, matching the inter-server POL path.
- `at_lookup` 3.6.0 -> 3.6.1. 3.6.0 is published, so this opens a new version rather than folding into an in-progress one.

## How to verify it

- Multiple new tests added

### DIFF
--- a/packages/at_lookup/CHANGELOG.md
+++ b/packages/at_lookup/CHANGELOG.md
@@ -1,3 +1,10 @@
+## 3.6.1
+
+- fix: strengthen the from challenge. A client now checks that a `from:`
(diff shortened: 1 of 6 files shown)


gkc:
The failing checks are a race; I've found the reason, am fixing.

cconstab (approved):

PR #2131, opened 2026-08-10
fix(at_chops): review fixes for RsaSignatureAlgo

gkc:
**- What I did**

Review fixes for #2114, based against that branch so it can be merged straight in.

- CHANGELOG named `RsaSignatureAlgorithm`/`RsaSigningAlgorithm` — neither class exists.
- The `RsaSigningAlgo` deprecation notice named `AtSigningAlgo`; the interface is `AtSigningAlgorithm`, and that string is the IDE tooltip consumers see.
- `DefaultSigningAlgo` and `PkamSigningAlgo` both pointed callers at `RsaSigningAlgo`, which #2114 deprecates — redirected to `RsaSignatureAlgo`. at_lookup has three live `PkamSigningAlgo` call sites that would have followed that advice.
- Moved `@Deprecated` below the doc comment. It applied either way, but above a blank line and the `///` block it reads as a dangling library annotation.
- Key-size mismatch errors now say what to do.
- Doc claimed 4096-bit keygen takes "seconds"; measured 560ms (2048-bit: 42ms).
- Added a legacy/new interop test — PKCS#1 v1.5 is deterministic, so the two must emit identical bytes, and that identity is what makes the swap transparent on the wire.
- Added known-answer vectors — every other test signs and verifies with itself, so all would stay green if the construction changed underneath.

Not addressed here, left for you as they are design calls rather than fixes: whether `verifyBytes` should throw or return false for bad key material (RSA throws, ML-DSA returns false, the interface documents neither); putting `signingAlgoType` on `AtSignatureAlgorithm`; wiring the class into `AtChopsImpl`. I also left the `## 3.5.0` heading alone since the version is yours to pick — but note pubspec.yaml is still on 3.4.1 and #2115 has to slot 3.4.2 in underneath it.

**- How I did it**

See commit & diffs.

**- How to verify it**

Tests pass. at_chops 192/192; analyze and format clean; at_lookup, at_client, at_auth and at_onboarding_cli clean under `--fatal-warnings`. Both new assertions were mutation-checked red before being trusted.

fix(at_chops): review fixes for RsaSignatureAlgo

### DIFF
--- a/packages/at_chops/CHANGELOG.md
+++ b/packages/at_chops/CHANGELOG.md
@@ -1,5 +1,7 @@
 ## 3.5.0
-- feat: add `RsaSignatureAlgorithm`, deprecating the old RsaSigningAlgorithm which implements the deprecated signing type.
+- feat: add `RsaSignatureAlgo`, a stateless `AtSignatureAlgorithm` implementation for RSA-2048 and RSA-4096 signing, with key material passed per call
+- deprecate: `RsaSigningAlgo`, which implements the deprecated `AtSigningAlgorithm` interface — use `RsaSignatureAlgo` instead. The two produce byte-identical signatures, so the swap is transparent on the wire
+- deprecate: redirect the `DefaultSigningAlgo` and `PkamSigningAlgo` deprecation notices at `RsaSignatureAlgo`; they previously pointed at `RsaSigningAlgo`, which is now itself deprecated
 
 ## 3.4.1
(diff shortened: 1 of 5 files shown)


Xlin123:
thanks @gkc

gkc:
> thanks @gkc

v welcome ... any thoughts on the things I mentioned in this PR description but didn't change in this PR?

Xlin123:
I am still planning to wait for the 2115, I'll update the pubspec and loosen the verifyBytes pattern

I'd like to tighten the verifyBytes to throw on 4.0.0

PR #2132, opened 2026-08-10
refactor(at_chops): review follow-ups for #2115

gkc:
Review follow-ups for #2115, on top of `st/pq-algo-size-spec` — merge into that branch to take them.

**- What I did**

- Dropped the X-Wing consolidation (`spec/x_wing_spec.dart`, and the two backends + `x_wing_algo_test.dart` back to trunk). A functionally identical consolidation of the same block is already in flight for these backends; landing both means resolving the same refactor twice.
- `EVP_DigestVerify` returning `< 0` now throws `StateError` instead of reading as `false` — a backend failure was still indistinguishable from a forged signature.
- Both ML-DSA-65 backends check their generated key pair's lengths, matching what ML-KEM-768 already does.
- Pinned the pure-Dart backend's "never throws" verify contract against right-length garbage (the FFI backend already had this; the pure-Dart cases were all rejected by the length gate first).
- `known`'s dartdoc cites the tripwire test rather than tooling that doesn't exist.
- Every algorithm-level length check now throws `ArgumentError`; `AtSigningException` stays at the `MlDsa65KeyPair` boundary where it's documented API.
- Restored trunk's `lookupFunction` wrapping in `ml_dsa_65_ffi.dart` — ~30 lines of reflow were burying the real edits.

**- How I did it**

See commit & diffs. One commit per item, so any of them can be dropped — the `ArgumentError` one in particular is a judgement call and moves four test assertions with it.

**- How to verify it**

Tests pass — at_chops 235/235 (including `--tags ffi` against libcrypto 3.6), at_auth 142/142, `dart analyze --fatal-warnings` clean on both.

No new entry — amends the in-progress at_chops 3.5.0 entries #2115 already added.


srieteja (approved):
Thank You @gkc !!

Issue #1898, opened 2026-04-13
deprecate at_backup_key_flutter / at_common_flutter / at_theme_flutter / at_sync_ui_flutter


Xlin123:
These packages either contain no benefitial code or are being removed in favour of a snippet in at_client_flutter/example/lib/snippets.

Xlin123:
at_backup_key_flutter and at_common_flutter have been deprecated, the other packages are also on their way out.

PR #2134, opened 2026-08-10
chore(at_sync_ui_flutter): move at_client_mobile dependency

Xlin123:
## What

- move at_client_mobile dependency so that I can publish at_sync_ui_flutter 1.2.0
- #1898 

## Skills impact

### DIFF
--- a/packages/at_sync_ui_flutter/pubspec.yaml
+++ b/packages/at_sync_ui_flutter/pubspec.yaml
@@ -12,12 +12,12 @@ environment:
 
 dependencies:
   at_client: ^3.7.0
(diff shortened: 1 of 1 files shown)


packages/at_sync_ui_flutter/pubspec.yaml
```diff
@@ -12,12 +12,12 @@ environment:
 
 dependencies:
   at_client: ^3.7.0
+  at_client_mobile: ^3.3.0
```
Xlin123, line 15:
```suggestion
  at_client_mobile: ^3.5.0
```

riic10 (approved):

PR #2139, opened 2026-08-12
feat(at_commons): the _apsk value reaches the wire, bare or array

gkc:
**- What I did**

- `EnrollVerbBuilder` gains `apsk`. `EnrollParams.apsk` shipped in 5.14.0 with no route to the wire, so nothing could send the value the atServer stores verbatim and publishes at approval.
- Adds `EnrollParams.apskLegacy` + the matching builder field: the **bare** RSA `_apsk` string, published verbatim.
  - Every deployed `_apsk` consumer base64-decodes the value as an RSA key, so a JSON one fails their parse — fail-closed, but service-breaking for anything already running. A plain-legacy enrollment needs to publish the shape they expect, through the same verb every other enrollment uses.
  - Two fields rather than one field of two types: `apsk` is published, so widening its type would be source-breaking, and neither the wire type nor the atServer's handling has to be widened later.
- Corrects the `apsk` entry dartdoc — `status` is `retired`, not the `verifyOnly` 5.14.0 documented, and an entry carries a `kid` like every other key entry in the protocol. Documentation only; the atServer stores the value verbatim.
- Version 5.14.0 → 5.15.0 (5.14.0 is published).

**Owed on the atServer side**, not in this PR: write `apskLegacy` as-is — *not* JSON-encoded, since a quoted string is not what a bare-RSA parser reads — and refuse a request carrying both fields.

**- How I did it**

See commit & diffs.

**- How to verify it**

Tests pass. The two new pins assert raw literals against the **built command** rather than reading the builder's own fields — a field-read test would have passed the entire time `apsk` could not be sent. Removing the two `..apsk = apsk` / `..apskLegacy = apskLegacy` lines turns exactly those two red.

`EnrollVerbBuilder.apsk` threads the existing `EnrollParams.apsk` to the wire, and `EnrollParams.apskLegacy` carries the bare RSA `_apsk` string a plain-legacy enrollment publishes verbatim.

### DIFF
--- a/packages/at_auth/lib/src/at_auth_impl.dart
+++ b/packages/at_auth/lib/src/at_auth_impl.dart
@@ -188,7 +188,7 @@ class AtAuthImpl implements AtAuth {
       atOnboardingRequest.atKeys = await atOnboardingRequest.atKeysIo?.read(
         atOnboardingRequest.atSign,
       );
-    } catch (e, _) {
+    } catch (e) {
       _logger.info(
         'Failed to read keys for atSign: ${atOnboardingRequest.atSign} | Cause: $e',
       ); //swallow the error, we just want to know if keys exist or not
--- a/packages/at_chops/test/at_chops_compatibility_test.dart
+++ b/packages/at_chops/test/at_chops_compatibility_test.dart
@@ -551,7 +551,7 @@ void main() {
           DefaultSigningAlgo(null, signingInput.hashingAlgoType);
       try {
         atChops.sign(signingInput);
-      } catch (e, _) {
+      } catch (e) {
         assert(e is AtSigningException);
         expect(e.toString(),
             'Exception: encryption key pair not set for rsa signing algo');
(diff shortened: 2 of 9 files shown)


packages/at_auth/lib/src/at_auth_impl.dart
```diff
@@ -188,7 +188,7 @@ class AtAuthImpl implements AtAuth {
       atOnboardingRequest.atKeys = await atOnboardingRequest.atKeysIo?.read(
         atOnboardingRequest.atSign,
       );
-    } catch (e, _) {
+    } catch (e) {
```
gkc, line 191: linter has gotten stricter

gkc:
@cconstab the new form of the apkam public signing key is a json structure with an array ; but for a while the old form (bare string, an RSA public key base64-encoded) will continue; and rather than shoe-horning both into the same field which is messy and yucky and other things, decided to just add apskLegacy here. 

Another atServer change will come once this has been published

cconstab (approved):

PR #2149, opened 2026-08-13
feat(wasm_shakedown): add wasm regression tests and compile gates for at_chops

Xlin123:
## What

Phase 0 of WASM `AtClient` work: a CI ratchet that keeps
the five core packages from silently gaining browser-hostile imports. 

`dart compile wasm` **does not reject `dart:io`** , it ships a stub that throws on
first use, so a browser-hostile import compiles clean and fails in the browser
instead. `dart compile js` is blind to it too. No Dart web compiler gates `dart:io`,
so a build step cannot serve as the gate. This lands the structural gate *before* any
porting work, baselined against today's violations, so each later phase shrinks a
baseline instead of discovering a regression after the fact.

Implements R1 and R4 from `docs/projects/wasm/implementation-plan.md` §2, advancing
gates T0.1–T0.2. 

## How

`tools/wasm_shakedown/`, a workspace-only dev tool. One gate: a dependency-tree
walk, driven per package from `packages/*/test/wasm/dep_tree_test.dart`, which
catches a platform library reachable from a barrel. It resolves `package:` URIs
through `.dart_tool/package_config.json`, so it crosses into sibling and third-party
packages rather than stopping at the package's own sources.

## Tests revealed one gap:

- `at_chops/test/at_pqc_test.dart` imports `dart:ffi` on line 2 but carries no
  `@Tags(['ffi'])`, unlike the other five FFI suites, so `--exclude-tags ffi` never
  excluded it and it fails to *compile* under dart2wasm, which does reject `dart:ffi`
  outright. Pre-existing tagging gap, one line to fix.

## How to verify
In CI, `wasm_ratchet` [ran here.](https://github.com/atsign-foundation/at_client_sdk/actions/runs/32723536854/job/97419806557?pr=2149#step:8:11)

Manually:
1. `dart pub get`
2. `cd tools/wasm_shakedown && dart analyze && dart test` — no issues, 16 pass
3. `for p in at_utils at_chops at_lookup at_client at_auth; do (cd packages/$p && dart test test/wasm/dep_tree_test.dart); done` — 9, 5, 3, 6, 3 pass

## Skills impact


Xlin123:
## First CI run: findings

Run [31823498926](https://github.com/atsign-foundation/at_client_sdk/actions/runs/31823498926). All six
jobs report **pass**, which is misleading — the T2 step is `continue-on-error`, so its
failure is invisible in the check list. Reading the logs:

### 1. T2 executed zero tests on all five packages

Not "some suites fail" — nothing ran. Every suite failed at **load**, before any test body:

```
Generated wasm module '…/logging_test.dart.node_test.dart.wasm.wasm', and JS init file '….mjs'.
…/logging_test.dart.node_test.dart.js:22
  const module = await instantiate(instancePromise, {});
                       ^
TypeError: instantiate is not a function
Failed to load "test/logging_test.dart": Node exited before connecting to the test channel.
```

| Package     | T2 result             |
| ----------- | --------------------- |
| at_utils    | 0 passed, 3 failed    |
| at_lookup   | 0 passed, 6 failed    |
| at_client   | 0 passed, 3 failed    |
| at_auth     | 0 passed, 12 failed   |
| at_chops    | 0 passed, 24 failed   |

The generated CJS bootstrap calls `instantiate` from the ESM `.mjs` init file and gets
`undefined` — a runner/SDK output-contract mismatch, nothing to do with our code.

**It does not reproduce locally.** `dart test -p node -c dart2wasm test/logging_test.dart`
in at_utils loads and runs fine here: 1 pass / 1 fail, the fail being the
`stdErrLoggingHandler` one. Two environment variables differ and neither is pinned —
`setup-flutter-and-dart` takes `dart-channel: stable`, and Node comes from the runner image:

| | CI | Local |
| --- | --- | --- |
| Dart | 3.13.0 | 3.12.0 |
| Node | 22.23.2 | 24.18.0 |

I haven't bisected which one it is, so I can't yet say whether this is a Dart 3.13 dart2wasm
regression or a Node 22 ESM-interop difference.

### 2. `--exclude-tags ffi` does not actually exclude at_chops' ffi suites

One at_chops failure is real and distinct from the loader break — a genuine compile error:

```
test/at_pqc_test.dart:2:1: Error: 'dart:ffi' can't be imported when compiling to Wasm.
Failed to load "test/at_pqc_test.dart": Bad state: dart2wasm failed.
```

`test/at_pqc_test.dart` imports `dart:ffi` on line 2 but carries no `@Tags(['ffi'])`; the other
five ffi suites (`aes_gcm_ffi_test`, `ml_dsa_65_ffi_test`, `ml_kem_768_ffi_test`,
`x25519_ffi_test`, `x_wing_ffi_test`) all do. Pre-existing tagging gap in at_chops, not caused
by this branch — the fix is one added tag line. Worth noting `dart:ffi` *is* rejected outright by
dart2wasm, which is the contrast the whole PR rests on: `dart:io` compiles clean and `dart:ffi`
does not.

### 3. The structural gate is green, and it's the part that works

The dependency-tree ratchet passed on all five packages against CI's own
`.dart_tool/package_config.json` — so the baselines are portable and weren't encoding anything
about my local pub cache. That was the actual open risk for R1/R4 and it's answered.

## What I think we should do

**Drop T2 (R5) from this PR.** An allowed-to-fail step that fails 100% of the time is
indistinguishable from one correctly reporting real failures — which is exactly what verify
step 5 was written to watch for, and it happened on the very first run. As shipped it costs
~1.5–2.5 min × 5 jobs to tell us nothing, and the "flip `t2_required` per package as each goes
green" plan has nothing to flip: no package can go green while every suite fails to load. If we
don't need to baseline it, we shouldn't.

Land R1/R4 — the gate that demonstrably works — and bring T2 back once it can actually run,
with `dart-channel` and Node pinned in the `wasm_ratchet` job so a green→amber transition means
something about our code rather than about whatever the runner image shipped that week.

Two consequences if we do that:

- The description's "at_utils is 25 pass / 2 fail" is a **local** measurement on Dart 3.12 that
  CI cannot currently confirm. It should be labelled as such, or dropped.
- `at_chops`' `--exclude-tags ffi` and `at_client`'s three-suite scoping both go with T2, and the
  at_pqc tagging gap stops being this PR's problem (still worth a separate one-liner).

Smaller thing in the same spirit: at_chops' web barrel baselines to `expectedOffenders: {}`.
That isn't a baseline, it's an assertion that the barrel is clean — fine to keep, but it reads as
if there's debt recorded there when there isn't.

packages/at_chops/test/wasm/dep_tree_test.dart
```diff
@@ -0,0 +1,45 @@
...
+/// `cryptography` is deliberately absent from the blocked count. Under web
+/// resolution its configurable imports select the pure-Dart implementation rather
+/// than the `dart:html` Web Crypto path — a structural fact only, not evidence
+/// that it computes the right answers under WasmGC.
+void main() {
+  ratchetGroup(
+    'package:at_chops/at_chops.dart',
+    package: 'at_chops',
+    allowedOffenders: const {},
+    maxBlockedPackages: 2,
```
Xlin123, line 27: as we push wasm breaks, we'll update the rachetGroup to have 0 offenders, 0 blocked packages.

tools/wasm_shakedown/README.md
```diff
@@ -0,0 +1,125 @@
+# wasm_shakedown
+
+Dev-only tooling that keeps at_chops compiling to wasm and stops it gaining
+browser-hostile imports. Not published; a workspace member used by
```
cpswan, line 4: Should we consider publishing this as a useful community resource?

Did we check there's not a package that does this already on pub.dev?

cpswan (approved):

tools/wasm_shakedown/README.md
Xlin123: ooo, will check now actually. good shout.

Xlin123: https://pub.dev/packages/wasm_check exists but currently is doing the shakedown from the pubspec.lock.

`wasm_check` is slightly different to what i wanted to accomplish with this tool. 

Maybe I'll send a message to the maintainer to see if we can improve their tool / add the functionality that we need.

Xlin123:
Thanks @cpswan 🚀

PR #2162, opened 2026-08-19
chore(wasm): remove dependencies where they aren't needed

Xlin123:
**- What I did**

Phase 1 of the WASM port: remove dependencies where they aren't needed.

- at_server_status: **remove `dart:io`** - it was imported for five `HttpStatus` constants, which `at_status.dart` now spells as integers. at_server_status comes off the WASM gate's blocked list, so `.github/wasm_gates.yaml` ratchets `max_blocked_packages` from 4 to 3.
- at_chops: **move `dart_periphery` to `dev_dependencies`** - nothing in `lib` uses it; only the zariot example does.
- docs: the wasm implementation plan and roadmap record Phase 1 as it stands, with #2163 and #2164 closed as superseded - D-15's `lookUps:` covers S1 and S2 and D-12's storage bundle covers S3.

This PR no longer changes at_client: trunk had already deleted `sync_isolate_manager.dart` and records it under 3.15.0-rc1.

**- How I did it**

See commits

**- How to verify it**

Tests pass. The WASM gates pass with at_auth at 3/3 blocked, and putting the `dart:io` import back in `at_status.dart` fails the ratchet at 4/3.

**- Description for the changelog**

- at_chops 3.7.0: `dart_periphery` is a dev dependency
- at_server_status 1.1.2-rc2: no `dart:io` import

**Skills impact**

- [x] No public API change - skill unaffected

PR #2163, opened 2026-08-19
feat(at_client): plumbing atlookup factories to the top of at_client

Xlin123:
## What

solution is tbd, I'm not really sure how i'm feeling on this one

## How to verify

1. 
2. 

## Skills impact

### DIFF
--- a/packages/at_client/CHANGELOG.md
+++ b/packages/at_client/CHANGELOG.md
@@ -1,16 +1,35 @@
 ## 3.15.0
-- feat: `RemoteSecondary` and `AtClientImpl.create` accept at_lookup's
-  `secureSocketFactory`, `socketListenerFactory` and
-  `outboundConnectionFactory`, and forward them to the `AtLookupImpl` they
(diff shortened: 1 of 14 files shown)


gkc:
Superseded, so closing.

Trunk makes the transport injectable a level above the socket factories: at_lookup declares `AtLookUpFactory`, and at_client's entry points take it as `lookUps:` (D-15 in `docs/projects/wasm/decisions.md`).

- S1: `RemoteSecondary` takes `AtLookUpFactory? lookUps` and builds its connection from it, falling back to `defaultLookUps(preference)`. An app that wants particular socket factories builds its own `AtLookupImpl` inside the factory.
- S2: `NotificationServiceImpl.create` takes `lookUps`, and the monitor's connection is built from it.
- The at_lookup uptake: at_client on trunk already depends on `at_lookup: ^3.7.0-rc2`.

#2162 below this still has changes trunk doesn't, so that one stays open.

PR #2164, opened 2026-08-19
feat(at_client): plumb the AtSyncQueue seam to AtClientImpl.create

Xlin123:
## How to verify

1. 
2. 

## Skills impact


gkc:
Superseded, so closing.

This plumbed a `syncQueue` through `AtClientImpl.create` to `LocalSecondary` (S3). Trunk now gets there through the client storage bundle instead: `AtClientImpl.create` takes an `AtClientStorage` (`storage:`), which owns both the keystore and the sync queue, and hands `storage.syncQueue` to `LocalSecondary` - see `at_client_impl.dart` and `test/storage/at_client_storage_wiring_test.dart`. Hive, SQLite and in-memory implementations ship.

D-12 in `docs/projects/wasm/decisions.md` supersedes S3 and rules that the queue doesn't get a route of its own, so a separate `syncQueue` parameter on `create` is the shape it rules out.

PR #2169, opened 2026-08-20
feat(at_chops): RFC 9180 HPKE seals, ML-DSA-65 PKAM, KEM seeds

gkc:
Second package of the post-quantum D1 release train, carved from `gkc-pq-d1-spike`. at_chops only; nothing else in the workspace changes.

**- What I did**

- `pqSeal` now does real RFC 9180 HPKE Base mode rather than a shape borrowed from it. Version `0x02` is the X-Wing suite (KEM `0x647A`, HKDF-SHA256, ChaCha20-Poly1305), `0x03` is pure ML-KEM-1024 (KEM `0x0042`, HKDF-SHA384, AES-256-GCM), and the caller picks which to emit.
- **BREAKING**: the old `0x01` seal is removed. It shared its KEM with `0x02`, so it was duplication rather than algorithm diversity, and no released build writes durable records sealed that way.
- **BREAKING**: `pqSeal` and `pqOpen` take `info` as required. Omitting it derived the same key schedule as an empty one, so two protocols that each said nothing shared a binding.
- New algorithms: `MlKem1024PureDartAlgo`, `ChaCha20Poly1305Algo`, `HkdfSha384`/`HmacSha384`, and separate `extract`/`expand` on `HkdfSha256`.
- `AtKemAlgorithm.newSeed` and `keyPairFromSeed`, with `KemSeedMixin` as the one implementation — a KEM key now stores and recovers the same way on every backend. Adding members to a public interface is what makes this a minor.
- `PkamMlDsa65SigningAlgo`, and `AtChopsImpl`'s pkam dispatch honours `signingAlgoType: mldsa65`. It signed RSA regardless of the requested algorithm before.
- `SigningAlgoType.strongestFirst` and `strongestOf` — the order a verifier uses to pick which of several signatures on one envelope to check.
- `ArgonHashParams.salt`: Argon2id takes a real salt, having passed the passphrase's own code units as the nonce. The old path stays as a documented fallback for key files already written. Adds `ArgonHashParams.owaspMinimum`.
- `pqSeal`/`pqOpen` wrap the KEM's `ArgumentError`, so a wrong-length key or a malformed envelope arrives as the documented exception type.
- Conformance vectors nobody here produced: NIST ACVP for ML-DSA-65, RFC 7748 for X25519, and the IETF HPKE working group's published vectors for both HPKE suites and for X-Wing.

**- How I did it**

See commit & diffs

**- How to verify it**

Tests pass

RFC 9180 HPKE seal versions `0x02`/`0x03`, ML-DSA-65 PKAM signing, a portable KEM seed contract, and a real Argon2id salt.


packages/at_chops/pubspec.yaml
```diff
@@ -1,6 +1,6 @@
 name: at_chops
 description: Package for at_protocol cryptographic and hashing operations
-version: 3.5.0
+version: 3.6.0
```
Xlin123, line 3: the PR description & changelog both define breaking changes, should this be 4.0.0 or is it clean?

packages/at_chops/lib/src/algorithm/encryption/pq_hpke.dart
```diff
@@ -1,48 +1,91 @@
...
-/// Reuses the RFC 9180 HPKE construction shape:
-///   - KEM      = caller-supplied [AtKemAlgorithm] (intended: X-Wing)
-///   - KDF      = HKDF-SHA256 ([HkdfSha256])
-///   - AEAD     = AES-256-GCM ([AesGcm256EncryptionAlgo])
+/// Raising this is a **fleet-wide** decision, not a local one: every reader has
+/// to understand the new version before any writer emits it, and a sealer has
+/// no way to discover what its recipients can open. Prefer passing
+/// `pqSeal(version:)` from a caller that knows — the key package's advertised
+/// suites, say — over changing the default.
+const int pqSealDefaultVersion = 0x02;
```
Xlin123, line 38: I'm in agreement to bump this to 0x02 / 0x03, but can we tighten the dependency on at_client in this PR?

right now it's ^3.3.0, which means we can potentially emit 0x01 - 0x03, and the builds with newer at_chops will reject the 0x01

packages/at_chops/lib/src/algorithm/encryption/pq_hpke.dart
```diff
@@ -71,33 +114,75 @@ class PqOpenException implements Exception {
...
-
-  // body = gcmCipherText || tag(16), per AesGcm256EncryptionAlgo's wire format.
-  final Uint8List body = await AesGcm256EncryptionAlgo(_aesKey(dk.key)).encrypt(
-    plaintext,
-    iv: InitialisationVector(dk.nonce),
-    aad: aad ?? const <int>[],
-  );
-
-  // envelope: ver(1) || ctLen(2,BE) || kemCt || gcmCipherText || tag(16)
+  final _SealVersion? row = _versions[version];
```
Xlin123, line 160: claude reasoned: 

pq_hpke.dart:150 — no check that the caller's kem
     matches row.suite.kemId. Sealing ML-KEM-1024     
     under version 0x02 (whose suite_id claims KEM    
     0x647A) round-trips only against a peer making   
     the same mistake, contradicting the documented   
     "version byte names the whole suite" invariant.  
     One throw closes it.

claude proposed: adding kemId to AtKemAlgorithms since theres already breaking points in the changelog.

packages/at_chops/CHANGELOG.md
```diff
@@ -1,3 +1,240 @@
...
+  derandomised encapsulation reproduces the published ciphertext. Third-party
+  bytes, mirrored by Go's standard library.
+- feat: **RFC 9180 HPKE Base mode as `pqSeal` version `0x02`** — the real
+  thing, not a shape borrowed from it. Suite: KEM `0x647A` (X-Wing /
+  MLKEM768-X25519), KDF `0x0001` (HKDF-SHA256), AEAD `0x0003`
+  (ChaCha20-Poly1305). The key schedule is RFC 9180 section 5.1 verbatim, with
+  `LabeledExtract`/`LabeledExpand` and the `suite_id` inside every label, and
+  it reproduces the IETF HPKE working group's published `key`, `base_nonce` and
+  `exporter_secret` for that suite plus all 10 of its published encryptions —
+  bytes nobody here produced. Version `0x01` (`atPQv1-base`) is unchanged and
```
Xlin123: >  Version `0x01` (`atPQv1-base`) is unchanged and  stays the default;

is this still true?

packages/at_chops/lib/src/algorithm/encryption/pq_hpke.dart
```diff
@@ -71,33 +114,75 @@ class PqOpenException implements Exception {
...
+        'cannot seal at version 0x${version.toRadixString(16)} — this build '
+        'supports ${pqSealSupportedVersions.map((v) => '0x${v.toRadixString(16)}').join(', ')}');
+  }
+
+  // Encapsulation rejects a wrong-length public key with an ArgumentError.
+  // Mapped here for the same reason pqOpen maps decapsulate's: the documented
+  // contract is that caller misuse arrives as PqSealException, and a raw
+  // ArgumentError escaping leaves a caller who catches the documented type
+  // with an uncaught error. A recipient's advertised key is peer-supplied on
+  // every path that reaches here, so this is reachable input, not a typo.
```
Xlin123:
```suggestion
```

duplicated comment block

Xlin123 (changes requested):
some changes to consider, and additionally the following (which I'm assuming will be covered in later PRs to come).

- new HashParams field won't be persisted in the `AtKeysPassphraseEnvelope` since it's not wired.
- iirc, we're currently hard coding the hashing algo for pkam as sha256 so mldsa won't roundtrip properly

some "low priority" from claude:
1. at_chops_impl.dart:296 — mldsa65 verification now
     silently falls back to the PKAM keypair's public 
     key when verificationInput.publicKey is null,    
     returning false instead of the old               
     AtSigningException. (The rest of that change is a
     real fix — the old code stored an unawaited      
     Future<bool> in a bool result.)                  
2. pq_hpke.dart:186 — ctLen written as two bytes    
     with no bound check; an external KEM with        
     >65535-byte ciphertext truncates silently.       
3. x_wing_core.dart:53 — combine sizes the buffer   
     from actual lengths but writes at hardcoded      
     32-byte offsets; a non-32-byte component yields a
     silently wrong shared secret. Pre-existing, but  
     now shared by both backends.

packages/at_chops/pubspec.yaml
gkc: Both breaks are real. We're publishing as a minor because there's nothing in existence for either of them to break.

There are exactly two. Callers now have to pass `info` to `pqSeal`/`pqOpen`, and anyone who writes their own KEM has to add two new methods. I went and counted: there are two callers in the whole repo and both already pass `info`, and every KEM implementation lives inside at_chops itself. Nothing outside this repository does either.

That reasoning was settled when the question first came up, but it never made it into the changelog — which explains why this is more than a patch and never says why it isn't a major. I'll add a sentence, since you won't be the last person to ask.

packages/at_chops/lib/src/algorithm/encryption/pq_hpke.dart
gkc: Three parts to this.

**It's already done.** The at_client branch asks for `at_chops: ^3.6.0`, and its two sealing calls now name the version explicitly instead of taking whatever at_chops happens to default to.

**It can't go in this PR.** This one is at_chops only — 44 files, all under `packages/at_chops`. Each package in the release sequence gets its own PR, and the check is that the carved-out package matches the source branch exactly, so adding at_client's file would fail it. at_client also can't ask for `^3.6.0` and publish until at_chops 3.6.0 is actually on pub.dev, which is what this PR is for.

**The range is narrower than 0x01–0x03.** Older at_chops has no version option at all — it always writes 0x01. Newer at_chops writes 0x02 unless told otherwise. So a published build can only ever produce 0x01 or 0x02. Writing 0x03 needs an argument that no published build passes.

One part no dependency change can fix, which is worth writing down: at_client 3.14.0 is already on pub.dev asking for "at_chops 3.3.0 or newer", and pub always picks the newest. The day 3.6.0 publishes, that same released at_client starts writing 0x02 without anyone changing a line, and two copies of it installed at different times can't read each other. These are the short-lived pairwise envelopes, they expire after seven days, and that whole area is marked experimental — so it isn't a promise we've broken. I'll record it alongside the note about retiring 0x01 rather than leave it undocumented.

packages/at_chops/lib/src/algorithm/encryption/pq_hpke.dart
gkc: Agreed on the problem, and it's slightly worse than described — you don't have to pass a version to hit it. `pqSeal` is public, `MlKem768PureDartAlgo` is public, and the default version is 0x02, so the wrong pairing happens just by calling the function the ordinary way.

I'd rather not put `kemId` on `AtKemAlgorithm`, for a reason that isn't obvious: two of our KEMs (the ML-KEM-768 pair) have no matching suite at all, and no number exists for them anywhere in the tree. They'd have to make one up or throw from a getter on a public interface — and it would be the third change to that interface in one release.

The alternative is smaller. The suite table already records the official numbers for the KEM, the key derivation and the cipher. The same published table has one more column we didn't copy: how long the KEM's output is — 1120 for X-Wing, 1568 for ML-KEM-1024. Add that, and have `pqSeal` compare it against what the KEM actually produced. Same protection, no change to any public interface, and because it checks the real bytes rather than what a KEM says about itself, it also covers a KEM we've never seen.

packages/at_chops/CHANGELOG.md
gkc: No. That was written when 0x02 was added, before 0x01 was removed — and because both happened under the same unreleased heading, the release notes now contradict themselves. Fixing it.

Three more in the same heading you didn't reach: two entries announce a specification document and a set of test vectors for 0x01 that this same release deletes or repurposes, and both point at a path under `docs/` that isn't in the published package, so anyone reading it on pub.dev follows a dead link. I'll clean all four up together.

packages/at_chops/lib/src/algorithm/encryption/pq_hpke.dart
gkc: Six lines, pasted twice. Your suggestion spans the whole block so it applies cleanly. Thanks.

gkc:
Thanks — good list. I've replied to the five inline comments in their own threads. Here are the five from the review body.

---

> new HashParams field won't be persisted in the `AtKeysPassphraseEnvelope` since it's not wired.

Correct, and here's why it looks odd: the change that adds the salt touches two packages, and only the at_chops half is in this PR. The at_auth half — writing the salt into the file and reading it back out — is already written on the source branch and lands with the at_auth PR, which comes after this one in the sequence.

Until then the salt is simply unused, so nothing changes for files already written. One gap I'll close in that PR: the file records the salt and the three cost settings, but not the key length, so a caller who changed that would write a file it couldn't read back.

---

> iirc, we're currently hard coding the hashing algo for pkam as sha256 so mldsa won't roundtrip properly

I checked both ends and it does round-trip.

It isn't hard-coded — it's a setting that defaults to sha256 — but the real point is that nothing on the ML-DSA path reads it. ML-DSA signs the message directly, so the new signing class takes no hashing option at all, and the atServer skips the field entirely once it sees `mldsa65`: it goes straight to ML-DSA verification before the hashing setting is ever used.

So the field still travels saying "sha256" and is ignored at both ends. Untidy, harmless.

---

> at_chops_impl.dart:296 — mldsa65 verification now silently falls back to the PKAM keypair's public key when verificationInput.publicKey is null, returning false instead of the old AtSigningException. (The rest of that change is a real fix — the old code stored an unawaited Future<bool> in a bool result.)

The change in behaviour is real, but it replaces something worse than it looks. The old class did its work asynchronously, which meant two things: the error it threw wasn't something a caller could catch — it escaped as an unhandled failure — and its true/false answer was being stored without being waited for, so **every** ML-DSA verification came back looking like a pass. The new class is synchronous, so both the error and the answer actually reach the caller.

The fallback itself is what the RSA path has always done — use the stored key when the caller doesn't supply one — so this makes the two consistent rather than making ML-DSA the loose one.

---

> pq_hpke.dart:186 — ctLen written as two bytes with no bound check; an external KEM with >65535-byte ciphertext truncates silently.

Real, and nothing can reach it today — the longest output any of our KEMs produces is 1568 bytes and the field holds 65535. It's also not new here; trunk has the same two lines. But the function accepts whatever KEM a caller hands it, so I'll add the check with the length named in the message. Four lines, and much cheaper than someone debugging it years from now from the receiving end.

---

> x_wing_core.dart:53 — combine sizes the buffer from actual lengths but writes at hardcoded 32-byte offsets; a non-32-byte component yields a silently wrong shared secret. Pre-existing, but now shared by both backends.

Half right, and I ran it rather than reasoned about it: a *short* input throws rather than passing silently. Only an over-long one is silently wrong.

As you say, it predates this PR — both copies on trunk have the same shape, so pulling them into one file kept the risk rather than creating it. The fix I'd want is rejecting any wrong-length input up front rather than tracking offsets, and I'd rather do that on its own than fold it in here.

---

**Changing in this PR:** the duplicated comment, the four stale changelog entries, the KEM/version check (as a length comparison rather than a new interface member), and the ciphertext-length guard.

**Already handled, coming with their own PRs:** the at_client dependency, and the passphrase salt wiring.

**Not changing:** the version number, the PKAM hashing field, and the ML-DSA verification fallback — reasoning above and in the inline threads.

gkc:
Pushed as `26a3093`. One thing came out differently from what I said above, so flagging it rather than leaving it in the diff to be spotted.

**The ciphertext-length guard isn't a separate check.** I said I'd add a bounds test on the two-byte length field. Written that way it was unreachable — the KEM/version check immediately above it already forces the length to exactly 1120 or 1568, so an `if (length > 0xffff)` below it can never fire. Dead code pretending to be a safeguard is worse than no safeguard, so instead the limit is asserted on `HpkeSuite` itself: a suite whose KEM wouldn't fit the field now fails to **compile**. That covers the case that can actually happen — someone adding a row to the table — rather than the one that can't.

**The framing KAT changed shape.** It used to pin the whole envelope as a hex literal with an 8-byte fake ciphertext. The new check refuses that, and a realistic 1120-byte one makes the literal about 2,300 characters, which pins the same bytes while showing a reader nothing. It now asserts the header directly — version byte, the big-endian length, the ciphertext slice, the total length — plus a sha256 of the whole envelope, so drift anywhere in it still fails.

**On your `kemId` suggestion,** one more argument for comparing lengths that I didn't make earlier: `kemId` would check what a KEM *says about itself*, and the length check reads what it actually produced. A KEM we've never seen gets the protection either way, but only the second one catches a backend whose declaration and behaviour disagree.

Verified before pushing: at_chops 425 tests pass and analyze is clean; at_client analyze is clean and its eight secret-sharing suites pass 141 with no seal refusals, so the real callers pair correctly. Both new tests were mutation-checked — disabling only the new branch turns both red naming `PqSealException`, and the KAT stays green.

Xlin123 (approved):
awesome thanks @gkc

gkc:
Thank you @Xlin123

PR #2179, opened 2026-08-22
feat(at_auth)!: 4.0.0-rc1 - pqc project and export split for wasm friendliness

gkc:
**- What I did**

at_auth 4.0.0-rc1. An rc, so consumers adopt it deliberately; other packages
here are updated only enough to keep compiling.

It is 92 files, and most of them do not need your attention. Where it does:

**Read these — API or behaviour actually changes**

| File | What to judge |
|---|---|
| `lib/at_auth_io.dart` | The barrel split. `at_auth.dart` no longer reaches `dart:io`; a `dart:io` consumer adds one import. Nothing left the package. |
| `lib/src/enroll/models/at_enrollment_request.dart` | `signingAlgo` is now **required** on both constructors. Deliberately not defaulted — an app enrolling over OTP always got RSA-2048 and could not ask otherwise. |
| `lib/src/keys/at_keys.dart` | +989/−57. The typed keyfile: enrollments own their key material, keyed by role and algorithm. The biggest single change in the PR. |
| `lib/src/enroll/apsk_advertisement.dart` | The `_apsk` record's shape. One active `rsa2048` key still spells as the bare public-key string; anything else as a JSON array. |
| `lib/src/auth/at_authenticator.dart` | New seam: at_lookup gets an authenticator rather than loose credential fields. |
| `lib/src/at_auth_impl.dart` | `AtOnboardingRequest.atKeysIo` no longer defaults to `FileAtKeysIo()`, and three "remove in v4" members are gone. |

**Skim or skip**

- `lib/src/enroll/at_enrollment_impl.dart` is **+65/−527** — split into
  `enrollment_submitter/approver/handshake/updater.dart`. Those four are new
  files holding relocated code; judge the split, not the logic.
- ~20 new test files, and the 807-line CHANGELOG.
- `probe_*.dart`, `socket_probe_io.dart`, `registrar_io_client.dart` —
  conditional-import plumbing that exists only to keep `dart:io` out of the core.

**The two files that are the review surface for format changes**

`test/at_keys_golden_test.dart` pins the at-rest keyfile against hand-written
goldens; `test/wire_literal_pins_test.dart` pins the verb commands. Both are
raw-literal pins, so an intended format change edits the pin in the same commit
— reading those diffs tells you what moved on disk and on the wire without
reading the writers.

**Authentication key vs advertised (`_apsk`) signing key**

Different audiences, hence separate axes. The APKAM key proves possession to the
atSign's **own atServer**, and only that atServer verifies it. What **peers**
verify — signed envelopes, key packages — is whatever `_apsk` advertises.

So an enrollment owns a signing key from birth and `_apsk` advertises that in
preference to the APKAM key. That is what lets credentials move to ML-DSA while
the fleet reads the advertisement unchanged: the operator upgrades their own
atServer, the fleet is not asked to.

**- How I did it**

See commit & diffs, and the inline comments I have left on the files above.

**- How to verify it**

Tests pass.


packages/at_auth/lib/at_auth_io.dart
gkc, line 1: **New public barrel — the consumer-visible half of this release.**

`at_auth.dart` no longer reaches `dart:io`. Everything needing a filesystem, a raw socket or the `dart:io` HTTP stack is exported from here instead. Nothing left the package — `FileAtKeysIo` is still at_auth's — so a `dart:io` consumer adds one import and changes nothing else.

Worth checking: that the split is where you would put it, and that nothing in the core barrel still transitively pulls `dart:io`. `test/wasm_barrel_test.dart` asserts the second.

packages/at_auth/lib/src/enroll/models/at_enrollment_request.dart
gkc, line 1: **`signingAlgo` is required on both constructors, and that is the deliberate part.**

An app enrolling over OTP always got an RSA-2048 APKAM keypair and had no way to ask otherwise. On an atSign whose deployment had moved to post-quantum, every install therefore created an RSA-authenticating enrollment which the client retrofitted away on its first start — a discarded enrollment per install, and an RSA credential live for the atServer's grace window.

It is required rather than defaulted so the compiler enumerates every call site and each states what it means. A default is what let a caller inherit an algorithm it never chose, which is the state that produced the defect.

Both constructors also now forward `advertisedSigningKey`, which the base class has always declared and neither passed on.

packages/at_auth/lib/src/keys/at_keys.dart
gkc, line 1: **The largest single change in the PR (+989/−57), and the at-rest format contract.**

The keyfile becomes typed: enrollments own their key material, addressed by role (`privateAuthentication`, `privateSigning`, …) and algorithm, rather than by the flat `apkamPublicKey`/`apkamPrivateKey` fields.

The flat fields do not go away — a legacy keyfile still reads, and a retrofitted one keeps the capped legacy enrollment's RSA credentials there while the typed section carries the live enrollment's. `authenticationFor(enrollmentId)` is the one place that decides which of the two a caller gets; reading the flat fields for an enrollment that has typed material of its own signs the PKAM challenge as the wrong principal.

If you read one thing here, read `authenticationFor` and `resolveAuthenticatingEnrollment`. The latter deliberately **throws** rather than choosing when a keyfile holds several live enrollments.

packages/at_auth/lib/src/enroll/apsk_advertisement.dart
gkc, line 1: **The `_apsk` record's shape — this is a wire format other atSigns read.**

A single active `rsa2048` key spells as the **bare public-key string**, which is what every deployed peer base64-decodes as an RSA key. Anything else — a second key, or a non-rsa2048 one — spells as a JSON array, which a deployed peer cannot parse.

That asymmetry is the rollout's whole mechanism rather than an implementation detail: it is what lets an enrollment's *authentication* key move to ML-DSA while its *advertised* key stays in the form the fleet can still read.

packages/at_auth/lib/src/auth/at_authenticator.dart
gkc, line 1: **New seam.** at_lookup is handed an authenticator rather than loose credential fields, so the decision about which credential shape authenticates a connection is made once, where the keys are.

The deprecated credential members on at_lookup stay — they are not removed here — so this is additive for anyone not using it.

packages/at_auth/lib/src/at_auth_impl.dart
gkc, line 1: **Two behavioural changes worth stopping on.**

1. `AtOnboardingRequest.atKeysIo` no longer defaults to `FileAtKeysIo()`. Onboarding mints key material and must persist it, and the core cannot assume a filesystem, so it throws naming what to set.

2. The `atOnboardingRequest.atKeys` branch is gone with the field. It never worked: `onboard()` overwrote whatever a caller set with the result of reading `atKeysIo` before anything read it, then threw "already onboarded" if that read returned keys — so the branch consuming it was unreachable.

packages/at_auth/lib/src/enroll/at_enrollment_impl.dart
gkc, line 1: **+65/−527 — this is a split, not a rewrite.**

The body moved into `enrollment_submitter.dart`, `enrollment_approver.dart`, `enrollment_handshake.dart` and `enrollment_updater.dart`. Those four are new files holding relocated code.

Judging the split is worth your time; re-reading the logic in the new files probably is not, except in `enrollment_submitter.dart`, which is where the enroll:request wire shape is composed.

packages/at_auth/test/at_keys_golden_test.dart
gkc, line 1: **Read this diff to see what changed on disk, without reading the writers.**

These are hand-written golden strings for the at-rest keyfile document, not round-trips through the code that writes it. A round-trip stays green when a field is renamed or a nesting reshaped, while orphaning every keyfile already written; a golden does not.

So an intended format change edits the golden in the same commit — and that edit is the review.

packages/at_auth/test/wire_literal_pins_test.dart
gkc, line 1: **The same, for the wire.** Exact-string pins over the enroll/otp verb commands at_auth emits.

The point is that the existing tests matched `startsWith('enroll:approve')` and would have stayed green through any change to the JSON that follows the prefix. These pin the whole command, so a diff here is a protocol change and nowhere else is.

packages/at_auth/test/at_keys_golden_test.dart
Xlin123, line 1: thanks I've been wanting to do something like this for a while

packages/at_auth/lib/at_auth_io.dart
Xlin123: i have made a rachet for this, i'll make a followup pr for some changes after this one.

Xlin123 (dismissed):
LGTM, i'll do a followup pr for wasm tests so go ahead and publish 4.0.0-rc1

srieteja:
Really cool stuff @gkc. A few things worth a look before merge, ordered by severity.

---

**1. `packages/at_auth/lib/src/keys/io/file_lock.dart:156` (and `:132`)**

*What:* `_release` and `_breakStale` both rename `lockPath` away to inspect the file's contents *before* deciding what to do with it, which vacates `lockPath` for the whole inspection window. A concurrent poller's `create(exclusive: true)` can succeed in that gap and become a second legitimate holder — and then the restore at `:162` renames the claimed file back onto `lockPath`, atomically replacing (and silently evicting) that new holder's lock. Two processes then believe they hold the lock, so the mutual exclusion this file exists to provide can break and concurrent writers can corrupt the keyfile.

*Recommendation:* rather than guarding the restore step (which just moves the window), identify "is this still my lock" without removing it from `lockPath` at all — compare inode identity via `stat`, or swap the rename dance for real OS advisory locking (`flock`/`fcntl`; worth checking whether Dart's `RandomAccessFile.lock()` covers the non-blocking poll shape you need here).

---

**2. `packages/at_auth/lib/src/keys/serialization/assurance.dart:22`**

*What:* `_reservedTopLevelKeys` doesn't include the legacy top-level `'keys'`, but `AtKeys.fromJson` strips it at parse time (`at_keys.dart:788-799`). So on the first flush of any pre-existing at_auth 3.3.0 keyfile, `_legacyJsonOf(existing)` carries `keys: []` while the candidate doesn't, and `_assertLegacyPreserved` (`:414`, called from `:297`) throws `map.legacy.keys is not preserved`. Those keyfiles read fine but can never be written — the first enrollment, retrofit or key-filing save fails and the mutation is lost.

*Recommendation:* add `'keys'` to `_reservedTopLevelKeys` in **both** `assurance.dart:22` and `AtKeys._reservedTopLevelKeys` (`at_keys.dart:79`) — they're identical today and the comment at `assurance.dart:18` asks that they stay that way, so changing only one trades this bug for the divergence that comment warns about. Alternatively, have the assurance tolerate an empty `keys` exactly as `fromJson` does. Either way it's worth adding a fixture test that round-trips a real 3.3.0 keyfile through `validateMapUpdate` — `at_keys_test.dart:140` covers `fromJson`/`toJson` only and never puts that document through a flush.

---

**3. `packages/at_auth/lib/src/at_auth_impl.dart:348`**

*What:* `onboard()` installs the PKAM authenticator before the server has assigned an enrollment id — correct at that instant, since step 3 is CRAM-authenticated — but it's never reinstalled once `enrollmentIdFromServer` comes back at `:359`. `enrollmentId` is bound by closure at install time, and `_pkam` writes that captured value straight to `PkamVerbBuilder..enrollmentlId` (`at_authenticator.dart:257`) with no fallback to `keys.enrollmentId`, so the argument passed to `pkamAuthenticate()` at `:379` is ignored and the activation PKAM goes out enrollment-less. `authenticate()` already gets this right via `enrollmentId ??= atAuthKeys.enrollmentId` at `:147`.

*Recommendation:* mirror `authenticate()` — reinstall the authenticator with the resolved id after `:359` and before the activation PKAM (`_installAuthenticator` is a plain setter overwrite, so re-calling is safe). A `enrollmentId ?? keys.enrollmentId` fallback inside `_pkam` would fix this site and any future caller that forgets, but it changes behaviour for every authenticator — probably the targeted reinstall for an rc, and the fallback as a follow-up if you like it.

---

**4. `packages/at_auth/lib/src/enroll/enrollment_submitter.dart:83` (`_handleFirstEnrollmentRequest`)**

*What:* the OTP path asserts `isPq && metadata?['keyPackage'] == null` before submitting (`:261`); this path has no equivalent check. Since `_buildMetadata` swallows any `metadataBuilder` exception and returns `null` (`:205`), a PQ first-enrollment whose builder throws silently activates the atSign with no encapsulation key — and per the doc at `:71`, `metadata.keyPackage` is only writable by the request that creates the enrollment record, so being the *first* enrollment there's no later path to add one. For an `rsa2048` first onboard it's quieter still: `_apskFor` at `:160` reads a null `keyPackage` as the legacy bare-apsk case, so the wire form looks like a deliberate legacy onboard rather than a failure.

Worth noting this isn't hypothetical now that `AtKeysIo.write` is create-only — a builder using the write-then-return idiom throws on the `InMemoryAtKeysIo` that `_buildMetadata` has already written to at `:201`.

*Recommendation:* add the same `isPq`/`keyPackage` guard here that `_handleAtEnrollmentRequest` already has, and consider letting a `metadataBuilder` failure propagate rather than converting it into silent permanent capability loss.

---

**5. `packages/at_auth/lib/src/enroll/enrollment_submitter.dart:317` (`_handleAtEnrollmentRequest`)**

*What:* `metadataBuilder` runs at `:246`, deliberately before the request so it can advertise a key package — which means everything it files resolves through `_containerFor` with a null enrollment id and lands in the atSign-scope container. After `enrollmentIdFromServer` is assigned at `:317`, only the flat APKAM keypair is re-filed under it; unlike `_handleSelfEnrollmentRequest` (`:550`) and `onboard()` (`at_auth_impl.dart:590`), this path never calls `adoptMaterials`. So whatever the builder filed (e.g. the PQ key package's private half) stays atSign-scoped, `keysForEnrollment(enrollmentIdFromServer)` never returns it — there's no atSign-scope fallback in `_activeAuthenticationMaterial` (`at_keys.dart:1143`) — and it's left behind when the enrollment is retired.

*Recommendation:* add the `adoptMaterials(constructionKeys.keys, enrollmentIdFromServer)` call here, matching the other two enrollment paths.

---

**6. `packages/at_auth/lib/src/keys/io/file_lock.dart:107`** (and the sibling `continue` at `:103`)

*What:* a non-contention `create(exclusive: true)` failure (`EACCES` on a read-only keys dir, `ENOSPC`) is indistinguishable from "held by someone" at `:73`. The subsequent `statSync()` on a file that was never created also throws, hits the bare `continue` at `:107`, and loops straight back — skipping both the deadline check at `:109` and the poll sleep at `:115`. Result is an unbounded, silent 100%-CPU spin instead of the documented timeout, with the real I/O error never surfaced.

*Recommendation:* either scope the initial catch to `EEXIST` and let genuine I/O errors propagate (note the errno differs on Windows), or — simpler and portable — route every non-acquire outcome through the deadline check and the sleep, keeping the last `FileSystemException` so the timeout can name it. Both `continue`s need it; fixing only `:107` leaves the spin reachable through `_breakStale` failing repeatedly for the same reason.

---

*Nice to have, not blocking:* the temp files created by `_release`/`_breakStale` (`.lock.releasing.*` / `.breaking.*`) aren't cleaned up if an exception fires mid-sequence — fine to leave for a follow-up.

gkc:
@srieteja thank you; there are some very good finds in here; I'm addressing each of them

srieteja (approved):

PR #2182, opened 2026-08-24
fix(at_commons): 5.17.0 - Metadata.fromJson preserves a null ttl/ttb/ttr

gkc:
**- What I did**

- `Metadata.fromJson` preserves a null `ttl`/`ttb`/`ttr` instead of reading it as `0`, so the `toJson`/`fromJson` round trip is lossless.
- `toJson` always writes the three, so an unset one goes out as `"ttl": null` and came back as `0` — leaving a reader unable to tell *"this request said nothing about it"* from an explicit 0. Those are different requests: `ttl:0` clears a record's expiry, `ttb:0` means available with no delay, `ttr:0` means do not cache.
- The caller this affects is the atServer's `update:json` handling, which reached that `0` for an axis the request never named, so a value-only json update dropped a record's expiry and stopped it being cached at the receiver. It now agrees with the metadata-fragment form of the same request, which has always left an unmentioned relative null.
- An explicit `0` still round trips as `0` — this is not "null and 0 become the same thing" — and a numeric string is still parsed. All three are pinned.
- **Version 5.17.0, not a major**: no API signature change (`Metadata.ttl` was already `int?`), the old behaviour was a lossy round trip rather than a contract, and the blast radius is one in-repo call site (`UpdateParams.fromJson`) plus at_server. A sweep of eleven sibling Dart repos found no other consumer of `Metadata.fromJson`. A major would have cost a floor raise in all 22 in-repo pubspecs that depend on at_commons for a change nothing else calls.

**- How I did it**

See commit & diffs.

Paired with atsign-foundation/at_server#2769, which fixes the same user-visible symptom server-side against a *published* at_commons. That change reads the same decoded map to the same answer once this lands, so the two agree rather than fight, and neither blocks the other.

**- How to verify it**

Tests pass.

- Mutation-proven: restoring the coercion reddens the round-trip test with `Expected: null, Actual: <0>`.
- at_commons 521/521, at_client 651 passed / 39 skipped, at_onboarding_cli 33/33. Every Dart workspace package analyzes clean under `--fatal-warnings`, and the `at_client` CI format gate is clean.

`Metadata.fromJson` preserves a null ttl/ttb/ttr instead of reading it as 0, so an unset value survives the json round trip and is no longer indistinguishable from an explicit 0.


gkc:
@cpswan belt-and-braces follow-up to the atServer PRs

cconstab (approved):

PR #2183, opened 2026-08-25
feat(wasm_shakedown): mature wasm_shakedown and verify at_auth 4.0.0-rc1 with it

Xlin123:
WASM gates move from two Dart files per package + a CI matrix line, to
`.github/wasm_gates.yaml` + `dart run wasm_shakedown`. **Adding a package is one
stanza.** Probe is generated at run time; both per-package files, all five dev
dependencies and `lib/ratchet.dart` are gone.

Introduced tests to CI for wasm_shakedown parsing and the actual shakedown.

No longer need two specifically named package tests to run the wasm_shakedown and will still continue if one package fails.

The majority of the lines are parsing & tests.

No published API changes, the five pubspecs only lose a `dev_dependency`, so no version
bumps. No new package joins the gates yet.


.github/wasm_gates.yaml
Xlin123, line 1: controls the gates in CI using the wasm_shakedown to perform a walk down .dart_tool/package_config.json

Xlin123:
also to note in this PR I've been looking at wasm_check as a tool, and it closely resembles the hand-rolled wasm_shakedown.

I've reached out to the maintainer about contributing and have matched our tool a bit closer so if we ever can make the switch it's a bit easier.

Xlin123:
https://github.com/atsign-foundation/at_client_sdk/actions/runs/32871390659/job/97880152184?pr=2183

srieteja (commented):
Reviewed `09fc94d28..a1bb3199b` (23 files). The design is right — the ratchet/control/probe split is the correct instinct, and the gate demonstrably works: injecting `dart:io` into an at_auth source fails the ratchet at 1/0 offenders, 5/4 blocked, while the compile probe stays green, exactly as the README says it should.

Two defects, both in the tool. **One worth fixing before merge.**

> **Independently reproduced.** A second agent (Gemini / Antigravity) re-ran both repros on a pristine checkout of `tools/wasm_shakedown` — no local overrides — and confirmed both findings *and* that both patches below actually fix them.

---

### 1. A mismatched barrel makes a gate vacuous — `config.dart:141`

`parse` validates the **stanza name** against the workspace package config, and `RatchetVerdict.packageKnown` re-checks it, for the reason stated in the code:

> Every check in this stanza is scoped to that name, so a name the package config does not know would read as a clean result

The **barrel** gets no equivalent check — `_barrel` (`:301`) only asserts it is a `package:` URI with a `/`, never comparing the authority to the stanza name. So the walk is rooted in one package while offenders are filtered by another, and nothing reconciles them.

```yaml
# stanza says at_client, barrel says at_chops — one copy-paste apart
at_client:
  ratchets:
    - barrel: package:at_chops/at_chops.dart
      allowed_offenders: []
      max_blocked_packages: 2
      min_files_walked: 500
  probe:
    - package:at_chops/at_chops.dart
```

```
  ratchet  package:at_chops/at_chops.dart — 582 files walked, 0/0 offenders, 2/2 blocked ok
  1 gated package, 1 ratchet, 0 controls, 1 compile — all green      # exit 0
```

Same stanza, correct barrel:

```
  ratchet  package:at_client/at_client.dart — 1256 files walked, 7/0 offenders, 8/2 blocked FAILED
  (at_client_impl, at_client_spec, remote_secondary, monitor, encryption_service,
   file_transfer_service, stream_notification_handler)
```

**7 offenders and 8 blocked packages reported as all green.** Every existing defence misses it:

| Defence | Why it doesn't fire |
|---|---|
| `packageKnown` | `at_client` *is* a workspace package — the name is fine, the barrel isn't |
| `min_files_walked: 500` | at_chops's graph is 582 files — deep, just the wrong tree |
| `nothingMissing` | at_chops's graph resolves cleanly |
| `controls` | optional in `_gate`; `config_test.dart` explicitly accepts "a gate with no controls at all" |
| compile `probe` | compiles at_chops, and passes |

Not hypothetical: the config header names at_utils/at_lookup/at_client as the next stanzas. Whoever ports at_client copies the at_chops stanza, changes the key, and misses one of four barrel lines.

**Fix** — thread the name into `_barrel`; all three call sites (`:176`, `:211`, `:263`) already have it in scope:

```dart
final authority = uri.substring('package:'.length, uri.indexOf('/'));
if (authority != package) {
  throw GateConfigException(_at(path, node,
      'barrel "$uri" belongs to $authority, not $package. Offenders are '
      'filtered by the stanza name, so a barrel from another package walks '
      'a graph this gate never examines and reports a clean result'));
}
```

Same rule as the stanza-name check, same reason. If a cross-package barrel is ever wanted, make it explicit rather than implicit.

---

### 2. Malformed YAML crashes instead of failing cleanly — `bin/wasm_shakedown.dart:59`

`GateConfig.parse` calls `loadYamlNode` (`config.dart:135`). `YamlException` is neither `GateConfigException` nor `StateError`, so main's handler misses it.

```bash
printf 'at_chops:\n  ratchets:\n   - barrel: "unclosed\n' > bad.yaml
dart run wasm_shakedown --config bad.yaml
```

| Bad input | Output | Exit |
|---|---|---|
| missing config file | `no gate config at … — this is the registry of gated packages …` | `2` |
| typo'd key | `bad.yaml:6: unknown key "min_files_walkd" in at_chops. Allowed here: …` | `2` |
| **YAML syntax error** | **`Unhandled exception:` + 24-frame stack trace** | **`255`** |

`cli_test.dart` already asserts `isNot(contains('Unhandled exception'))` for the sibling path, so the intended contract is explicit — this path just isn't wired to it.

**Fix** — `YamlException implements FormatException`, and `GateConfigException` does not, so the clauses don't overlap:

```dart
} on GateConfigException catch (e) {
  _die('$e');
} on FormatException catch (e) {   // YamlException: a syntax error in the config
  _die('$configPath: $e');
} on StateError catch (e) {
  _die(e.message);
}
```

---

### Nits

- `.github/workflows/at_libraries.yaml:195-196` — double blank line before the `WASM gates` step.
- `.github/wasm_gates.yaml:33` — the `# Tighten when convenient: add reaches_file` TODO is a one-liner today: `reaches_file: lib/src/algorithm/ffi/openssl_ffi_bindings.dart`.
- Consider making `controls` required in `_gate` rather than only asserted for the shipped config, so a new stanza can't skip the positive control.

### Checked and cleared

So these don't get re-raised: root-level `dart run wasm_shakedown` works despite the root package not depending on it; root `dart pub get` is fine with `flutter: sdk: flutter` members; `wasm_ratchet` inherits no `working-directory`, so the relative `--config` resolves; `_optionValue`'s index advance is correct; multiple probe barrels share one generated entry point with no race; `underCeiling`/`deepEnough` match the deleted `ratchetGroup` exactly; and the deleted per-package tests lose no coverage — the "conditional takes only the default branch" negative is now enforced indirectly, since a web walk down the io branch pulls in `socket_probe_io.dart` as an at_auth-owned offender against `allowed_offenders: []`.

srieteja (approved):

PR #2184, opened 2026-08-26
fix: pass passPhrase to FileAtKeysIo during onboarding

riic10:
## What

The onboard() path constructed FileAtKeysIo without forwarding passPhrase from AtOnboardingPreference, so `onboard -P` silently wrote an unencrypted atKeys file. The authenticate() path already passed it correctly; this aligns onboard() to match.

This is related to adding support for password-protected .atKeys files: 

atsign-foundation/at_activate_web#12

## How

Added passPhrase: atOnboardingPreference.passPhrase to the FileAtKeysIo constructor at line 195 of at_onboarding_service_impl.dart. When passPhrase is null (the default for callers not using password protection), FileAtKeysIo writes plaintext just like before.

## How to verify

Run at_activate onboard -P <password> ... and confirm the resulting .atKeys file is a JSON envelope with content, iv, and hashingAlgoType keys (not plaintext key material).

## Skills impact

- [] Public API changed — I have opened or will open an `at_client_skills` PR to reflect this


JeremyTubongbanua (approved):

JeremyTubongbanua:
A pub.dev release of at_onboarding_cli will need to be done post-merge into trunk

srieteja (approved):

PR #2193, opened 2026-08-31
docs(wasm): scope V1 to remote-only; correct the crypto claims

srieteja:
## What

- Scopes **V1 of the browser lane to remote-only** — no local store, no SQLite, no VFS, no `sqlite3.wasm`, no Hive. Local storage returns in V2 as a speed optimisation.
- Records rulings **D-12…D-19** as binding.
- **Corrects two crypto errors that are live on `trunk`** and currently mis-specify a task.
- Adds the **enterprise identity provider** concern (Entra/Okta), which had zero coverage in any document.
- Adds the **deployment contract** — the thing all three browser concerns depend on and no document owned.
- Adds gate families for **confidentiality, remote-only correctness, deployment and surface stability**; the existing key-material gates measured only wall-clock.

Why remote-only: the browser lane had two independently hard problems — a durable local database and a secure key store. Remote-only deletes the first outright, and the survivor is already designed. It also removes the two pieces of process-global state we do not own (`Hive`, `sqlite3`'s open-override).

## How

**`decisions.md`** — D-12…D-19; OQ-5/11/12 closed or constrained, OQ-13…15 opened; header range corrected to D-1..D-19.

| # | Ruling |
| --- | --- |
| D-12 | V1 is remote-only |
| D-13 | Remote-only is an **injected write-through `Secondary`**, not a nullable `localSecondary` |
| D-14 | Dart runs on the main thread |
| D-15 | Every JS/TS protocol operation returns a `Promise`; no sync escape hatch |
| D-16 | Redirect-based OIDC only; no popups |
| D-17 | No cross-origin-isolated VFS, and the VFS choice is not a benchmark decision |
| D-18 | No new process globals; one client per atSign, explicit |
| D-19 | JS facade authored once in TypeScript; `.js` and `.d.ts` both generated |

**`design.md`** — the crypto correction:

- `package:cryptography` does **not** reach Web Crypto via `dart:html`. It uses `dart:js_interop` with a **runtime** `window.isSecureContext` probe. The claim was inherited verbatim from the predecessor `plan.md`.
- The `dart:html` dependency is **`better_cryptography`** — which sits on the **default AES path**, selects its backend at compile time via `dart.library.html`, and has **no secure-context gate at all**. Outside a secure context the expected outcome is a throw on every encrypt, not a fallback.
- The two packages resolve by different mechanisms and fail differently, so they are now separate rows and may not be merged.
- Task **C1** is re-specified: two of its three questions are answerable from source. The Argon2id question is settled — `cryptography` 2.9.0 has no Argon2id override and Argon2id is absent from the WebCrypto spec, so it always resolves to `DartArgon2id`. The deferred Argon2id UX work stands.

**`js-api.md`** — new **§8a deployment contract** (secure context, `wasm-unsafe-eval`, `application/wasm`, and why COOP/COEP are deliberately absent); new **§5.0** Promise-only mandate; **§4/§8** switched to a single TypeScript-authored facade, including the ESM import-hoisting trap around the Node `globalThis.self` shim.

**`acceptance.md`** — new **§9a**: X-K1–4 (confidentiality), X-R1–5 (remote-only), X-D1–3 (deployment), X-S1–3 (surface stability), plus the bucket (a)/(b) triage rule as a standing requirement.

**`enterprise-identity.md`** — new. SCIM lifecycle mapping, the missing atSign-level disable (with `config:block` as the adjacent-but-wrong-direction primitive), the registrar asks, and six non-foreclosure constraints.

**`roadmap.md`, `implementation-plan.md`** — scope banners; document map updated to seven docs.

### Notable retractions

Earlier drafts were wrong on four points, all corrected here:

| Claimed | Actual |
| --- | --- |
| Transport needs another team | atServer speaks WebSocket natively; the WSS proxy is built; a TypeScript client drives the path end-to-end |
| `at_persistence_secondary_server` is another repo | It is in `at_server` — ours to schedule |
| `unrevoke`/`delete`/`fetch`/`listns` are server gaps | All nine enroll operations exist on the wire. These are **library-API** gaps |
| Non-extractable PKAM keys are reachable | Closed three ways; the third is an active refactor moving deliberately the other way |

## How to verify

Documentation only — no code changes, so verification is documentary.

1. `grep -rn "dart:html" docs/projects/wasm/` — every remaining mention attributes it to `better_cryptography`, never to `cryptography`.
2. `grep -rn "COOP\|COEP\|isSecureContext\|SharedArrayBuffer" docs/projects/wasm/` — returns hits in the `js-api.md` §8a deployment contract and D-17. Before this PR it returned none.
3. `decisions.md` — every open question has an owner, a resolution, or a named blocker. No unowned rows.
4. Internal links resolve: `for f in docs/projects/wasm/*.md; do grep -o '](\([a-z-]*\.md\))' "$f"; done` — all targets exist.
5. `roadmap.md` document map lists seven docs and cites D-1..D-19.

## Skills impact

🤖 Generated with [Claude Code](https://claude.com/claude-code)


srieteja:
Superseded by #2209, which reconciles this same content onto the storage-bundle ruling in #2207.

**This must not merge as-is.** #2207 introduces its own `D-12`; this PR introduces a different `D-12` and runs to D-19. The two merge textually clean, so trunk would silently end up with two `### D-12` headings. #2209 renumbers ours to D-13..D-20 and targets `gkc-wasm-docs-to-trunk` directly.

Marking draft until #2207 lands, then this closes.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

srieteja:
#2207 has merged to trunk. Superseded by #2209, which carries this content renumbered to D-13..D-20 so it composes with the D-12 that landed. Closing.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #2197, opened 2026-08-31
e2e enrollment teardown tears down other CI runs: observed failing, with the concurrency measured


gkc:
## What happens

`tests/at_end2end_test/test/enrollment_teardown.dart` operates on **every** enrollment on the shared `@ce2e1`–`@ce2e4` atSigns, with no filter for the run that created them:

- lines 21–32: fetches every `EnrollmentStatus.pending` enrollment and **denies** it
- lines 37–55: fetches every `EnrollmentStatus.approved` enrollment and **revokes** it with `force: true`

So when two CI runs overlap, whichever finishes its suite first denies and revokes the other's in-flight enrollments. The victim then fails wherever it next touches one.

## Observed firing, 2026-08-31

Until now this was a reasoned mechanism with no captured instance. Here is one.

**Run [33431448303](https://github.com/atsign-foundation/at_client_sdk/actions/runs/33431448303)** (branch `gkc-pq-d1-spike`, 19:35–19:54 UTC), job `end2end_test_14`:

```
06:08 +3 -1: A test to submit and approve an enrollment for @ce2e2 [E]
  ErrorCode: AT0032 - Exception: Illegal state exception :
  Failed to approve enrollment id: 314ad62e-df65-4249-9443-d451b557bd57.
  Cannot approve a denied enrollment.
    package:at_auth/src/enroll/enrollment_approver.dart 87:9   EnrollmentApprover.approve
    package:at_client/src/service/enrollment_service_impl.dart 153:22  EnrollmentServiceImpl.approve
    test/enrollment_setup.dart 132:11
```

The enrollment was **submitted at 19:43:24** and the approve **failed at 19:44:13** — 49 seconds later, already `denied`. Nothing in that run's own log denies it.

A concurrent `at_client_sdk` workflow run on PR #2182 ran **19:39–19:46**, straddling both timestamps. Its teardown denies every pending enrollment on the shared atSigns, which is exactly the state the approve then refused.

## The differential

| Run | Concurrent `at_client_sdk` runs | `end2end_test_14` |
| --- | ---: | --- |
| `bd1d6e242`, 19:35–19:54 | **4** | **FAIL** |
| `d75492f66`, 17:58–18:17 | 1 | pass |
| `02ac0e0f0`, 17:38–17:56 | 1 | pass |

[...]

PR #2200, opened 2026-09-04
fix(at_client_flutter): make keychain storage use `:` delimiter instead of `_`

JeremyTubongbanua:
## What

- Uses `:` as delimiter in keychain instead of `_`
- Updated changelog and version to 1.1.6 (originally 1.1.5)
- Migration flow implemented where if data is found in `_`, it is duplicated in `:`. Nothing is deleted in `_` for now.
- `deleteAllAtKeysData()` function deletes both `_` name and `:` name.
- Updated tests

## Why I made this change

- NoPorts Desktop keychain appears to be "wiped" by our users. It's not actually wiped, it just changed where to look for keys, and making this change will revert it back to its previous behaviour.
- I think the naming looks cleaner as well

## Skills impact


gkc (changes requested):
This is a breaking change for developers who have built against at_client_flutter with the `_`

Need to have a graceful migration built into the code. (e.g. look for `_` first, if found, migrate to `:`)

gkc:
@JeremyTubongbanua 
> Migration flow implemented where if data is found in _, it is duplicated in :. Nothing is deleted in _ for now.

OK ... when _would_ the old `_` form be deleted?

gkc:
PS the e2e check failed because of [this](https://github.com/atsign-foundation/at_client_sdk/issues/2197)

JeremyTubongbanua:
> OK ... when _would_ the old `_` form be deleted?

My thinking is deleting keys under `_` can be dangerous if we do it too soon.

`1.1.5 -> 1.1.6` -- just make it so that `:` works (if we find `_`, duplicate `_` data to `:`, otherwise, read `:`). Let's say an app that runs on `1.1.6` moves everything, deletes it, then goes back to a `1.1.5` app, suddenly all their keys are gone because we deleted everything under `_`. Something I'm now considering as I write this is what if a new `1.1.6` app writes to `:` but an older `1.1.5` app cannot find those keys, so maybe we should be writing to both, but reading from `:`... 

`1.2.0` -- we start deleting `_`. The same migration flow and fallback still exists: where apps will first look in `_`, if it's there, we duplicate over to `:`, but now we delete `_` data. 

`2.0.0` -- we remove all references of `_`. 

Let me know what you think @gkc !

JeremyTubongbanua:
I'm going to convert to draft and think about it a bit more

gkc:
Given that this only happens in app sandbox, we don't need to worry about there being simultaneously an older version and a newer version. I liked your proposed sequencing above with one modification for release sequencing
- 1.x.x:
```
  if (`:` isn't present && `_` is present) {
    copy from `_` to `:` // but leave `_` in place
  }
```
- 2.0.0:
```
  if (`:` isn't present && `_` is present) {
    copy from `_` to `:`
    remove `_`
  }
```

JeremyTubongbanua:
Note to self: on merge, publish at_client_flutter to pub.dev

gkc:
> Note to self: on merge, publish at_client_flutter to pub.dev

Everything is behind the pqc release train now. Can't publish except as a release candidate

gkc:
Merged trunk into this branch (f1eda34f9) to clear the conflicts with #2233, plus one follow-up commit.

Conflicts:
- `keychain_store.dart`: kept `_getImproperAtKeysStoreName`, dropped `EnrollmentStore` (trunk removed it)
- `keychain_storage_test.dart`: kept all the migration tests, dropped the `EnrollmentData` schema test (its API is gone on trunk)
- `CHANGELOG.md`: your entries now sit under `2.0.0-rc1`; dropped the `## 1.1.5` heading since 1.1.5 was never released

Follow-up (2e3d07143): this now ships in 2.0.0, so the CHANGELOG entry and the TODOs name 3.0.0 for removing the `_` store, and "pre-1.1.6" is now "pre-2.0.0". No behaviour change - `_` is still copied to `:` and left in place.

`flutter analyze`, `flutter test` and the format gate all pass locally.

gkc (approved):

cconstab (approved):

JeremyTubongbanua:
@Xlin123 @srieteja please take a look at this change for awareness

PR #2202, opened 2026-09-04
fix(test): give each enrollment a keypair of its own, and name the enrollment

gkc:
## What

Stops the at_client_sdk test packs failing once
[at_server#2787](https://github.com/atsign-foundation/at_server/pull/2787) merges
to trunk. That PR moves the flat PKAM credential into an enrollment called
`primary`, after which two things these fixtures do are refused: enrolling a
keypair another enrollment already holds, and authenticating with a bare
`pkam:` on an atSign onboarded through `enroll:request`. Measured against a
server carrying that work, three tests in
`tests/at_functional_test/test/enrollment_test.dart` fail.

The changes are compatible with the current atServer **and** with how it
behaves after at_server#2787, so this can merge in either order.

## How

Each `enroll:request` mints its own keypair; the hand-built client is given the
enrollment id it was issued and the keypair that enrollment holds; the scan
assertion inverts, because `scan` filters by the connection's enrollment id.

The CRAM enrolment then writes this atSign's own PKAM key back, which is what
keeps it green on both servers: today's CRAM auto-approve copies the enrolling
key over `privatekey:at_pkam_publickey`, so without the restore the owner
credential would be replaced and every later legacy login in the file would
fail. After at_server#2787 the same write is a no-op.

## How to verify

1. `cd tests/at_functional_test && ./runLocal.sh` against the published image —
   82 tests, no failures.
2. The same with `test/docker-compose.yaml` pointed at a virtual environment
   built from at_server#2787 — 82 tests, no failures.

Both were run. Unrelated flake seen while verifying, in
`test/atclient_sync_callback_test.dart`: 3 failures and 3 passes over 6 runs, on
both images, and it runs before the file this PR touches.

## Skills impact

### DIFF
--- a/tests/at_functional_test/test/enrollment_test.dart
+++ b/tests/at_functional_test/test/enrollment_test.dart
@@ -27,6 +27,22 @@ void main() {
   late String alicePkamPublicKey;
   String encryptedAPKAMSymmetricKey = '';
 
+  /// A brand-new RSA-2048 keypair, for a request that must carry a key no
+  /// enrollment on this atSign already holds.
+  ///
+  /// Every enrollment needs a keypair of its own: the atServer refuses a
+  /// request installing key material that any stored enrollment already
+  /// carries, in any status. Two requests sharing one key therefore make the
+  /// second fail, and the failure is durable server state that outlives the
+  /// run that created it.
+  ({String publicKey, String privateKey}) freshApkamPair() {
+    final pair = RsaKeyPair.generate();
+    return (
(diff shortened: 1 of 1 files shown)


gkc:
@cconstab this is a functional-test-fixture-only change in prep for https://github.com/atsign-foundation/at_server/pull/2787 so the at_client test packs don't suddenly start failing and causing confusion

cconstab (approved):

PR #2204, opened 2026-09-05
feat(at_client): AtClientStorage owns the keystore and the sync queue

gkc:
## Why now
This whole thing came up during testing on the pqc feature branch and cost me a whole day. So I'm fixing the structural problems now, and doing so with an eye on the wasm split which will be following soon after the pqc release.

## What I did

Why this PR: an `AtClient`'s local keystore and its sync queue (the local writes still to be pushed to the atServer) were opened independently — the keystore at the configured path, the queue on Hive's global instance under a box named only by atSign. Messy messy messy.

This PR is a stepping stone - the first of a short series of PRs. It introduces the client-library object that owns the atclient storage (currently just the keystore and the sync queue, maybe more in futuure), wires it in as the default with today's Hive behaviour, and adds the claim rules — without yet changing how storage is configured or released. Later PRs add in-memory and SQLite backends, let an app inject storage through a factory in place of `hiveStoragePath`, make `stop()` release storage, and move the test packs onto in-memory storage. Small steps keep each one reviewable and let trunk pick up the fix early.

## Details:
- `AtClientStorage`: keystore and sync queue as one object with a claim. A client attaches on creation; a second client is refused; after a detach a different principal (atSign + enrollment) is refused until `forgetPrincipal()` (hand over, keep the data) or `clear()` (empty both halves). `detach()` records the departing holder.
- `HiveAtClientStorage` wraps the existing `StorageManager` and `AtSyncQueue` and is the default; `LocalSecondary` takes its queue from the storage instead of opening its own. `AtSyncQueue` gains `clear()`.
- Not in this PR: refusing a second `HiveAtClientStorage` for the same atSign. Nothing releases storage until `stop()` learns to, so that guard would refuse every test that rebuilds a client for one atSign; it arrives with the release change.

## How I did it
See commit & diffs

## How to verify it
Tests pass. new test file `test/storage/hive_at_client_storage_test.dart` pins the rules - if any of the guards are removed by accident in future in the lib code then a test will fail; `at_client_storage_wiring_test.dart` pins that a created client's `LocalSecondary` shares the storage's queue by identity.

feat: `AtClientStorage` — a client's keystore and sync queue as one bundle, with `HiveAtClientStorage` as the default.

### DIFF
--- a/packages/at_client/lib/src/client/at_client_impl.dart
+++ b/packages/at_client/lib/src/client/at_client_impl.dart
@@ -8,7 +8,7 @@ import 'package:at_chops/at_chops.dart';
 import 'package:at_client/at_client.dart';
 import 'package:at_client/src/client/secondary.dart';
 import 'package:at_client/src/client/verb_builder_manager.dart';
-import 'package:at_client/src/manager/storage_manager.dart';
+import 'package:at_client/src/sync/at_sync_queue.dart';
 import 'package:at_client/src/response/response.dart';
 import 'package:at_client/src/service/encryption_service.dart';
 import 'package:at_client/src/service/file_transfer_service.dart';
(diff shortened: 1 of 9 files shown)


packages/at_client/lib/src/client/at_client_impl.dart
```diff
@@ -388,14 +391,23 @@ class AtClientImpl implements AtClient {
 
   Future<void> _init({AtLookUp? atLookUp}) async {
     if (_preference!.isLocalStoreRequired) {
+      AtSyncQueue? syncQueue;
       if (_localSecondaryKeyStore == null) {
-        _storageManager = StorageManager(preference);
-        await _storageManager!.init(_atSign, preference!.keyStoreSecret);
+        final storagePath = preference!.hiveStoragePath;
```
gkc, line 396: hiveStoragePath will be deprecated in the next PR; this is a stepping stone PR

cconstab (approved):

PR #2208, opened 2026-09-05
feat(at_client): isolate a client's storage, and release it on stop()

gkc:
## Intro / context
What this series of PRs (this one, then #2210 and #2211 will deliver:
- at_client_web will be able to inject the storage  it needs (whatever that is - we’ll likely start with fully in memory, ephemeral for the session)
- CLI apps can have fully ephemeral local storage so they leave nothing on disk
- We can switch away from hive as the default on the client
- Basically a whole load of storage enhancements tech debt removal, and structural cleanup

Now on to this PR specifically - it is small in terms of lib code changes - just paving the way, and sorting out the test pack to match. the vast majority of the PR is the changes to the test pack to use the new api

## Why

Two enrollments of one atSign must not share local storage: an enrollment holds key material
only for its granted namespaces, so a shared store hands one enrollment another's records and
pending writes. Storage was keyed by the atSign alone.

Four holes had to close with it — a client must be able to release its storage; a store must
refuse a second opener rather than share silently; a caller must be able to say where its
storage lives and on which backend; and a queued write must not be lost when a later write
supersedes it.

## What

See commits & diffs. Storage is keyed by location (the directory plus the atSign), the base
refuses a second opener, `stop()` closes what the client owns and only detaches what it was
given, `create` / `setCurrentAtSign` / `AtServiceFactory.atClient` take `storage:`, and queue
entries carry a `seq` so the drain removes only the version it pushed.

⚠️ Behaviour change: a stopped client no longer stays cached with its store open.

## How to verify it

Tests pass. Locally functional +82, e2e +32, onboarding-CLI +15. In CI the onboarding-CLI job
went fail → pass, and `end2end_test_14` 11 passed/18 failed → green. (A `bypasscache` stale
read in that job is intermittent and independent of this branch: it also fails on the merge
base.)

A client's local storage is isolated by location and released on `stop()`, a caller may supply
its own, and a delete that supersedes a queued update is no longer lost on push.

### DIFF
--- a/packages/at_client/lib/src/storage/at_client_storage.dart
+++ b/packages/at_client/lib/src/storage/at_client_storage.dart
@@ -34,10 +34,20 @@ abstract class AtClientStorage {
 }
 
 /// The claim rules every [AtClientStorage] shares; a backend supplies
-/// [openBackend] and [clearData].
+/// [location], [openBackend], [closeBackend] and [clearData].
 abstract class AtClientStorageBase implements AtClientStorage {
   AtClient? _owner;
   String? _lastPrincipal;
+  bool _closed = false;
+
+  /// The storages whose backend is open, keyed by [location].
+  ///
+  /// Two storages over one location are one store on disk, so the second is
+  /// refused rather than left to share silently. Keyed by location and not by
+  /// atSign: several clients of one atSign are legitimate so long as each was
+  /// given its own location, which is how two enrollments stay isolated.
+  static final Map<String, AtClientStorageBase> _openByLocation =
+      <String, AtClientStorageBase>{};
 
   /// The `(atSign, enrollmentId)` a client acts as.
   static String principalOf(AtClient client) =>
@@ -45,9 +55,19 @@ abstract class AtClientStorageBase implements AtClientStorage {
 
   bool get isAttached => _owner != null;
 
+  /// The store this points at, in a form two storages over the same records
+  /// report identically — everything that decides which records they resolve
+  /// to, not only where the files sit: a backend keyed by atSign within a
+  /// directory includes the atSign, since two atSigns there share nothing. An
+  /// instance whose data is private to itself reports a token unique to it.
+  String get location;
+
   @override
   Future<void> attach(AtClient owner) async {
     if (identical(_owner, owner)) return;
+    if (_closed) {
+      throw StateError('this storage has been closed and cannot be reopened');
+    }
     final holder = _owner;
     if (holder != null) {
       throw StateError('this storage is held by ${principalOf(holder)}; a '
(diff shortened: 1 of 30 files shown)


gkc:
Paused pending a design decision (D-13): local storage must be isolated per `(atSign, enrollmentId)`, not per atSign, because two enrollments of one atSign have different namespace scopes and must not share a store. The `stop()`-releases-storage change here is sound, but the per-atSign `HiveAtClientStorage` guard it carries is the wrong shape — it hid a sharing collision instead of isolating the data. Reworking behind the isolation change (X4a). See `docs/projects/wasm/decisions.md` D-13.

gkc:
Both CI failures classified, both ours and both known: `functional_tests_at_onboarding_cli` is the per-atSign Hive-box sharing that D-13 / X4a fixes (isolate storage per `(atSign, enrollmentId)`); `end2end_test_14` is a fixture issue — the `.atKeys` carry an enrollmentId but the test authenticates without supplying it, so a #2787 server refuses bare `pkam:` with AT0401. Neither is the release code in this PR; both are handled by the X4a rework and the enrollment-fixture sweep.

packages/at_client/lib/src/storage/at_client_storage.dart
```diff
@@ -60,7 +80,16 @@ abstract class AtClientStorageBase implements AtClientStorage {
...
     }
+    final here = location;
+    final occupant = _openByLocation[here];
+    if (occupant != null && !identical(occupant, this)) {
+      final held = occupant._lastPrincipal;
+      throw StateError('another storage is already open at $here'
+          '${held == null ? '' : ', last held by $held'}; close it before '
+          'opening a second there, or give this one its own location');
+    }
     await openBackend();
```
srieteja: **Claim:** `AtClientStorageBase.attach` has a check-then-act race across an `await` boundary — two concurrent storage instances for the same location can both pass the occupancy check before either records its claim.

**Mechanism:** the occupancy guard reads `_openByLocation[here]` synchronously, then `await openBackend()` (this line) runs *before* the claim is recorded at `_openByLocation[here] = this` on the next line. `HiveAtClientStorage.openBackend()` genuinely yields — it awaits `StorageManager.init()` and `AtSyncQueue.open()`. If `AtClientImpl.create()` is called twice concurrently for the same atSign + storage path with no injected storage, each call builds its own `HiveAtClientStorage` pointed at the same `location`. Both instances observe `_openByLocation[here] == null`, both pass the guard, and whichever `attach()` finishes second silently overwrites the first's map entry — defeating the invariant the comment above `_openByLocation` claims to hold ("two storages over one location... the second is refused").

**Repro:** fire two concurrent `AtClientImpl.create('@alice', prefs)` calls (no `storage:` injected) sharing one `hiveStoragePath`. Both attach successfully instead of the second throwing `StateError`. Stop the first client — `close()` only removes its own entry from `_openByLocation` when `identical(_openByLocation[here], this)`; since it was overwritten, that check is false, but `close()` still unconditionally calls `await closeBackend()`, closing the underlying Hive box the surviving second client is still using. The second client's subsequent reads/writes then hit a closed Hive box.

**Suggested fix:** record the claim (`_openByLocation[here] = this`) *before* `await openBackend()`, rolling it back on failure — or serialize `attach()` per location with a lock/mutex so the check and the claim happen atomically.

gkc: @srieteja can you supply a PR on top of this that I can review and merge in to this one if it looks good, please? Reading code diffs is easier than deciphering llm-speak

gkc: Good catch, by the way @srieteja, thank you

srieteja: Here is the fix: https://github.com/atsign-foundation/at_client_sdk/pull/2212

cconstab (approved):

PR #2212, opened 2026-09-06
fix(at_client): close the attach() race on AtClientStorageBase._openByLocation

srieteja:
## Summary
- Fixes a concurrency bug found while reviewing #2208 (see [inline comment](https://github.com/atsign-foundation/at_client_sdk/pull/2208#discussion_r3944595998)): `AtClientStorageBase.attach()` recorded its claim on `_openByLocation` *after* `await openBackend()`, so two concurrent `attach()` calls for the same location could both pass the occupancy check before either claimed it.
- Concurrent `AtClientImpl.create()` calls for the same atSign + storage path would then both attach successfully instead of the second throwing `StateError` as the guard intends. Stopping the first client would close the backend the second was still using.

## How
- `_openByLocation[here] = this` now happens before `await openBackend()`, with a rollback in a `catch` if `openBackend()` throws, so the check-then-claim is atomic with respect to the await.
- Added a regression test (`a second attach() racing the first cannot claim the same location before the first finishes opening`) using a gated fake storage to deterministically interleave two `attach()` calls; confirmed it fails against the pre-fix code and passes with the fix.
- `dart analyze` clean; full `test/storage/` suite (38 tests) passes.

Opened against `gkc-at-client-storage-release` so it can be merged straight into that branch before #2208 lands on trunk.

## Test plan
- [x] New regression test fails on pre-fix source, passes on fixed source
- [x] `dart test test/storage/` — 38/38 pass
- [x] `dart analyze` on changed files — no issues

### DIFF
--- a/packages/at_client/lib/src/storage/at_client_storage.dart
+++ b/packages/at_client/lib/src/storage/at_client_storage.dart
@@ -88,8 +88,16 @@ abstract class AtClientStorageBase implements AtClientStorage {
           '${held == null ? '' : ', last held by $held'}; close it before '
           'opening a second there, or give this one its own location');
     }
-    await openBackend();
+    // Claimed before the await, not after: two attach() calls for the same
+    // location can otherwise both read no occupant and both pass the check
+    // above while the first is still suspended inside openBackend().
     _openByLocation[here] = this;
+    try {
+      await openBackend();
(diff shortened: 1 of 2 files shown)


gkc (approved):
LGTM - thank you @srieteja !

PR #2219, opened 2026-09-07
fix(at_client): restore clientConfig, and close two storage races

srieteja:
## Summary
- Three regressions from the auth/storage refactor landed in [#2204](https://github.com/atsign-foundation/at_client_sdk/pull/2204)–[#2211](https://github.com/atsign-foundation/at_client_sdk/pull/2211), found by adversarial review of [#2208](https://github.com/atsign-foundation/at_client_sdk/pull/2208) and [#2211](https://github.com/atsign-foundation/at_client_sdk/pull/2211).
- `RemoteSecondary._installAuthenticator()` never threaded `_getClientConfig()` into any of the four authenticator shapes it builds, so every authenticated connection now sends `from:@alice` with no `:clientConfig:` segment — the atServer stops receiving client version/id/appName/platform.
- `AtClientImpl.create()` silently discarded a caller-supplied `storage` whenever a client for that atSign was already cached: the returned client kept its old store, and the caller's storage sat unattached while believed live.
- `AtClientStorageBase.close()` released its `_openByLocation` claim *before* awaiting `closeBackend()`, reopening (in reverse) the exact race `attach()` guards against — a concurrent `attach()` at the same location could open a backend over boxes/db still being torn down.

## How
- `remote_secondary.dart`: pass `clientConfig: _getClientConfig()` on all four `authenticatorFor*()` calls in `_installAuthenticator()`.
- `at_client_impl.dart`: on a cache hit in `create()`, throw `IllegalArgumentException` if `storage != null` and isn't already held by the cached client — mirrors the `_storageIsUnchanged` guard `AtClientManager.setCurrentAtSign` already has, which this lower-level factory had no equivalent of.
- `at_client_storage.dart`: move the `_openByLocation` release in `close()` to *after* `await closeBackend()`, matching `attach()`'s claim-before-await ordering.
- Added regression tests for all three (each fails on the pre-fix code, confirmed by reverting locally and re-running).
- Two-line CHANGELOG entries under the still-unpublished `3.15.0-rc1` heading.

## Test plan
- [x] `dart test test/remote_secondary_test.dart` — new clientConfig test fails without the fix, passes with it
- [x] `dart test test/storage/at_client_storage_release_test.dart` — new close()-race and create()-guard tests fail without the fix, pass with it
- [x] `dart test` (full `at_client` suite) — 718 passed, 0 failed
- [x] `dart analyze` on all three changed lib files — no new issues (only pre-existing deprecation infos)
- [x] `dart format` clean

### DIFF
--- a/packages/at_client/CHANGELOG.md
+++ b/packages/at_client/CHANGELOG.md
@@ -152,6 +152,12 @@
   crashed for a non-nullable item type).
 - fix: `AtCollection` reads no longer duplicate the preceding item when a key
   expires or is deleted between the scan and its per-key read.
+- fix: `RemoteSecondary`'s authenticator path sends client version, id,
+  appName and platform on every authenticated connection.
+- fix: `AtClientImpl.create()` rejects a caller-supplied `storage` that a
+  cached client for the same atSign isn't already holding.
+- fix: closing one client's storage can no longer race a concurrent
+  `attach()` into opening a backend over it mid-teardown.
 
 ## 3.14.0
 - feat (experimental): per-APKAM same-atSign secret-sharing substrate —
--- a/packages/at_client/lib/src/client/at_client_impl.dart
+++ b/packages/at_client/lib/src/client/at_client_impl.dart
@@ -357,6 +357,14 @@ class AtClientImpl implements AtClient {
     AtClientImpl? atClientImpl;
     if (atClientInstanceMap.containsKey(currentAtSign)) {
       atClientImpl = atClientInstanceMap[currentAtSign];
+      // Refuse a storage the cached client doesn't already hold.
+      if (storage != null && !storage.isHeldBy(atClientImpl!)) {
+        throw IllegalArgumentException(
+            'AtClientImpl.create($currentAtSign, ...): a client for this '
+            'atSign already exists and holds different storage. Call '
+            'stop() on it first, or omit storage to keep reusing the '
+            'storage the existing client already holds.');
+      }
       await atClientImpl!.start();
       // Re-using a cached AtClient skips _init. Adopt the supplied preference's
       // crypto config so providers (and a changed defaultProviderId) added
--- a/packages/at_client/lib/src/client/remote_secondary.dart
+++ b/packages/at_client/lib/src/client/remote_secondary.dart
@@ -72,6 +72,7 @@ class RemoteSecondary implements Secondary {
         _atSign,
         enrollmentId: lookUp.enrollmentId,
         chops: _atChops,
+        clientConfig: _getClientConfig(),
       );
       return;
     }
@@ -90,6 +91,7 @@ class RemoteSecondary implements Secondary {
         // The same preference fields the constructor stamps on the lookup, so
         // the authenticator and the ladder it replaces read them alike.
         hashingAlgo: _preference.hashingAlgoType,
+        clientConfig: _getClientConfig(),
       );
       return;
     }
@@ -102,6 +104,7 @@ class RemoteSecondary implements Secondary {
         _atSign,
         privateKey,
         enrollmentId: lookUp.enrollmentId,
+        clientConfig: _getClientConfig(),
       );
       return;
     }
@@ -112,7 +115,11 @@ class RemoteSecondary implements Secondary {
     // and must keep working through the seam.
     final cramSecret = _cramSecret;
     if (cramSecret != null) {
-      lookUp.authenticator = authenticatorForCramSecret(_atSign, cramSecret);
+      lookUp.authenticator = authenticatorForCramSecret(
+        _atSign,
+        cramSecret,
+        clientConfig: _getClientConfig(),
+      );
       return;
     }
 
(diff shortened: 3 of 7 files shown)


srieteja:
# PR #2219 CI failure — `at_onboarding_cli` regressed by this branch's storage fixes

**Branch:** `fix/client-config-and-storage-races`
**Run:** https://github.com/atsign-foundation/at_client_sdk/actions/runs/34119792766
**Failing jobs:** `build_and_test (at_onboarding_cli)`, `functional_tests_at_onboarding_cli (at_onboarding_cli_functional_tests)`
**Everything else:** passing (49/51 checks green — at_client's own unit suite, static analysis, e2e, wasm, etc.)

## Root cause

Both failures are the new guard added in [at_client_impl.dart](packages/at_client/lib/src/client/at_client_impl.dart) — `create()` now refuses a caller-supplied `storage` that a cached client for the atSign isn't already holding — plus the release-ordering fix in [at_client_storage.dart](packages/at_client/lib/src/storage/at_client_storage.dart). `at_onboarding_cli`'s `AtOnboardingServiceImpl._initAtClient` re-authenticates the same atSign without calling `stop()` on the prior client first, which the old code tolerated silently and the new guard now rejects.

### Failure 1 — `build_and_test (at_onboarding_cli)`

```
test/at_onboarding_cli_test.dart: A group of tests to verify at_chops creation in
onboarding_cli A test to check atOnboardingService.authenticate() returns true

Exception: AtClientImpl.create(@alice🛠, ...): a client for this atSign already
exists and holds different storage. Call stop() on it first, or omit storage to
keep reusing the storage the existing client already holds.

package:at_client/src/client/at_client_impl.dart 362:9        AtClientImpl.create
package:at_client/src/manager/at_client_manager.dart 328:31   DefaultAtServiceFactory.atClient
package:at_client/src/manager/at_client_manager.dart 146:45    AtClientManager.setCurrentAtSign
===== asynchronous gap ===========================
package:at_onboarding_cli/src/onboard/at_onboarding_service_impl.dart 122:5  AtOnboardingServiceImpl._initAtClient
===== asynchronous gap ===========================
package:at_onboarding_cli/src/onboard/at_onboarding_service_impl.dart 742:7  AtOnboardingServiceImpl.authenticate
```
Result: 35 passed, 1 failed.

### Failure 2 — `functional_tests_at_onboarding_cli`

Two tests fail with a different symptom from the same branch change (the location-claim in `AtClientStorageBase.attach`):

```
test/at_onboarding_cli_test.dart: A group of tests to assert on authenticate
functionality A test to verify update and llookup verbs with authenticated
atLookup instance

Bad state: another storage is already open at
.../storage/hive/client::@alice🛠, last held by @alice🛠|legacy; close it
before opening a second there, or give this one its own location

package:at_client/src/storage/at_client_storage.dart 106:7   AtClientStorageBase.attach
package:at_client/src/client/at_client_impl.dart 470:23       AtClientImpl._init
package:at_client/src/client/at_client_impl.dart 393:28       AtClientImpl.create
```
Second occurrence identical, for `@eve🛠`.

## Why this is a real regression, not a flake

`at_onboarding_cli`'s `authenticate()` path re-creates the `AtClient` for an atSign it has already onboarded in the same test run, without an intervening `stop()`. Pre-fix, `AtClientImpl.create()` silently reused the cached client and ignored the newly-supplied storage; post-fix it correctly refuses to discard that storage — which is the behavior [pr2208-review.md](pr2208-review.md)'s c01-F1 finding asked for — but `at_onboarding_cli` was relying on the old silent-reuse behavior to re-authenticate without cleanup.

Fix belongs in `at_onboarding_cli` (call `stop()`/`detach()` before re-authenticating, or make `_initAtClient` omit `storage` on a re-auth so it keeps reusing what the cached client already holds) — not a loosening of the new guard in `at_client`, which is the behavior the guard exists to enforce.

gkc:
@srieteja 

> `RemoteSecondary._installAuthenticator()` never threaded `_getClientConfig()` into any of the four authenticator shapes it builds, so every authenticated connection now sends `from:@alice` with no `:clientConfig:` segment — the atServer stops receiving client version/id/appName/platform.

^^ I will fix this in a follow-up


> `AtClientImpl.create()` silently discarded a caller-supplied `storage` whenever a client for that atSign was already cached: the returned client kept its old store, and the caller's storage sat unattached while believed live.

^^ I will fix this in a follow-up. Silent is a problem, it should be refused; other than that this is intentional behaviour.


> `AtClientStorageBase.close()` released its `_openByLocation` claim *before* awaiting `closeBackend()`, reopening (in reverse) the exact race `attach()` guards against — a concurrent `attach()` at the same location could open a backend over boxes/db still being torn down.

^^ Thanks - I will fix this in a follow-up.


Suggest you close this PR; I'll open a ticket for the above issues so I don't forget.

gkc:
@srieteja re my previous message: I am in the middle of dealing with the merge from trunk back into the pqc feature branch and it is monstrous, so I'd like to avoid having to deal with another merge later ... I'll be fixing forward from the feature branch from now on

srieteja:
> @srieteja re my previous message: I am in the middle of dealing with the merge from trunk back into the pqc feature branch and it is monstrous, so I'd like to avoid having to deal with another merge later ... I'll be fixing forward from the feature branch from now on

Sure @gkc. Closing this now.

PR #2220, opened 2026-09-07
feat(at_chops): AES-CTR via OpenSSL FFI, one-shot and incremental

### LINKED ISSUE #2216: Ensure that atChops provides AES CTR via OpenSSL to provide aceleration to NoPorts srv tunnels

As we rollount OpenSSL FFI we need to make sure that the srv can use AES CTR mode. I am pretty sure it is not covered right now.

Also make sure that NoPorts works as expected if the right libs are available.

AES CTR accelerated by hardware if the  right lib is in place on the device it is running on.

_No response_

### Additional context

_No response_

srieteja:
## What

- AES-CTR gets an OpenSSL FFI backend, the last bulk-traffic primitive in at_chops without one. Resolved automatically by `AtPqc`; hosts without libcrypto keep the pure-Dart path, byte for byte.
- Two shapes for two consumers: one-shot `AtPqc.aesCtr` (relay auth) and incremental `AesCtrFfiCipher` (NoPorts `srv` tunnels).
- **Does not close atsign-foundation/at_client_sdk#2216** — that closes when the NoPorts side lands (atsign-foundation/noports#2904, draft until 3.7.0 publishes).

## How

**Bindings**
- `EVP_aes_128/192/256_ctr` bound alongside the existing GCM symbol; every other EVP symbol needed already existed.
- `libCryptoSupportsAesCtr` goes through `EVP_CIPHER_fetch`, so it's provider-aware, not a bare `dlsym` check.

**One-shot — `AesCtrFfiAlgo`**
- PKCS7-pads before CTR, deliberately: CTR needs no padding, but `AESEncryptionAlgo` always has, and relay auth has that on the wire. Omitting it would break auth between a host that resolves FFI and one that resolves pure-Dart.
- Both native buffers are sized from the padded length, not the plaintext length — GCM's output ≤ input, but here the padded length crosses that boundary. 1/15/17-byte cases pin it.
- Accepts 16/24/32-byte keys, matching `AesCtrFactory`; copying the GCM algo's 32-only rule would've regressed a path that already works pure-Dart.

**Incremental — `AesCtrFfiCipher`**
- One `EVP_CIPHER_CTX` from construction to `dispose()` — the first native resource in at_chops that outlives a method call.
- CTR's block size is 1, so `EVP_EncryptUpdate` emits exactly `inl` bytes and the ctx holds the partial-block keystream offset — that's what makes arbitrary chunk sizes work and `EVP_EncryptFinal_ex` a no-op.
- No padding here; the tunnel is a raw keystream. The two shapes differ on purpose, documented in the dartdoc.
- `dispose()` is idempotent; `update()` after it throws `StateError`. `Finalizer` is a backstop only.
- IV is length-checked to exactly 16 at construction — a short IV is an out-of-bounds native read, not a wrong answer.

**Verification**
- `test/aes_ctr_ffi_test.dart` (`@Tags(['ffi'])`): round-trip at all three key lengths, cross-decrypt both directions against `AESEncryptionAlgo` (the padding guard), empty/non-block-aligned inputs, chunked `update()` at 1/7/16/1000 bytes matching the one-shot output, post-`dispose()` contract.
- `wire_literal_pins_test.dart` gains a frozen key/IV/plaintext → hardcoded ciphertext hex, satisfied by both backends.
- 458 tests default, 75 under `-t ffi`, all green.

**Release**
- 3.6.1 → 3.7.0. New API, no breaks.
- FFI classes exported from `at_chops_ffi.dart` only; the default barrel stays pure-Dart and web-safe.


cconstab:
This came up in code rewiew....

**`inl` binding width (pre-existing, but first reachable here)**

`input.length` is passed as the `inl` argument of `EVP_EncryptUpdate`. The shared typedef in `openssl_ffi_bindings.dart` declares that parameter as `IntPtr` (line 221, last parameter; same on 226 for `EvpDecryptUpdateNative`):

```dart
typedef EvpEncryptUpdateNative = Int32 Function(Pointer<EVP_CIPHER_CTX>,
    Pointer<Uint8>, Pointer<Int32>, Pointer<Uint8>, IntPtr);
```

The OpenSSL prototype is:

```c
int EVP_EncryptUpdate(EVP_CIPHER_CTX *ctx, unsigned char *out,
                      int *outl, const unsigned char *in, int inl);
```

`inl` is a C `int`, not `size_t`. On 64-bit ABIs the callee reads the low 32 bits of the register, so:

- an input of `2^32 + n` bytes is processed as `n` bytes, `outl` comes back as `n`, we copy `n` bytes out, and no error is raised (silent truncation);
- an input in `[2^31, 2^32)` reads as negative, OpenSSL returns 0, and the `<= 0` check catches it.

The typedef predates this PR (the GCM binding shares it), but `update()` is the first API where a caller hands us arbitrary-sized chunks over a long-lived context, so this is the first place the truncation is actually reachable.

Either fix is fine:

1. Change both typedefs to `Int32` so the Dart signature matches the C one. Dart then throws on out-of-range values instead of wrapping.
2. Or keep the typedef and add a guard after the empty-input return on line 122:

   ```dart
   if (input.length > 0x7fffffff) {
     throw ArgumentError.value(input.length, 'input',
         'exceeds the int32 limit of EVP_EncryptUpdate');
   }
   ```

   with the same check before `_encryptUpdate` / `_decryptUpdate` in `aes_ctr_ffi_algo.dart`.

Option 1 is the more honest binding; option 2 is the smaller diff.

srieteja:
Thank you for the review @cconstab. I have applied both the fixes recommended in your comment through [036366a](https://github.com/atsign-foundation/at_client_sdk/pull/2220/commits/036366ac44ec47827ba39278efab06104643e5f2). Both were applied because option 1's claim was incorrect: Dart does not throw on out-of-range values; it silently truncates (verified by a probe). But it does ensure that the Dart signature matches the C signature. Option 2 actually fixes the issue you originally pointed out.

cconstab (dismissed):

gkc:
@srieteja I've got a few small changes which I'm putting into a PR on this PR

gkc:
@srieteja https://github.com/atsign-foundation/at_client_sdk/pull/2226

cconstab (approved):

PR #2214, opened 2026-09-06
feat(at_utils)!: split dart:io off the default barrel, open 4.0.0-rc1

srieteja:
## What

Takes `dart:io` out of at_utils' default barrel, so `package:at_utils/at_utils.dart` can sit on a web build graph, and opens at_utils 4.0.0 as `4.0.0-rc1`. This is step 3 of the publish ladder in `docs/projects/wasm/implementation-plan.md` (I1–I4), and the plan for this package is `docs/major-v4/at_utils.md`.

**Where to look**

| File | What to decide |
| --- | --- |
| `packages/at_utils/lib/at_utils.dart`, `at_utils_io.dart` | Whether the split is right. `PseudoServerSocket`, `ApplicationConfiguration` and the file, stderr and CLI logging handlers move to the new `at_utils_io.dart`, which re-exports `at_utils.dart` |
| `packages/at_utils/lib/src/logging/atsignlogger.dart` | `AtSignLogger.stdErrLoggingHandler` is removed. It held a `StdErrLoggingHandler`, so leaving it would have kept `dart:io` reachable from the neutral barrel |
| `packages/at_utils/lib/src/atsign_util.dart` | `AtUtils.formatAtSign`, deprecated since 3.x, is removed |
| `packages/at_client/lib/src/rpc/at_rpc.dart` | ⚠️ `AtRpcClient`'s logger no longer forces stderr. It uses `AtSignLogger`'s default handler, so its output goes wherever the app points that |
| `.github/wasm_gates.yaml` | at_utils is gated at 0 blocked packages. at_chops' ceiling drops from 2 to 0, and at_auth's to 2 now that #2162 has taken at_server_status off it too |
| `.github/workflows/at_client_sdk.yaml` | A new `analyze_tests` job for the four `tests/` harnesses, which are workspace members nothing analyzed |
| `pubspec.yaml` (root) | ⚠️ A temporary `dependency_overrides` git ref for `at_persistence_secondary_server`, pointing at at_server's `st/at_persistence-at_utils-4.0.0-compat` - at_server trunk plus `at_utils: ">=3.0.19 <5.0.0"`. It comes out once at_persistence_secondary_server is published with that range |

**Safe to skip**

- The 16 `at_utils: ^4.0.0-rc1` constraints in workspace pubspecs and example apps. `^4.0.0` would not admit the prerelease.
- The `at_utils` path overrides added to example apps that already override their other local packages.
[...]

PR #2223, opened 2026-09-08
feat(at_lookup)!: 4.0.0 — the transport split

srieteja:
Gets `dart:io` out of `package:at_lookup/at_lookup.dart` behind an `AtTransport` seam, ending in a green `at_lookup` stanza in `.github/wasm_gates.yaml`. Next rung on the WASM publish ladder after at_utils 4.0.0 (#2214), using the same **subtractive** split both at_utils and at_auth used: the default barrel is neutral, native code moves to `at_lookup_io.dart`, consumers opt in.

> **Stacked on #2214.** Base is `trunk`, but the branch is cut from `st/at_utils-v4`, so this diff carries that PR's commits until it merges. Review `packages/at_lookup` and `.github/wasm_gates.yaml`; everything under `packages/at_utils` belongs to #2214.

> **Draft:** steps 8-9 of the plan (credential-ladder removals; version/changelog/docs) are still to land on this branch.

## What

- **`AtTransport`** — a byte channel (`inbound` / `add` / `flush` / `destroy` / `description`) replaces `Socket` on the `AtConnection` interface. `SecureSocketTransport` is the native implementation; a `WebSocketTransport` is the seam's point, deferred to a later release.
- **`at_lookup.dart` is web-safe.** `dart compile wasm` passes on the neutral barrel and the import-graph walk reports zero `dart:io` offenders, both enforced in CI by a new `wasm_gates.yaml` stanza.
- **`at_lookup_io.dart` holds everything native** — `SecureSocketTransport(Factory)`, `SecureSocketUtil`, `CacheableSecondaryAddressFinder`, `AtLookupSecureSocketFactory`, and an `atLookupOverSecureSocket()` shim for the old constructor shape.
- **`MonitorClient` is deleted** — deprecated, zero callers.
- **Consumer rule, one line for the CHANGELOG:** if your code names a socket, add the `at_lookup_io.dart` import; if it does not, nothing changes.

## How

- **`AtConnection`**: `Socket getSocket()` → `Stream<List<int>> get inbound` + `void add(List<int>)`. `BaseConnection` holds an `AtTransport`; `write` becomes `add(utf8.encode(...))` + `flush()`; `setOption` drops (`SecureSocketUtil` already sets `tcpNoDelay`). Reconciled two pre-existing inconsistencies in the same pass — `close()` was `async` over a synchronous body, and `write` disagreed on its return type between interface and impl.
- **`destroy()` is synchronous, deliberately.** Nothing can act on its completion; graceful shutdown is `await flush(); destroy();` spelled at the call site. On a web transport an awaitable teardown is actively harmful — a WebSocket close handshake waits on a peer that may never answer.
- **`inbound` MUST return the underlying stream, never a `StreamController` re-broadcast** — stated as an invariant in the doc comment. A controller converts counted pauses into a local buffer and destroys the TCP back-pressure `_pauseIfBufferFull` exists to provide. `socket_delivery_test.dart`'s `pauseCount` assertion is the mechanical guard.
- **`AtLookupTransport` → `AtLookupTransportFactories`.** Two exported types named "Transport" meaning different things (a factory bundle vs a byte channel) had already misled once in writing. Contained to at_lookup — 4 in-package sites; downstream reaches it through `secureSocketTransport()` and never spells the return type.
- **`AtLookupImpl`'s `?? AtLookupSecureSocketFactory()` defaults are gone** — they physically pulled `dart:io` into the neutral library, and `@Deprecated` is invisible to a graph walk. `secondaryAddressFinder` and `transportFactory` are now required and non-nullable; `secureSocketConfig` moves into the native factory's constructor, where TLS config belongs. The constructor keeps its name: 18 of 38 call sites already passed both arguments, and privatising would force a `@visibleForTesting` public constructor back in under a worse name.
- **`CacheableSecondaryAddressFinder` moves wholesale**, not retyped — retyping yields a type that compiles in a browser then throws on the raw-TLS directory lookup. 12 files outside at_lookup gain the `at_lookup_io.dart` import, every one a workspace member; a neutral `ProxySecondaryAddressFinder` covers callers that only need the interface.
- **`AtLookupSecureSocketFactory` survives** and is documented as *not* an `AtTransportFactory`: `SecondaryUrlFinder` carries a `SecureSocketConfig` **per call**, which `AtTransportFactory.connect(host, port, {timeout})` has no parameter for.
- **Tests**: the 16 `(x.getSocket() as MockSecureSocket).destroyed` assertions used the production interface as an observation channel — replaced by holding the injected double. `MockSecureSocket` and `FakeAtServerSocket` held one role and collapsed into `FakeAtServerTransport`, over a **single-subscription** `StreamController` (a broadcast controller ignores `pause()`, so a back-pressure assertion written against one passes vacuously).
- **Gate verified red as well as green.** Re-adding the `secure_socket_util.dart` export turns the ratchet red on both axes (`1/0 offenders, 1/0 blocked`) while `dart compile wasm` stays **green** — `dart compile wasm` does not reject `dart:io`, it stubs it and throws at call time, which is why the graph walk exists alongside the compile.

### Downstream, already in these commits

`resolution: workspace` means breaking at_lookup changes and their uptake land together: at_client, at_client_flutter, at_auth, at_onboarding_cli and at_server_status are migrated here. **Only workspace members** — a non-member resolves at_lookup from pub.dev and cannot see this branch's barrel at all.

### Not in this PR

- **at_server** pins `at_lookup: ^3.6.0` (and one exact `3.0.49`), so 4.0.0 does not break its build — it will not pick the major up. Migration obligation recorded in the plan, including `outbound_connection_wrapper.dart`'s `// ignore: implementation_imports` private-source import of `outbound_message_listener.dart`, which no barrel-level analysis can see.
- **The in-tree Flutter packages.** `at_contacts_flutter`, `at_events_flutter`, `at_location_flutter` and `at_notify_flutter` are not workspace members and pin `at_lookup: ^3.0.49`, which can never resolve 4.0.0. Their `CacheableSecondaryAddressFinder` sites migrate when 4.0.0 publishes and the pins widen — same category as at_server, and recorded next to it in the plan.
- **Publishing.** at_lookup 4.0.0's pubspec must declare `at_utils: ^4.0.0-rc1` or its web-safety claim is a lie to external consumers — and that is unpublishable until at_persistence_secondary_server widens its `at_utils: ^3.0.19` pin. The gate work below is not blocked by it.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


gkc:
Where this stands after this weekend's trunk work (#2233's `lookUps:` injection, D-15, and #2162):

**The at_lookup split is still needed.** `lookUps:` lets an app choose where its connections come from, but it doesn't make at_lookup's default barrel web-safe. D-15 says as much: "at_lookup's own major stays where it was: T3 … T4 … T5 … and T8", and the factory type "still reaches `dart:io` transitively today". Running this PR's own `at_lookup` gate stanza against trunk's at_lookup code:

```
package:at_lookup/at_lookup.dart — 633 files walked, 6/0 offenders, 1/0 blocked FAILED
  lib/src/at_lookup_impl.dart
  lib/src/cache/cacheable_secondary_address_finder.dart
  lib/src/connection/at_connection.dart
  lib/src/connection/base_connection.dart
  lib/src/monitor_client.dart
  lib/src/util/secure_socket_util.dart
```

`dart compile wasm` still passes on it, as you found. Trunk still has `Socket getSocket()` on `AtConnection`, the `?? AtLookupSecureSocketFactory()` defaults in `AtLookupImpl`, `MonitorClient`, and no at_lookup gate. So `AtTransport`, the byte-speaking `AtConnection`, moving the native code into `at_lookup_io.dart` and the gate stanza are all still wanted.

**The downstream uptake is mostly superseded.** Trunk now builds lookups through `secureSocketLookUps` (in `at_lookup_io.dart`) and at_client's `defaultLookUps`, so most of the per-site finder and transport plumbing here no longer has a site to land on:
- real `AtLookupImpl(` / `AtLookUp.withSecureSocket(` constructions outside at_lookup went from 49 at this branch's merge base to 7 on trunk. In package `lib` code that is 9 down to 2, and the two left are at_auth's `atLookUp ??=` fallbacks in `at_auth_impl.dart` and `enrollment_handshake.dart`. at_auth's `validateAtServer` also still builds its own `CacheableSecondaryAddressFinder` and socket probe when none is supplied, which is not a lookup construction but is the same kind of fallback. (The rest of trunk's live-pack and unit-test sites are 3 and 2; the other lines matching those names are `MockAtLookupImpl()` doubles and doc comments.)
- at_client_flutter's `enrollment_service.dart` is deleted on trunk, and most of the other downstream files here have been rewritten since (`at_onboarding_service_impl.dart` +48 -897, `at_auth_impl.dart` +41 -178)
- merging trunk in conflicts in 17 files, including at_lookup's own `at_lookup_io.dart`, `at_lookup.dart` and `at_lookup_impl.dart`, where trunk added `AtLookUpFactory`, `secureSocketLookUps` and `onConnect`. The `atLookupOverSecureSocket()` shim and the `AtLookupTransport` rename overlap with what `secureSocketLookUps` now builds on.

**Suggested way forward:** rather than merging trunk into this branch, rebuild it from trunk once #2214 has landed. Bring over the at_lookup package commits, hang `SecureSocketTransport` off `secureSocketLookUps`, and redo the uptake against the factory. That should now be at_auth's three fallbacks (the two lookups and `validateAtServer`'s finder and probe), at_server_status's `secureSocketLookUps()` default, and the tests that still construct lookups directly.

PR #2232, opened 2026-09-13
test: the doc rails resolve every link, count every cell, read every row

gkc:
## What

- `docs_structure_test.dart` gains three rails over the tracked Markdown under `docs/`: every relative link reaches a file or directory and every anchor a heading (GitHub's duplicate-slug suffixes honoured, code spans and fences skipped); every table row carries its header's cell count; and the owed-says-done guard reads each TODO row's own cells rather than only the section a row names.
- PQ plan: the `ApkamSigning` row names only what is still owed (NoPorts' migration to `signingKeys`); the doc-set-rails row leaves the list with this.
- wasm plan: two links to `plans/wasm/…`, a directory that never existed in git, are plain text now, and the plan's own count of those references moves with them.

## How

See commit. The eight malformed rows the plan recorded on 2026-09-08 were rebuilt in e38aa41fc; the cell rail sees five at ad354b85f and zero today, so the zero is measured rather than blind.

## How to verify

Tests pass. `docs_structure_test` 29/29, and each new test was mutation-checked: a planted bad link and anchor, a planted extra cell, and a planted ✅ on a row each go red naming the site.

## Skills impact


gkc:
closing as unnecessary

PR #2239, opened 2026-09-14
test(functional): mint locks and timeouts that survive a slow runner

gkc:
**- What I did**

Made the functional pack's mint-lock tests and key package amendment tests survive a slow CI runner.

**- Why**

- `functional_tests_storage (sqlite)` failed 2 of the last 40 runs with the same three tests, on branches that don't share code changes (#2233's branch and #2238). Both failing jobs took ~13 min for a test step that normally takes 6.5-7 min, so the whole runner was running at about half speed.
- Example - the rotation tests give their ring a 1-second mint lock so they can wait out the cooldown quickly, justified by a mint measuring 35ms locally. The ring stamps a lease when it takes the lock and abandons the mint rather than publishing once the lease is spent, which is correct behaviour, so on a slow runner a mint that ran past 1s threw `the mint lock ... expired while this client was minting` and failed the test.
- The key package amendment tests that failed ran on the default 30s timeout. They normally take 11-14s, and on the slow runs they hit 30s. Their siblings in the same file already set 4-5 minutes.

**- How I did it**

- One `liveMintLockTtl` (5s) and `waitOutMintLock()` in the pack's `test_utils.dart`, replacing four separate one-second copies (rotation, rollout ladder, published ring, signing-root mint lock).
- Explicit 4-minute timeouts on the three amendment tests that had none.

**- How to verify it**

Tests pass, including the full functional pack locally with sqlite storage.

Reproduced the CI failure and its absence with the atServer container throttled to 0.15 CPU, a fresh virtualenv per session, sessions interleaved, only the lock ttl differing between arms:

| session | lock ttl | lease expired | result |
| --- | --- | --- | --- |
| 1 | 1s | yes | UC-A5.3 failed, as in CI |
| 2 | 5s | no | 5/5 passed |
| 3 | 1s | yes | UC-G2.5 failed, as in CI |
| 4 | 5s | no | 5/5 passed |

Each extra cooldown wait adds about 4s; there are eight of them.


cpswan:
@gkc I'm unfamiliar with the 'mint' terminology here, and coming up short with my prompting/searching. Can you please add just a little more explanation.

gkc:
> @gkc I'm unfamiliar with the 'mint' terminology here, and coming up short with my prompting/searching. Can you please add just a little more explanation.

In general I use 'mint' for anything that is generating new cryptographic material.

"Mint lock" refers to situations where the stuff being minted needs to be minted by just one thing. SOmetimes those locks are internal to the client, sometimes they are cross-client via short-lived immutable records on the atServer

cpswan (approved):

PR #2242, opened 2026-09-14
fix(at_client): a monitor reconnect resumes from the last notification received

srieteja:
## What

- With `fetchOfflineNotifications` off, a monitor reconnect no longer re-delivers every notification received since the notification service was created.
- The first connect still resumes from the service's creation time, so replies sent before `monitor:` went out are still delivered.
- Follow-up to #2237.

## How

- `AtLookupImpl` re-reads `getLastNotificationTime()` on every reconnect and sends `monitor:<epochMillis>`; the atServer replays notifications strictly after it.
- `NotificationServiceImpl` now holds `_lastReceivedMillis`, advanced in `handleNotificationReceipt` for non-stats notifications, only ever forwards, and before the store write and its `isStopped` return.
- With offline fetch off, `getLastNotificationTime()` answers `_lastReceivedMillis ?? _createdAt`; it still never reads or seeds the store.
  - In memory by design: a restart with offline fetch off must start from creation again.
- The offline-on path is unchanged.
- Tests:
  - `notification_service_test.dart`: the watermark follows a receipt without reading the store, is not moved back by an earlier one, and ignores stats notifications. The first two fail against trunk.
  - `monitor_reconnect_no_replay_live_test.dart` (new): notify, drop the notification connection, notify again; the first is delivered exactly once. `monitor_reconnect_live_test.dart` cannot catch this, since a stale watermark replays a superset.
  - The live test has not been run against trunk; the unit tests are the red-on-trunk evidence.
- `AtClientPreference.fetchOfflineNotifications` doc: "delivered once either way".


gkc:
Good catch @srieteja thank you, how did you find it?

gkc (approved):

PR #2244, opened 2026-09-15
fix: stop reaching the atServer once refused or stopped

gkc:
**- What I did**

- fix(at_client): stop the sync push loop cleanly and count its progress
- fix: stop reaching the atServer once refused or stopped
- fix: give up only on an answer from the atServer
- fix(at_client): finish a stop however far its work had got
- fix(at_client): answer a failed sync request once
- test: the functional pack stops leaving unsigned key packages and running clients behind, and the CLI pack stops its cached clients before deleting their storage

**- Why**

- Read out of the logs of green `functional_tests` jobs, as #2240 was. A stopped client kept reconnecting and authenticating (32 refusals in one job after its own stop), a revoked enrollment's notification reconnect retried every 34s for the life of the process, the sync stop signal was logged at severe as a failed batch entry, and two tests left unsigned key packages on `@alice🛠` that every later seal logged SEVERE about - 860 lines across four jobs.
- A review of the branch found the rest: the refusal check counted any authentication failure as final, an expired enrollment is AT0028 on the wire and was not covered, and a `sync()` request carrying `onDone` was rerun on every microtask after a failure.

**- How I did it**

See commits

**- How to verify it**

Tests pass, including all three live packs locally. In the functional pack the `not a signed envelope` and `Box not found` lines fall from about 215 and 123 a job to 0.

A client stops reaching its atServer once it is stopped, or once the atServer has answered that its enrollment is denied, pending, revoked or expired.


gkc:
Closing this, as I'm going to rebuild it from trunk on a new branch. The live-pack log noise that prompted it needs a deterministic way to stop an AtClient (services stopped, every connection and timer closed, and work in flight able to see the stop and log accordingly), but this PR went further and changed what the client does when the atServer refuses its credentials. A review found that the refusal handling brought defects of its own, including an isolate freeze when an onError callback retries sync(), a notification listener that stays deaf after a restart, and a stopped client whose process never exits. The new PR will carry over the sync abandon on stop, the onDone rerun fix and the test-pack hygiene from here.

Issue #2117, opened 2026-08-03
feat: intent-based client operations — write the intent, then progress it through its states


gkc:
## The problem

A lot of client operations need connectivity to complete, and each one carries its own hand-rolled
handling for what happens when it isn't there: its own retry, its own queueing, its own partial-
failure recovery, its own ordering assumptions. The logic is spread across call sites, hard to
reason about, and each new operation re-invents it — usually slightly differently.

## Proposal

An **intent-based** approach. Rather than a call that must complete now:

1. Write the intent, and the data associated with it, durably.
2. Progress that intent through the states appropriate to it, as connectivity and preconditions
   allow.

## Why it's worth doing

- **One mechanism instead of many.** Retry, queueing and recovery become properties of the intent
  machinery, not of each call site.
- **Ordering falls out of it.** #2116 (ordering across keystore and notification operations)
  becomes a property of the intent log rather than something each pair of operations has to
  arrange between themselves.
- **Uniform offline behaviour.** "What happens if this is called offline" gets one answer.
- **Partial failure becomes inspectable.** An intent stuck part-way through is a record you can
  look at, rather than an operation that silently didn't happen.
- **It removes a class of trade-off we are currently making by hand.** The secret-sharing envelope
  write is a live example: ordering was bought by giving up offline tolerance because there was no
  way to have both.

## Scope

Not part of the PQ / D1 program. Design-level; would want its own shape-up before any
implementation.

gkc:
A catalogue of coupled operations, for the sweep when this is implemented.

Each row below is a sequence where one durable step lands and a later step completes it. An interruption between the two can leave the result wrong or lost. The interruption might be a client stop, the process exiting or crashing, the atServer going away, or a lost network response. Line numbers refer to trunk at `688486e44`.

The search covered `packages/at_client/lib` and `packages/at_auth/lib`. It didn't cover at_lookup, at_onboarding_cli or the Flutter packages, so there may be more. I read each sequence and its repair path in source, except the rows marked (inventory), which a search found and I haven't re-read.

| Case | Operation | Durable step, then completing step | What an interruption leaves | Repair today | Triggers |
|---|---|---|---|---|---|
[...]

PR #2245, opened 2026-09-15
fix: stop an AtClient deterministically

gkc:
**- Why**
- I wanted to clean out the noise from the test packs' output logs 

which led to:
- I needed a guaranteed way to stop a client, where work in progress can tell it was purposefully stopped by the application rather than just logging a failure

**- What I did**
- `AtClient.stop()` now stops deterministically: the sync and notification services stop, every connection and timer closes, and `stop()` returns once all of that has happened, so the process can exit
- Work a stop cuts short fails with `StoppedException` (at_commons, deliberately not an `AtException`), so fallback catches let it through and background work logs one warning rather than severe lines with stack traces
- `AtLookUp.close()` ends a lookup for good; `dropConnection()` is the old behaviour
- at_lookup's TLS connect is rebuilt on `RawSocket` + `RawSecureSocket`, because a `SecureSocket` handshake stalled on a silent peer can't be ended and held `stop()` (and the process) for ever
- Fixed the half-finished operations a stop could leave behind: key material filed before it's advertised or before its envelope is deleted, a filed signing root or nskey generation published at the next start, a minted enrollment key conveyed before `enroll:approve`, the notification watermark saved after delivery
- Replaces #2244

**- How I did it**
See commits. Worth a look in review:
- the name `StoppedException`
- at_auth's `apkamSymmetricKeyResolver` is now a stream of candidate keys (breaking within 4.0.0-rc); the enrollee keeps the one that decrypts its keys; this is a workaround for a potential problem (approver approves but then exits before conveyance) which will be properly solved when we implement #2117 
- `notify` cut short by a stop now throws `StoppedException` instead of returning an undelivered result
- a signing root the record doesn't publish signs nothing, but a failed read of that record still signs

**- How to verify it**
Tests pass, including a new test that a stopped busy client lets its process exit

fix: `AtClient.stop()` closes every connection and timer and returns promptly, and work it cuts short fails with `StoppedException`


packages/at_client/lib/src/lifecycle/at_connection.dart
```diff
@@ -198,8 +216,11 @@ class AtConnection {
   /// offline with [AtConnectionCause.stopped], emitted as the last change,
   /// and [changes] is done.
   Future<void> close() async {
-    await report(AtConnectionState.offline(AtConnectionCause.stopped));
+    if (_closed) return;
+    final reported = report(_stoppedState);
     _closed = true;
+    _closing.complete();
+    await reported;
     await _changes.close();
```
JeremyTubongbanua: `await _changes.close()` will hang forever if any consumer of the public `changes` stream has a paused subscription

Since `_changes` is of type `StreamController.broadcast()`. The `close()` returns a Future that won't complete until subscribers processes every done event. But a paused subscription will never finish the processing events, causing `_changes.close()` to hang forever.

My suggestion is to use `unawaited` (e.g. `unawaited(_changes.close())` so that paused subscribers still get it, it's just that close() won't hang forever now.

packages/at_client/lib/src/manager/monitor.dart
```diff
@@ -334,6 +354,19 @@ class Monitor {
...
+
+  /// Stops the monitor for good: ends [lookUp], so a start in flight fails at
+  /// once, and returns when the teardown has run.
+  Future<void> close() async {
+    if (_closed) return;
+    _closed = true;
+    stop();
+    await lookUp.close();
+    await _lifecycle;
+    await currentStateStreamController.close();
```
JeremyTubongbanua: same story here on line 367 - this `close()` function will hang forever if paused subscribers . Same fix here I think just `unawaited(currentStateStreamController.close())`

packages/at_client/lib/src/client/at_client_impl.dart
JeremyTubongbanua, line 1428: Claude just says that there's missing regression test coverage here for this line:

There's no test that covers the scenario of a paused `connection.changes` and then calls `stop` . Since this PR's goal is to make stops deterministic, a paused listener is a valid scenario. Same for Monitor.close() with a paused currentStateStream subscription

JeremyTubongbanua (changes requested):
the first 2 requested changes are abt some edge cases when dealing with paused subscribers 

i'd advise adding tests for paused subscribers

packages/at_client/lib/src/service/envelope_enrollment_conveyance.dart
```diff
@@ -48,6 +50,35 @@ class EnvelopeEnrollmentConveyance implements EnrollmentConveyance {
...
+    if (!sharing.isRegistered) {
+      throw AtEnrollmentException(
+          'Enrollment ${pending.enrollmentId} expects this approver to '
+          'convey its symmetric key, but this client has not registered a key '
+          'package to seal it from. Call register() on '
+          'AtClientSecretSharing.forClient(atClient) before approving.');
+    }
+    final package = readAdvertisedKeyPackage(pending.metadata?['keyPackage'],
+        enrollmentId: pending.enrollmentId!);
+    if (package == null) return;
```
JeremyTubongbanua, line 84: Maybe we should `throw` instead of silently returning here.

`readAdvertisedKeyPackage` returning `null` for a non-null input means the package is structurally unparseable.  The caller in approve() continues into `_atEnrollmentImpl.approve()` with the minted key, which is irreversible - the enrollee gets approved but never receives the key.

gkc:
Thanks @JeremyTubongbanua!

packages/at_client/lib/src/client/at_client_impl.dart
JeremyTubongbanua, line 1346: Concurrent `stop()` callers currently do not share similar teardown completion behaviour, which in my opinion violates the deterministic stop that this PR aims to achieve.

good context for an agent:

```
 Yes. This is a valid concurrency bug in the current PR head, 399159101ac421412b8c750777d28a8734efd831.

 At at_client_impl.dart:1323-1326
 (https://github.com/atsign-foundation/at_client_sdk/blob/399159101ac421412b8c750777d28a8734efd831/package
 s/at_client/lib/src/client/at_client_impl.dart#L1323-L1326):

 1. Caller A enters _stop(), sets _isStopped = true, then suspends while awaiting teardown.
 2. Caller B calls stop() before A finishes.
 3. B sees _isStopped == true and returns an already-completed future.
 4. Therefore, await B.stop() completes while services, connections, or storage may still be live. B also
    misses any teardown defect later reported to A.

 That contradicts this PR's stated contract that stop() returns only after teardown completes. The current
 test calls stop() sequentially, so it does not exercise this race.

 Using the _closing ??= pattern from AtLookupImpl
 (https://github.com/atsign-foundation/at_client_sdk/blob/399159101ac421412b8c750777d28a8734efd831/package
 s/at_lookup/lib/src/at_lookup_impl.dart#L936-L947) is the correct approach: cache the in-progress
 teardown future and return it to every caller.

 One caveat: stopHandingOverStorage() also calls _stop(), but with different storage semantics. Both entry
 points need to share the same teardown state deliberately. A naive _stopping ??= only in stop() would
 leave the mixed stop() versus stopHandingOverStorage() race unresolved.
```

gkc:
Thanks @JeremyTubongbanua - responded to your earlier comments; looking at your most recent one now

**The paused-subscriber hangs.** Confirmed with a probe: closing a broadcast controller with one paused subscriber is still pending at 2s, where an active subscriber or none closes straight away. `AtConnection.close()` and `Monitor.close()` both awaited exactly that inside `stop()`, so a paused subscriber held the stop open. Sweeping for the same shape found a third: `AtClient.dataEvents`, which an app is just as free to pause. All three closes are now unawaited, and the probe also shows the paused subscriber still gets its done event when it resumes, so nothing is lost by not waiting.

**The missing coverage.** Three tests, one per stream: a paused listener on `changes`, on the monitor's state stream, and on `dataEvents`. Each asserts the close returns and that the done event still arrives after a resume. Restoring each `await` reddens exactly its own test.

**The unparseable key package.** Agreed, and with one wrinkle: refusing on every package we cannot read would also block approvals across a version skew, which the seam test deliberately allows ("a package written by a newer client is nothing the approver can fix"). So `readAdvertisedKeyPackage` now reports why it failed, and only a package that is not a key package refuses before `enroll:approve` - one a newer client wrote approves as before. Two tests, and two mutations: dropping the refusal reddens the first, widening it to any unreadable package reddens the second.

packages/at_client/lib/src/client/at_client_impl.dart
gkc: Good catch - reproduced before fixing it, at 53344deb9. With the first caller parked mid-close, the second `stop()` returned straight away, so it was handed a client whose connections were still live.

Both entry points now join one in-flight teardown, the `_closing ??=` shape from `AtLookupImpl`, so every caller returns only once that teardown has run and sees the same defect if it raised one.

Your caveat about the two entry points is decided deliberately rather than left to the race: `stopHandingOverStorage()` joins the same teardown and hands the storage on ONLY to the caller that started it. A second caller gets null - the teardown it joined may be an ordinary stop that releases the storage, and handing the same store to a second successor would give two clients one store. `start()` clears the cached future, or a restarted client's next stop would be answered from the old one and tear down nothing.

Three tests, each with the mutation that reddens it: a second stop that must wait (returning on the flag alone reddens it), a hand-over joining an ordinary stop that must hand back nothing (handing it over anyway reddens it), and a restarted client whose next stop runs a real teardown (keeping the stale future reddens it).

at_client is at 2075 tests, analyze and the format gate clean.

JeremyTubongbanua (dismissed):

Xlin123:
lgtm, nits in #2251

srieteja:
**1. [`pq_client_bootstrap.dart:237`](packages/at_client/lib/src/client/pq_client_bootstrap.dart#L237) — medium**

The new `fileReceivedSecret` hook runs *after* `secretStore.putIfNewer` has already committed the secret. If the hook throws (keyfile write fails), the secret is never emitted on `receivedSecrets` — and because the dedupe gate (`putIfNewer` returning `false`) skips the hook on every later sweep of the same envelope, the filing is never retried before the envelope is swept away. Net effect: key material silently lost, no error surfaced to the caller.

**2. [`collections.dart:4616`](packages/at_client/lib/src/collections/collections.dart#L4616) — medium**

`expiringSoonEvents()` emits into its own private `StreamController`. On client stop, the backing scheduler disposes correctly, but nothing closes that controller — a consumer doing `await for` on the stream hangs forever. `watch()` / `availableEvents()` on the same collection terminate cleanly on stop; this stream doesn't.

**3. [`sync_service.dart:294`](packages/at_client/lib/src/service/sync_service.dart#L294) — low**

The new `progress.stopped` branch in `waitUntilCaughtUp` completes the future with `StoppedException` but skips `detach()`. Every call on an already-closed service leaks one listener (and its closure) for the life of the service object.

Fixes for all three in #2252.

srieteja (approved):

PR #2248, opened 2026-09-16
test(at_chops): direct AesCtrFactory coverage, and AESEncryptionAlgo's error/fallback paths

srieteja:
## Summary
- `AesCtrFactory.createEncryptionAlgo` had no direct test — only exercised indirectly through `AESEncryptionAlgo`'s valid-length calls. Adds variant selection by key length (`secretKeyLength`) and `AtEncryptionException` on an invalid length.
- `AESEncryptionAlgo` gets the same invalid-key-length case its FFI sibling (`AesCtrFfiAlgo`) already has.
- Pins the zero-IV fallback in `AESEncryptionAlgo`/`aes.dart` ("from the bad old days when we weren't setting IVs") so a refactor can't silently drop it.

## Test plan
- [x] `dart test test/aes_encryption_algo_test.dart` — 8/8 pass
- [x] `dart test` (full at_chops suite) — 485/485 pass
- [x] `dart analyze` — clean


srieteja:
Combined into #2247 per request.

PR #2249, opened 2026-09-16
fix(at_lookup): deliver a notification behind anything else on the wire

gkc:
**- What I did**
- A resilience suite for the monitor connection: dropped and faulted connections, ten outages in a row, an atServer that refuses to be reached or refuses the credential, a half-open connection the heartbeat rebuilds, rubbish on the wire, a handler that throws, two subscribers and one that attaches after the outage. Each one proves the same subscriber keeps receiving afterwards, that its stream never ends, and that no connection is left open
- These tests found a potential framing bug. It would cost a client every notification from the stray line until some later reply clears the buffer, by default up to a heartbeat away, on a connection that stays up throughout
- **NB** I am unaware of any way in which this bug could manifest itself right now so this is just defense in depth rather than an active fix. But I wouldn't discount the possibility that in weird or highly adverse network conditions it could happen I guess? maybe?
- `fix(at_lookup)`: the listener reads the connection as the line reader it is. A `\n` ends a message from the atServer and the atServer writes one at a time, so a line is a message: the prompt it may carry at the start comes off, and what is left says where it goes - `data:` and `error:` to the reader waiting for a reply, `notification:` to whoever asked for notifications, and a type added to the protocol later to its own route. That replaces hunting for a reply's terminating prompt, so a reply and a notification arriving in one chunk now both survive, where one of them used to be lost
- The reconnect delays and the monitor's start retries are settable, so the suite runs in a second

**- Why**
- A long-running service such as a daemon serving requests delivered via atsign protocol - e.g. the NoPorts daemon - is critically dependent on the notification connection healing itself and notifications continuing to flow

**- How I did it**
See commits. Stacked on #2245, which the suite needs for `Monitor.close()`; merge that first.

**- How to verify it**
Tests pass, including all four live packs locally

fix: defense in depth: a notification would no longer be lost if weird / unexpected traffic preceded it on the connection from the atServer, and a reply arriving in the same flow as a notification is no longer lost either


JeremyTubongbanua (approved):

JeremyTubongbanua (approved):

packages/at_client/test/monitor_resilience_test.dart
```diff
@@ -0,0 +1,412 @@
...
+      final late = <String>[];
+      service.subscribe(regex: '.*').listen((n) => late.add(n.id));
+      await deliver('after');
+
+      expect(late, ['after'],
+          reason: 'a program that subscribes on demand - one stream per job - '
+              'must not depend on having subscribed before the last outage');
+    });
+
+    test('stopListening then startListening again resumes the flow', () async {
```
Xlin123, line 379: this test stops listening, but never resumes like the title suggests

Xlin123 (changes requested):

packages/at_client/test/monitor_resilience_test.dart
gkc: Good catch - it only stopped. It restarts now: a notification before the stop, then the listener observed back at notConnected with no connection left open, then startListening, a second connection, and a notification after it - asserting both arrive, in order. 0014f53a3

srieteja (approved):

srieteja:
`packages/at_lookup/lib/src/connection/outbound_message_listener.dart:229`

Narrowing the notification test to the buffer's last line reopens a truncation bug: a continuation line inside a multi-line `data:`/`error:` value that happens to start with `notification:` is now emitted on the notification stream *and* deleted from the response buffer. `read()` still sees an intact `data:`/`error:` prefix, so the truncated reply comes back with no error.

**Repro:** atServer sends `data:first\nnotification: {"id":"n"}\nlast\n@alice@` in one write.
- Before: `read()` returns the full three-line value, no notification fires.
- After: notification stream emits the middle line, `read()` returns `data:first\nlast`.

Same trade-off this file's header already prices for the coalesced-frame case (the ⛔ list) — a fourth instance, not a new category. No clean automated fix: the only patch that stops this (whole-buffer-prefix check) reverts this PR's own fix and fails 3 of the tests it just added (`outbound_listener_framing_test.dart`, group `'a reply and a notification in one stream'`).

gkc:
Good catch, and I reproduced it both ways before changing anything - your repro in one write gave reply `data:first\nlast` with the notification emitted and no error, against `data:first\nnotification: {"id":"n"}\nlast` and no notification on the old code. So the narrowing was real.

But there is a clean fix, and it doesn't need a new delimiter. Two things are guaranteed: a `\n` ends a message from the atServer, and the atServer writes a single message at a time. That message may have a prompt (`@` or `@alice@`) at the START - which by design typically arrives in the same flow as the end of the previous line - or none. So it's a line reader: take the line, take off a leading prompt, and look at what it begins with - `data:`/`error:` is the response to a request this client made, `notification:` is asynchronous and goes to whoever asked for notifications, and a type added to the protocol later gets its own route rather than being mistaken for either. Entirely deterministic.

That's what db87ac431 does. It also closes the other gap the header described - a reply and a notification coalesced into one chunk used to lose the reply, and now keeps both - and a notification with nothing listening is dropped where it is, with a warning, instead of joining the next reply.

The tests that pinned a `data:` value containing newlines were pinning something the protocol doesn't allow, so they now pin the guarantee instead: a newline ends the message, and a prompt belongs to the message it precedes. at_lookup 176 and at_client 2094 pass, and I've mutated both properties to check the new tests bite.

srieteja (approved):

PR #2250, opened 2026-09-16
fix: close the gaps the deterministic-stop review found

srieteja:
Fixes on top of #2245 for the findings from its review.

## What
- **at_lookup dependency** — at_lookup now requires `at_commons: ^5.18.0`, the first version with `StoppedException`. Workspace resolution hid the old `^5.16.0` from CI.
- **`waitUntilCaughtUp()` after `close()`** — a wait that starts after `close()` now fails with `StoppedException`. It used to hang.
- **`AtRpcClient.call` during shutdown** — no uncaught async error when the send throws, and no `TypeError` when responses end while a send is in flight.
- **`pubKeyCache` growth** — expired entries are dropped on every insert, so the cache is limited to live keys as of the last insert.
- **`RemoteSecondary.closeConnection()`** — *documented, not changed.* It is now terminal: both SDK callers want that, and a non-terminal variant would need a downcast on `AtLookUp`. The dartdoc and the CHANGELOG point to `AtLookupMuxable.dropConnection()` for connections that should reconnect.

## How
- `SyncServiceImpl.addProgressListener` sends a `stopped` `SyncProgress` in a microtask **only when `_closed`**. `stop()` can be undone, and its documented flow is stop → add listeners → `start()`, so a listener added after a plain `stop()` gets no event.
  - A new private `_buildStoppedProgress()` builds the event for both this path and `_drainSyncQueue`.
- `AtRpcClient.call` holds its completer in a local variable and calls `future.ignore()` on it right away. If the send throws, the completer is removed from the map; otherwise `call` returns the local's future, so clearing the map can't break it.
- `cachePubKey` runs `removeWhere` on expired entries before inserting.
- New tests, each confirmed to fail without the `lib/` change:
  - `client_stop_test.dart`: a wait that starts after `close()`, plus a guard that a listener added after `stop()` hears nothing.
  - `rpc/at_rpc_client_stop_test.dart`: both shutdown orderings, run under `runZonedGuarded` and asserting no uncaught errors.
  - `envelope_signing_test.dart`: an expired entry for a different key is swept on insert.
- `dart analyze` is clean. `dart test` passes: at_client 2067, at_lookup 173.


gkc:
Thanks @srieteja

PR #2252, opened 2026-09-17
fix: review findings from #2245

srieteja:
Fixes what the review on #2245 turned up, applied on top of this branch.

- The newly installed `sharing.fileReceivedSecret` hook can throw, and because it runs after `secretStore.putIfNewer` has already committed, a failed filing is never retried and the secret is never emitted on `receivedSecrets` — waiters on that secret hang out their timeout and the key material is lost from durable storage. — `packages/at_client/lib/src/client/pq_client_bootstrap.dart:237`
- `expiringSoonEvents()` emits into a private StreamController that the backing `_CItemTimerScheduler` never closes on dispose — a consumer doing `await for` on the stream hangs forever after client stop. — `packages/at_client/lib/src/collections/collections.dart:4616`
- The `progress.stopped` branch in `waitUntilCaughtUp` completes the future with StoppedException but skips `detach()` — every call on an already-closed service leaks one listener for the life of the service object. — `packages/at_client/lib/src/service/sync_service.dart:294`

Full review: https://github.com/atsign-foundation/at_client_sdk/pull/2245 (report kept alongside this run's findings). `dart test` passes: at_client 2075/2075.


gkc:
Thanks - merging this, and then fixing the first one properly on the branch. Here's what I checked and what I'm doing.

**sync_service** - right, and merging as is. Every other terminal path in `waitUntilCaughtUp` calls `detach()` - both completes and the timeout - and only the `stopped` branch returned without it.

**collections** - right, and merging as is. Worth saying why it works, because `scheduler.dispose()` looks like it only runs from `onCancel`, where closing the controller would achieve nothing: it also runs from `collections.dart:4618`, where the scheduler listens to `collection.updates` with `onDone: () => unawaited(dispose())`. That is the client-stop path - the collection's event stream ends, the scheduler disposes, and nothing closed `ctrl`, so an `await for` waits for ever. `onDispose: ctrl.close` closes exactly that one. `availableEvents` emits through `watch()` rather than a private controller, so it doesn't need the same treatment.

**pq_client_bootstrap** - the diagnosis is right and the fix isn't, so I'm taking the diagnosis and changing the fix. The throw is deliberate: `fileReceivedSecret`'s dartdoc says "a throw keeps the envelope for the next sweep", and `sweepOnce` is built on it - on a throw it keeps both the claim and the envelope and logs "it is kept for a retry at the next start". Swallowing means the envelope is deleted instead, and leaves both of those sentences false.

But you're right that the retry is dead, and the cause is a line earlier than the one you changed. `putIfNewer` commits at `pairwise_secret_sharing.dart:1006`, before the filing at :1014. So at the next start the secret is already stored, `stored == false` short-circuits at :1007-1011, and the filing and the `receivedSecrets` emit are both skipped. The envelope is kept for a retry that can no longer do anything.

One small correction: the secret isn't lost from durable storage - `putIfNewer` committed it before the filing ran. What's lost is the filing.

So the fix is to let the early return distinguish "we hold something newer, ignore this" from "we hold exactly this one, and the filing may still be owed", and let the throw keep doing its job. I'll push that on top.

PR #2253, opened 2026-09-17
fix: fallback secondaryAddressFinder to CacheableSecondaryAddressFinder

JeremyTubongbanua:
## What

- `withSecureSocket` has a nullable parameter: `SecondaryAddressFinder? secondaryAddressFinder,`
- Fallbacks to `CacheableSecondaryAddressFinder`

## Skills impact


packages/at_lookup/lib/src/at_lookup.dart
```diff
@@ -115,7 +117,12 @@ abstract interface class AtLookUp {
       rootDomain.rootPort,
       secureSocketConfig: transport.secureSocketConfig,
       clientConfig: clientConfig,
-      secondaryAddressFinder: secondaryAddressFinder,
+      secondaryAddressFinder: secondaryAddressFinder ??
+        CacheableSecondaryAddressFinder(
+          rootDomain.rootDomain,
```
gkc, line 122: @JeremyTubongbanua this is already done in AtLookupImpl, why do you also need it here?

gkc (changes requested):

packages/at_lookup/lib/src/at_lookup.dart
JeremyTubongbanua: Good point

